PR #1, opened 2025-12-05
Add CI

ZacBlanco:
Add a basic CI setup using GitHub self-hosted runners and github workflows.

### Overview

Two directories under `./.github/`:

- `runners/`: scripts for deploying CI runners for this repository. The scripts use docker compose to deploy a small number of runner processes which register themselves with this repository on github.
- `workflows/`: Workflow descriptions for GitHub Actions

### Runners Setup

1. Install docker and docker compose on CI host
2. Modify docker-compose.yml to the desired number of runners and specific resource limits (CPU/memory) e.g. 4 medium runners or 2 large runners, etc
3. Set up the GITHUB_RUNNER_TOKEN, CONAN_ADMIN_USERNAME, and CONAN_ADMIN_PASSWORD variables inside `variables.env`
4. Run `./github/runners/setup-runners.sh`. This will build the container images and host all the runner services. You can check everything works correctly by inspecting the logs and seeing if the runners are registered against the repository

### DIFF
--- a/bolt/common/memory/ByteStream.h
+++ b/bolt/common/memory/ByteStream.h
@@ -319,14 +319,8 @@ class ByteOutputStream {
       return;
     }
 
-    auto* target = reinterpret_cast<T*>(current_->buffer + current_->position);
-    const auto* end = target + values.size();
-    auto* valuePtr = &values[0];
-    while (target != end) {
-      *target = *valuePtr;
-      ++target;
-      ++valuePtr;
-    }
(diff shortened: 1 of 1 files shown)


yangzhg (commented):
Why is there so much code formatting?

ZacBlanco:
The new image uses a new version of clang-format. The old debian image we used to use had clang-format 10. The new CI image uses clang-format 14. There seems to have been behavior changes between these major versions

PR #10, opened 2025-12-10
Refactor ignoreCorruptFiles

XinShuoWang:
In Spark, there is an feature named [ignoreCorruptFile](https://spark.apache.org/docs/latest/sql-data-sources-generic-options.html#ignore-corrupt-files), which allows you to using the configuration `spark.sql.files.ignoreCorruptFiles` to ignore corrupt files while reading data from files.

Bolt has implemented this feature, but the code is not clean. I want to refactor this feature, including below points:

- Use macro to simplify code and configuration.
- Adding some unit tests.

### DIFF
--- a/bolt/connectors/hive/IgnoreCorruptFile.h
+++ b/bolt/connectors/hive/IgnoreCorruptFile.h
@@ -19,8 +19,8 @@
 #include <fmt/core.h>
 #include <folly/Format.h>
 #include <glog/logging.h>
-#include <gtest/gtest.h>
 #include <cstdint>
+#include <cstdlib>
 #include <iostream>
 #include <set>
 #include <string>
@@ -35,9 +35,7 @@ class IgnoreCorruptFileHelper {
       int64_t taskMaxFailures,
       bool enableIgnoreCorruptFiles,
       const std::string& userDefineIgnoreExceptions) {
-    const ::testing::TestInfo* test_info =
-        ::testing::UnitTest::GetInstance()->current_test_info();
-    bool inTestEnv = (test_info != nullptr);
+    bool inTestEnv = isInTestEnv();
     if (hasInitialized_ && !inTestEnv) {
       return;
     }
@@ -91,6 +89,10 @@ class IgnoreCorruptFileHelper {
   }
 
  private:
+  static bool isInTestEnv() {
+    return std::getenv("BOLT_IN_GTEST") != nullptr;
+  }
+
   static void enrichExceptionSetFromConf(
       std::string exceptionStr,
       std::vector<std::string>& exceptionKeyWords,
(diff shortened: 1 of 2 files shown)


bolt/common/caching/AsyncDataCache.cpp
```diff
@@ -300,9 +300,7 @@ bool CoalescedLoad::loadOrFuture(folly::SemiFuture<bool>* wait) {
     setEndState(State::kLoading, State::kLoaded, State::kCancelled);
   } catch (std::exception& e) {
     // preload can fail in async thread, then retry in sync load
-    if (isAsyncPreloadThread() &&
-        std::string(e.what()).find(kAsyncPreloadThreadName) !=
-            std::string::npos) {
```
zhangxffff, line 305: Why `std::string(e.what()).find(kAsyncPreloadThreadName) != std::string::npos` was removed here?

bolt/connectors/hive/IgnoreCorruptFile.h
```diff
@@ -0,0 +1,201 @@
...
+#include "bolt/connectors/hive/storage_adapters/hdfs/StorageException.h"
+
+class IgnoreCorruptFileHelper {
+ public:
+  static void globalInitialize(
+      int64_t taskMaxFailures,
+      bool enableIgnoreCorruptFiles,
+      const std::string& userDefineIgnoreExceptions) {
+    const ::testing::TestInfo* test_info =
+        ::testing::UnitTest::GetInstance()->current_test_info();
```
zhangxffff: Do we have gtest dependency in a build without test? Can we wrapper test related logic inside a macro? So that we can avoid gtest dependency in release build.

bolt/common/caching/AsyncDataCache.cpp
XinShuoWang: > Why `std::string(e.what()).find(kAsyncPreloadThreadName) != std::string::npos` was removed here?

Ensure that Scan-related errors can be propagated upwards to `bolt/connectors/hive/HiveDataSource.cpp` and handled, so that the IgnoreCorruptFile-related logic doesn't need to be added in the asynchronous Preload logic.

bolt/connectors/hive/IgnoreCorruptFile.h
XinShuoWang: > Do we have gtest dependency in a build without test? Can we wrapper test related logic inside a macro? So that we can avoid gtest dependency in release build.

1. I didn't change the link settings when compile release version, so I think `gtest` was already linked by default in the original configuration.
2. I think it's great that you pointed this out, and I will try to remove the gtest dependency here.

XinShuoWang: > Do we have gtest dependency in a build without test? Can we wrapper test related logic inside a macro? So that we can avoid gtest dependency in release build.

I have completed the modifications in this commit:https://github.com/bytedance/bolt/pull/10/changes/7d4c8b122741a16ebef329bad78f849f921a87c0

PR #12, opened 2025-12-11
fix: refine setup script for bolt

kexianda:
* provide script for install gcc from source
* check conan installation, if no conan available install miniconda and conan


scripts/install-bolt-deps.sh
```diff
@@ -19,37 +19,16 @@ CUR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
...
-    cd -
-}
-
-export_conan_recipes_to_local_cache
+if ! conan remote list | grep -q 'bolt-recipes-local'; then
+    conan remote add bolt-recipes-local ${CUR_DIR}/scripts/conan --index 0
+fi 
+
+# Sync to conan-center-index@bad5c95 2025.08.01
+conan_center_commit_id="bad5c95"
```
ZacBlanco, line 26: Can you explain why we need this part with a clone of conan-center-index? Is there a specific recipe we need?

scripts/setup-dev-env.sh
```diff
@@ -39,29 +41,31 @@ function check_compiler() {
...
-        echo "❌ clang is not installed"
-        return 1
-    fi
-
-    clang_version=$(clang --version | grep -oP '(?<=^clang version )\d+' | head -n 1)
-
-    if [ "$clang_version" -eq 16 ]; then
-        echo "✅ clang 16 is founded"
-        return 0
-    fi
```
ZacBlanco, line 54: Are we going to remove support for clang?

scripts/install-bolt-deps.sh
kexianda: The recipes in conan-center-index are updated very frequently.  Some dependencies are bumped up, some versions are removed.  
If Bolt is built from scratch,  the updated third-parties can cause compiling errors. For instance, duckdb/0.8.1 was removed in the latest conan-center-index repo.

scripts/setup-dev-env.sh
kexianda: I've tested  coroutine + clang 16 + libstdc++(standard library).  It can crash sometimes.  But  coroutine + clang + libc++ will works.   Before Makefile/conanfile is refined,  remove clang version check here.

kexianda: for clang, we have to specify the standard library as libc++.   I will fix it.

scripts/install-bolt-deps.sh
kexianda: Conan center index has no release tags,  fix it to a commit id to avoid unstable failures.

ZacBlanco: I see this makes sense. What do other projects do that rely on the public conancenter? It seems like a bad decision from the conan maintainers to remove old recipe versions. 

Also, do you think it would be better to use a git submodule for the repository rather than putting the commit hash in this file? Regardless of either approach I also think we should use `--depth 1` to reduce cloning times

scripts/conan/recipes/duckdb/all/patches/0.9.2-0001-fix-cmake.patch
ZacBlanco, line 1: It seems you have copied the entire recipe for all available conan versions. To help keep the repository footprint smaller, I think it might be better to only include the patches and conandata.yml entries for the version that bolt depends on. It will help significantly reduce the size of the PR by adding only the necessary files

scripts/conan/recipes/gnu-config/all/conanfile.py
```diff
@@ -0,0 +1,65 @@
...
+required_conan_version = ">=1.52.0"
+
+
+class GnuConfigConan(ConanFile):
+    name = "gnu-config"
+    description = "The GNU config.guess and config.sub scripts"
+    homepage = "https://savannah.gnu.org/projects/config/"
+    url = "https://github.com/conan-io/conan-center-index"
+    topics = ("gnu", "config", "autotools", "canonical", "host", "build", "target", "triplet")
+    license = "GPL-3.0-or-later", "autoconf-special-exception"
```
frankobe: GPL license conflicts with apache 2.0 license, it should be avoided

scripts/install-gcc.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
frankobe: This is the license from ASF, not from Bolt

kexianda: fixed

scripts/conan/recipes/gnu-config/all/conanfile.py
kexianda: Fixed.  Now, all the copied conan recipes are removed, just bolt's patches  are applied for the conan recipes.
Bolt did not  directly depends on gnu-config.  Fix the bug in gnu-config recipes for the users who want to build from scratch.

PR #15, opened 2025-12-11
ci: Add debug to build-test matrix

ZacBlanco:


zhangxffff:
`make unittest_release_spark` is useful to testing some spark related logic, maybe we can also add as a check.

ZacBlanco:
I agree, we should probably have some spark variation of the tests run. However, our CI resources are limited at the moment so I am hesitant to commit to adding it. I don't want CI queues to be too long. I will add this to my list of potential workflows to add though.

PR #19, opened 2025-12-12
ci: Add benchmark and release_spark to build-test workflow

ZacBlanco:

Adds missing CI jobs which are critical to ensuring bolt's functionality is correct.

### Type of Change

- [ ] 📝 Documentation only

### Description

1. Adds a benchmarks-build job to ensure benchmarks are always buildable. We don't need to run the benchmarks in CI because the environment is not stable enough to ensure consistent performance. It is up to users to run them. CI needs to just make sure they are not broken.
2. Adds `release_spark` to build matrix alongside `release`. The bolt backend is a critical part of the Gluten project. And when users build bolt for gluten they should be building with the `release_spark` target. This ensures that we build and test with the Gluten-compatible profile so that we know it always works.
3. Refactors the build-test action to have a common setup composite workflow
4. Fixes the CI image to put the python venv bin on the path first so the venv python and pip, and other installed pip-installed package binaries are available without having to source or activate the venv within the CI image

### Performance Impact

N/A

### Release Note

N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/.github/actions/bolt-build-base/action.yml
+++ b/.github/actions/bolt-build-base/action.yml
@@ -7,7 +7,6 @@ runs:
       shell: bash
       run: |
         conan profile detect
-        sed -i 's/gnu14/gnu17/g' ~/.conan2/profiles/default
         cat >> ~/.conan2/profiles/default <<EOF
 
         [conf]


.github/actions/bolt-build-base/action.yml
```diff
@@ -0,0 +1,27 @@
+name: 'Bolt Build Setup Steps'
+description: 'Executes necessary commands to setup environment for the bolt build'
+runs:
+  using: "composite"
+  steps:
+    - name: Setup conan profile
+      shell: bash
+      run: |
+        conan profile detect
+        sed -i 's/gnu14/gnu17/g' ~/.conan2/profiles/default
```
kexianda: for gcc 12,  the default flag is gun17.   This line can be removed.

ZacBlanco: Thanks for pointing that out! Fixed.

kexianda (approved):

PR #30, opened 2025-12-16
ci: Fix devcontainer and improve setup script

ZacBlanco:

Fixes build issues with linking the conan LLVM dependency inside the devcontainer and improves setup scripts for better E2E CI time.

### Type of Change

- [ ] 📝 Documentation only

### Description

Fixes a few issues the devcontainer was having after updating the build scripts

- Adds mold's `ld` to the front of the path so that it is the default system linker. This fixes a build issue in the devcontainer where the original system `ld` did not support the --color-diagnostics flag. By making mold the first on the path it fixes the issue and also has the bonus effect of making all linking across the system faster when projects search for `ld` on `PATH`
- Installs mold via download instead of build from source
- Makes a non-root user the default user (but still gives passwordless sudo access)

In the devcontainer setup:

- lowercases the https_proxy and no_proxy build args so they are respected by tools
- adds default containerEnv variables for https_proxy and no_proxy
- sets remoteUser to the default "code" user in the container
- sets the conan and ccache volumes to named volumes. This is for compatibility with macos and linux containers. Podman desktop on MacOS had issues with bind mount open file limits.

In install-bolt-deps
- Improves script by applying changes from shellcheck linter
- Does a shallow checkout of conan-center-index to improve setup time

### Performance Impact

N/A

### Release Note

Release Note:

```text
Release Note:
- Make mold the default linker in the devcontainer
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/.github/runners/Dockerfile.ci
+++ b/.github/runners/Dockerfile.ci
@@ -8,7 +8,7 @@ ARG DEB_REGION=""
 ARG https_proxy=""
 ARG no_proxy=""
 
-# check if deb region is not empty string. If it is not, run the sed command. If it is empty, do nothing.
+# check if deb region is not empty string. If it is, replace the default debian repository with the one for the region.
 RUN if [ "$DEB_REGION" != "" ]; then \
     sed -i "s|http://deb.debian.org/debian|http://ftp.${DEB_REGION}.debian.org/debian|g" /etc/apt/sources.list.d/debian.sources; \
 fi
@@ -26,7 +26,6 @@ RUN arch=$(arch) && \
     ln -s /opt/cmake-$CMAKE_VERSION-linux-${arch}/bin/ctest /usr/local/bin/ctest && \
     rm /tmp/cmake.tar.gz
 
-# Install vim, sudo, git, gdb, python3, pip, ssh, ninja, bison, maven
 RUN apt-get update && apt-get install -y \
     sudo \
     bash \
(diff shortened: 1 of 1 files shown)


scripts/install-bolt-deps.sh
```diff
@@ -13,19 +13,32 @@
...
+# Does a shallow checkout of conan-center-index at the given commit id in $1
+checkout_conan_center_index() {
+    rm -rf "${cci_home}"
+    mkdir -p "${cci_home}"
+    pushd "${cci_home}"
+    git init
+    git remote add origin https://github.com/conan-io/conan-center-index.git
+    git fetch --depth 1 origin "${1}"
+    git checkout FETCH_HEAD
+    popd
```
ZacBlanco, line 30: This will significantly improve E2E times in the CI (by about 30s to 1 minute) due to improved cloning time

.github/runners/Dockerfile.ci
```diff
@@ -62,13 +63,26 @@ RUN apt-get update && apt-get install -y \
...
-RUN https_proxy=${HTTPS_PROXY} git clone https://github.com/rui314/mold.git /tmp/mold
-RUN cd /tmp/mold && \
-    git checkout $MOLD_VERSION && \
-    mkdir build && \
-    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=c++ -B build && \
-    cmake --build build -j$(nproc) && \
-    sudo cmake --build build --target install
+RUN https_proxy=${HTTPS_PROXY} curl -L -O --output-dir /tmp https://github.com/rui314/mold/releases/download/v$MOLD_VERSION/mold-$MOLD_VERSION-$(arch)-linux.tar.gz
+RUN cd /tmp && \
+    sudo tar -C /usr/local --strip-components=1 -xzf mold-$MOLD_VERSION-$(arch)-linux.tar.gz
```
ZacBlanco: downloading pre-built binaries instead of compiling to save time during devcontainer build

scripts/install-bolt-deps.sh
ZacBlanco: This was the setup time with this new method:

image


And this is the time without it

image

kexianda (approved):
+1

PR #37, opened 2025-12-16
[celeborn] Add celeborn conan dependency

### LINKED ISSUE #36: [Chore] [Celeborn] Add Celeborn C++ client as conan dependency
### Chore Type

Dependency Update (Conan/Third-party)

### Task Description

Add celeborn client as conan dependency

### Motivation / Benefit

Bolt relies on the [Celeborn C++ client](https://github.com/apache/celeborn/tree/main/cpp)￼ to communicate directly with the Celeborn shuffle service, eliminating the need for a JNI layer. Therefore, we need to add the Celeborn C++ client as a Conan dependency to enable seamless integration with Celeborn.

### Related Files (Optional)

_No response_

### Implementation Plan (Optional)

_No response_

### Additional Context

_No response_

zhangxffff:

Issue Number: close #36 

### Type of Change

- [ ] 📝 Documentation only

### Description

Pack celeborn c++ client as a conan recipe and add shuffle writer/reader demo with celeborn shuffle client

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>


liushengxuan:
LGTM

.licenserc.yaml
```diff
@@ -112,6 +112,8 @@ header:
     - 'LICENSE'
     - 'MANIFEST.in'
     - 'scripts/conan/patches/**'
+    - 'scripts/conan/recipes/**/*.patch'
```
ZacBlanco: The `**/*.patch` pattern above should already match this

.licenserc.yaml
```diff
@@ -112,6 +112,8 @@ header:
...
+    - 'scripts/conan/recipes/**/*.yml'
```
ZacBlanco: Since yml/yaml allows comments, I think we should put the licenses in the yaml files

conanfile.py
```diff
@@ -356,6 +356,7 @@ def requirements(self):
         self.requires("utf8proc/2.11.0", transitive_headers=True, transitive_libs=True)
         self.requires("date/3.0.4-bolt", transitive_headers=True, transitive_libs=True)
         self.requires("libbacktrace/cci.20210118")
+        self.requires("celeborn-cpp-client/main-20251212")
```
ZacBlanco: Wondering if we need to put this dependency/feature behind a flag. How large is celeborn? Will all users of bolt want or need this? If this is a spark-only feature, I think it might need to go behind the spark flag.

scripts/conan/recipes/celeborn-cpp-client/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,38 @@
...
+  conf.registerProperty(
+      conf::CelebornConf::kClientPushLimitInFlightSleepDeltaMs,
+      std::to_string(0));
+  conf.registerProperty(
+      conf::CelebornConf::kClientPushMaxReqsInFlightTotal, std::to_string(0));
+  conf.registerProperty(
+      conf::CelebornConf::kClientPushMaxReqsInFlightPerWorker,
+      std::to_string(0));
+
+  std::unique_ptr<PushState> pushState = std::make_unique<PushState>(conf);
```
ZacBlanco: Is this code enough to properly test the celeborn dependency? Should it try to construct a client/server?

scripts/conan/recipes/wangle/config.yml
ZacBlanco, line 1: Can we add a `test_package` for wangle to ensure the recipe works correctly?

scripts/conan/recipes/fizz/config.yml
ZacBlanco, line 1: Can we add a `test_package` for fizz to ensure the recipe works correctly?

bolt/shuffle/sparksql/partition_writer/rss/RssClient.h
```diff
@@ -29,6 +29,7 @@
  * This modified file is released under the same license.
  * --------------------------------------------------------------------------
  */
+#include <cstdint>
```
ZacBlanco, line 32: Why is this needed?

bolt/shuffle/sparksql/CelebornReaderStreamIterator.cpp
ZacBlanco, line 1: Could you add some tests for this class?

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.cpp
ZacBlanco, line 1: Could we add some tests for this class?

.licenserc.yaml
zhangxffff: removed

.licenserc.yaml
zhangxffff: Removed and added the license header in the newly added YAML file.

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.cpp
zhangxffff: `bolt/shuffle/sparksql/tests/CelebornClientTest.cpp` was added to test `NativeCelebornClient` and `CelebornReaderStreamIterator` with mocked celeborn shuffle client.

bolt/shuffle/sparksql/CelebornReaderStreamIterator.cpp
zhangxffff: Testcases added in `bolt/shuffle/sparksql/tests/CelebornClientTest.cpp `

bolt/shuffle/sparksql/partition_writer/rss/RssClient.h
zhangxffff: int32_t and int64_t are used in this header. Adding <cstdint> makes the dependency explicit and avoids relying on indirect includes.

scripts/conan/recipes/fizz/config.yml
zhangxffff: Added a test_package to validate the fizz recipe.

scripts/conan/recipes/wangle/config.yml
zhangxffff: Also added a test_package for test_package

conanfile.py
zhangxffff: Add an option BOLT_ENABLE_CELEBORN_CPP_CLIENT to enable this feature.

bolt/shuffle/sparksql/tests/CelebornClientTest.cpp
```diff
@@ -0,0 +1,460 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "bolt/common/base/Exceptions.h"
+#include "bolt/shuffle/sparksql/CelebornReaderStreamIterator.h"
+#include "bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.h"
+
+namespace bytedance::bolt::shuffle::sparksql::test {
+std::shared_ptr<const celeborn::protocol::PartitionLocation> MakeLocation(
```
ZacBlanco: nit: generally method names start with a lowercase letter in the bolt codebase. Classes start with uppercase. This is currently not strictly enforced by clang-tidy, but this is the most common pattern.

ZacBlanco: I see this occurs across more methods in the file, please update all locations

bolt/shuffle/sparksql/CelebornReaderStreamIterator.cpp
```diff
@@ -0,0 +1,161 @@
...
+    int32_t endMapIndex,
+    bool needCompression)
+    : shuffleClient_(std::move(shuffleClient)),
+      shuffleId_(shuffleId),
+      partitionIds_(std::move(partitionIds)),
+      attemptNumber_(attemptNumber),
+      startMapIndex_(startMapIndex),
+      endMapIndex_(endMapIndex),
+      needCompression_(needCompression) {
+  BOLT_CHECK(shuffleClient_, "ShuffleClient is null for Celeborn reader");
```
ZacBlanco: nit: consider using `BOLT_CHECK_NOT_NULL`

bolt/shuffle/sparksql/CelebornReaderStreamIterator.cpp
```diff
@@ -0,0 +1,161 @@
...
+  }
+  auto partitionId = partitionIds_[nextPartitionIndex_++];
+  auto celebornStream = shuffleClient_->readPartition(
+      shuffleId_,
+      partitionId,
+      attemptNumber_,
+      startMapIndex_,
+      endMapIndex_,
+      needCompression_);
+  BOLT_CHECK(
```
ZacBlanco: nit: consider `BOLT_CHECK_NOT_NULL`

bolt/shuffle/sparksql/CMakeLists.txt
```diff
@@ -25,34 +25,58 @@
...
+    APPEND
+    bolt_shuffle_spark_impl_sources
+    CelebornReaderStreamIterator.cpp
+    partition_writer/rss/NativeCelebornClient.cpp)
+  find_package(celeborn-cpp-client CONFIG REQUIRED)
+endif()
+
+add_library(bolt_shuffle_spark_impl ${bolt_shuffle_spark_impl_sources})
+
+target_link_libraries(
```
ZacBlanco, line 59: I'm not sure if we should entirely replace the implementation at compile time. Would it be possible to use a configurable flag at runtime to decide which shuffle client to use?

bolt/shuffle/sparksql/CMakeLists.txt
```diff
@@ -25,34 +25,58 @@
...
-    partition_writer/rss/RemotePartitionWriter.cpp
-    partition_writer/rss/CelebornPartitionWriter.cpp
-    compression/Compression.cpp
-    Payload.cpp
-    ShuffleMemoryPool.cpp
-    Spill.cpp
-    Utils.cpp)
+  PUBLIC arrow::arrow ryu::ryu bolt_vector bolt_core bolt_exec)
+
+if(BOLT_ENABLE_CELEBORN_CPP_CLIENT)
```
SteNicholas: Could this branch merge with line 55?

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.h
```diff
@@ -0,0 +1,53 @@
...
+  NativeCelebornClient(
+      std::shared_ptr<celeborn::client::ShuffleClient> shuffleClient,
+      int32_t shuffleId,
+      int32_t mapId,
+      int32_t attemptId,
+      int32_t numMappers,
+      int32_t numPartitions);
+  ~NativeCelebornClient() override = default;
+
+  int32_t pushPartitionData(int32_t partitionId, char* bytes, int64_t size)
```
SteNicholas, line 40:
```suggestion
  int32_t pushData(int32_t partitionId, char* bytes, int64_t size)
```

zhangxffff: > 

`NativeCelebornClient::pushPartitionData` overrides `RssClient::pushPartitionData`. Since `RssClient` is also used as the base class for the RSS writer in the Gluten Bolt backend, renaming it here would not be straightforward. Maybe we can rename it togather with gluten bolt backend later.

bolt/shuffle/sparksql/CMakeLists.txt
zhangxffff: Done — merged branch with line 55. Please re-review the latest commit.

zhangxffff:
@ZacBlanco Thanks again for the review. I’ve addressed all the comments and updated the branch. When you have a moment, could you please take another look?

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.cpp
```diff
@@ -0,0 +1,65 @@
...
+      0,
+      size,
+      numMappers_,
+      numPartitions_);
+}
+
+void NativeCelebornClient::stop() {
+  if (!shuffleClient_) {
+    return;
+  }
```
ZacBlanco: This method doesn't seem to actually do anything. Is that intentional?

ZacBlanco (changes requested):
Sorry for the delay. Thank you for addressing the comments. I had just a few more

bolt/shuffle/sparksql/CMakeLists.txt
zhangxffff: Currently, this implementation is not actually used outside of unit tests. After integration with Gluten, an option will be introduced and controlled by Gluten to determine whether to use the Celeborn C++ client directly or the Gluten-wrapped Java Celeborn client. This will be implemented in a follow-up PR, after which the choice of client can be controlled via a configurable flag at runtime.

bolt/shuffle/sparksql/tests/CelebornClientTest.cpp
zhangxffff: > nit: generally method names start with a lowercase letter in the bolt codebase. Classes start with uppercase. This is currently not strictly enforced by clang-tidy, but this is the most common pattern.

Thanks for pointing this out. I’ve updated all the occurrences in this file to follow the naming convention.

bolt/shuffle/sparksql/CelebornReaderStreamIterator.cpp
zhangxffff: > nit: consider using `BOLT_CHECK_NOT_NULL`

Thanks for the suggestion. I've updated this to use BOLT_CHECK_NOT_NULL.

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.cpp
zhangxffff: > This method doesn't seem to actually do anything. Is that intentional?

Thanks for catching this. It’s a mistake here: we should call shuffleClient_->mapperEnd() on stop. I’ve fixed it..

bolt/shuffle/sparksql/CMakeLists.txt
ZacBlanco: Thanks for the explanation. My concern is that doing it this way seems to create more options in our build matrix which is already quite complex. It would be easiest from a build and CI perspective to have as few build options as possible, and just run tests with different sets of configuration flags.

I personally would prefer a runtime flags to be able to configure either, but I understand there is already little to no use of the existing shuffle client outside of the tests. I personally don't know what the best approach would be here due to my limited background with bolt's shuffle client.

@markjin1990 do you have any strong opinions about this?

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.cpp
```diff
@@ -0,0 +1,64 @@
...
+      partitionId,
+      reinterpret_cast<const uint8_t*>(bytes),
+      0,
+      size,
+      numMappers_,
+      numPartitions_);
+}
+
+void NativeCelebornClient::stop() {
+  shuffleClient_->mapperEnd(shuffleId_, mapId_, attemptId_, numMappers_);
```
ZacBlanco, line 67: Should we do a defensive check in case `stop` is called twice? I have not dug too deep, but I see in the celeborn client code there doesn't seem to be a guard to prevent this from being called twice: https://github.com/apache/celeborn/blob/d30c02e3690a5ecfe3cbb32f2d21ba101d37c679/cpp/celeborn/client/ShuffleClient.cpp#L220-L242

Or if it is called twice, should we just let it throw an exception (in the case that an exception does get thrown)?

ZacBlanco (commented):
One last question

bolt/shuffle/sparksql/CMakeLists.txt
zhangxffff: Thanks for raising this concern. I understand your point. We currently have quite a few build-time options, but only a small subset are actually used in practice. For this particular option, I share the same concern that it may not be meaningfully used. If we end up always building with it enabled anyway, it might be cleaner to remove this option and include it unconditionally in the Spark build.

bolt/shuffle/sparksql/partition_writer/rss/NativeCelebornClient.cpp
zhangxffff: Thanks for reminding me of this. It’s better to add a defensive check here, and I’ll include it in `stop`.

zhangxffff:
@ZacBlanco Many thanks for your review. I’ve removed the `BOLT_ENABLE_CELEBORN_CPP_CLIENT` related option, added a guard to throw an exception if `stop` or `pushData` is called after `NativeCelebornClient::stop`, and added a test case to cover this behavior. Could you please review it again when you have time?

ZacBlanco (approved):
LGTM

PR #75, opened 2026-01-05
[feat][serde] Adopt arrow serializer in VectorSerde

### LINKED ISSUE #73: [Feature] Adopt arrow serializer in VectorSerde
### Feature Category

Performance Optimization

### Problem / Use Case

Currently, Bolt using PrestoSerde to serialize RowVector when spill, but it seems PrestoSerde has performace issue. For example, in one workload, Row based spills more data (407G vs. 301G), but uses much less time for serialization (6.28h vs. 11.30h) and compression + flush (4.19h vs. 5.6h). It seems for spilling, the bottleneck is not on I/O (25min vs. 8min) but the serialization.

  - Basic structure for serializer: append * n -> flush 
  - Reason why Presto serializer is slow:
    - Copy once to StreamArena when append with flatten cost (if need)
    - Another copy to Outputstream when flush happened.
- Current Bolt (SEP.25.2025) has 3 serializers (in bolt/serializers)
  - Row based: *RowSerializer 
  - Column based: Only PrestoSerializer 
  - Newly added: ArrowSerializer

### Proposed Solution

- Using Arrow to make serializer faster
  - Bolt RowVector is Apache Arrow compatible.
  - Basic Idea: Bolt RowVector <-> Arrow <-> ArrowIPC

### References / Prior Art

_No response_

### Importance

Medium (Nice to have)

### Willingness to Contribute

Yes, I can submit a PR

zhangxffff:

Issue Number: close #73 

### Type of Change

- [ ] 📝 Documentation only

#### Brief Implementation Details

- Bridge reuse & fixes, handled with exportToArrowIPC flag.
  - Code for Bolt RowVector <-> Arrow
  - Support additional types (UNKNOWN, TimestampWithTimeZone, INTERVAL_DAY_TIME)
  - Handle deeply nested types (ARRAY/MAP/ROW), and materialize offsets where required.
  - IPC disallows nested dictionaries -> Ensures Arrow-IPC compliance by detecting and expanding dictionary-in-dictionary.
- ArrowSerializer
  - Inherit VectorSerializer interface
    - designed as a drop-in replacement for the PrestoSerializer.
  - append: Convert RowVector -> Arrow RecordBatch and retain Arrow buffers/holders (reclaim-safe, maximizes zero-copy downstream).
  - flush: Write the RecordBatch to the Arrow IPC stream (byte writes are unavoidable). To minimize copying, prewrite an empty header, write the body first, then back-patch the header fields.
  - Deprecated/encoding-preserving APIs are not implemented.
- deserialize()
  - Open an Arrow IPC reader:
    - Wrap ByteInputStream with BoltArrowInputStream and open a RecordBatchStreamReader (zero-copy stream from the source).
    - Compress code only for future support, currently unsupported/untested and kept as placeholder code.
  - Read the first batch:
    - If empty, return an empty/reshaped RowVector.
    - Otherwise, export Arrow C Data -> import as a Bolt RowVector, then validate the schema against the requested RowType.
  - Zero-copy policy:
    - Single batch & adoptable (*result == nullptr and resultOffset == 0) -> zero-copy: adopt the imported RowVector and drain the reader.
    - Otherwise, need to merge with existing RowVector -> materialize/copy.

#### How and when to enable

Since converting a RowVector to Arrow has a non-trivial overhead, we only enable Arrow serialization when spilling a single partition (e.g., OrderBy or HashAggregation). We add a new option, spill_single_partition_serde_kind, to control the serialization format for single-partition spills. In addition, to simplify passing parameters from Spiller to SpillWriter, we introduce a SpillIOConfig struct to bundle the relevant I/O settings.

### Performance Impact

#### String Type
image

- Case 1: all-inline (almost same as presto)
  - Bytes: 108.04% (because of null ratio 0.1)
- Case 2 (Extreme, maximize benefit) card = 100 for 8000 len string dict
  - Bytes: 0.39% of presto, 202x faster
- Case 3 (card increase, less maximize benefit)
  - Bytes: 17.17% use, 18x faster

#### other scalar type
image

- Brief result comparison (compare with Presto Serializer)
  - ser : serialize time
  - de : deserialize time
  - sum = ser + de
  - bytes = size of serialized data
- Most scalar type 2-3x faster
- Almost the same size of bytes.
- Cons: Bytes do not decrease when row null increases.

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/serializers/tests/ArrowSerializerTest.cpp
+++ b/bolt/serializers/tests/ArrowSerializerTest.cpp
@@ -48,6 +48,10 @@ class ArrowSerializerTest
     memory::MemoryManager::testingSetInstance({});
   }
 
+  static void TearDownTestCase() {
+    deregisterVectorSerde();
+  }
+
   void SetUp() override {
     serde_ = std::make_unique<serializer::arrowserde::ArrowVectorSerde>();
   }
--- a/bolt/serializers/tests/PrestoSerializerTest.cpp
+++ b/bolt/serializers/tests/PrestoSerializerTest.cpp
@@ -69,6 +69,10 @@ class PrestoSerializerTest
     memory::MemoryManager::testingSetInstance(memory::MemoryManager::Options{});
   }
 
+  static void TearDownTestCase() {
+    deregisterVectorSerde();
+  }
+
   void SetUp() override {
     serde_ = std::make_unique<serializer::presto::PrestoVectorSerde>();
   }


zhangxffff:
Key new changes:
1. Added support for compression and decompression.
2. Implemented memory allocation via bolt::MemoryPool during serialization/deserialization (serde).
3. Introduced options to enable Arrow-based serde for single-partition spills.
4. Refactored configuration passing from Spiller to SpillWriter using a packed struct.
5. Added new unit tests to verify Arrow serializer performance in OrderBy and HashAggregate spill scenarios.

bolt/core/QueryConfig.h
```diff
@@ -1096,6 +1100,10 @@ class QueryConfig {
     return get<std::string>(kRowBasedSpillMode, kDefaultRowBasedSpillMode);
   }
 
+  std::string singlePartitionSpillSerdeKind() const {
```
markjin1990, line 1110: Should we add available options in the comment section?

markjin1990:
@zhangxffff Can you elaborate on what is "single partition" spill in OrderBy and HashAgg? Also, does the claimed performance gain include includes the "non-trivial" overhead converting a RowVector to Arrow? I am not sure if I understand the reason we limit arrow serde in only single partition spill.

zhangxffff:
> @zhangxffff Can you elaborate on what is "single partition" spill in OrderBy and HashAgg? Also, does the claimed performance gain include includes the "non-trivial" overhead converting a RowVector to Arrow? I am not sure if I understand the reason we limit arrow serde in only single partition spill.

A “single partition” spill means that the entire RowVector is written into a single spill file, and the RowVector is serialized as a whole. In operators such as OrderBy and HashAggregate, spilled data always originates from a single RowVector, so spills naturally occur as a single partition. In contrast, for HashJoin, when the memory limit is low, the hash table may be split into multiple partitions so that each partition can be loaded into memory independently. In this case, the RowVectors for both the hash build and probe sides must also be partitioned by the join key, then serialized and spilled into multiple files.

The current performance gains includes the overhead of converting a RowVector to Arrow. However, when the batch size is small, the Arrow conversion overhead can outweigh the benefits and may even lead to performance regressions in extreme cases. Therefore, enabling Arrow-based serialization only for single-partition spills is a conservative and safe approach to realize performance benefits while avoiding potential slowdowns.

markjin1990 (commented):
@zhangxffff Also, is the ArrowSerializer adopted in bolt shuffle writer? It was using PrestoSerializer for complex data types if I remember correctly.

zhangxffff:
> @zhangxffff Also, is the ArrowSerializer adopted in bolt shuffle writer? It was using PrestoSerializer for complex data types if I remember correctly.

Yes, ShuffleWriter currently uses PrestoSerde to serialize complex types. However, ShuffleWriter splits the input RowVector by partition, which means the serialization granularity per partition is typically quite small. As a result, the RowVector-to-Arrow conversion overhead becomes non-trivial, so we do not adopt ArrowSerializer in the Bolt shuffle writer. 

For the Bolt shuffle writer, I propose replacing PrestoSerde with an optimized row-format SerDe as described in this proposal: #80

bolt/core/QueryConfig.h
zhangxffff: Thanks for the review. I’ll add the supported options to the comments.

zhangxffff: The supported options have been added to the comments. Thanks, and please feel free to review again when you have time.

bolt/serializers/ArrowSerializer.h
```diff
@@ -0,0 +1,153 @@
...
+
+    static ArrowSerdeOptions fromPrestoOptions(
+        presto::PrestoVectorSerde::PrestoOptions* prestoOptions);
+
+   private:
+    static ArrowOptions defaultArrowOptions(bool losslessTs = false) {
+      ArrowOptions arrowOptions;
+      arrowOptions.timestampUnit =
+          losslessTs ? TimestampUnit::kNano : TimestampUnit::kMicro;
+      arrowOptions.flattenDictionary = true;
```
markjin1990, line 74: I think according to Jun Hyun, a main reason why we see a "non-trivial" overhead is because we do flattenDictionary and flattenConstant here, because "ArrowIPC requires same type encoding for all batches" whereas it seems not guaranteed here in Bolt. If possible, we can leave a TODO here and cc you and me, so that we can further improve the performance of this part in future to allow opportunities to apply ArrowSerializer in more scenarios.

zhangxffff: Thanks for the suggestion! I've added a TODO in ArrowSerializer.h documenting this overhead issue and the benchmark tests. 
Also added two benchmark cases in ArrowSerializerBenchmark.cpp:
- "Fixed Overhead Analysis" - tests flat data from 1 to 32K rows
- "Dictionary-Wrapped RowVector" - simulates HashJoin spill scenario
The benchmark result shows Arrow is slower below ~1K rows (flat) or ~8K rows (dictionary-wrapped), but faster for larger batches.

PR #83, opened 2026-01-06
ci: Add debug build to CI matrix

ZacBlanco:

Adds the debug and test build to the CI matrix

### Type of Change

- [ ] 📝 Documentation only

### Description

Adds debug mode build to CI builds.

### Performance Impact

### Release Note

N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/version/CMakeLists.txt
+++ b/bolt/version/CMakeLists.txt
@@ -102,3 +102,4 @@ configure_file(
   "${CMAKE_SOURCE_DIR}/bolt/version/version.h.in" "${OUTPUT_DIR}/version.h"
   @ONLY
 )
+install(FILES "${OUTPUT_DIR}/version.h" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${REL_PATH}")


conanfile.py
```diff
@@ -481,6 +483,9 @@ def generate(self):
         tc = CMakeToolchain(self, generator="Ninja")
 
         tc.cache_variables["MAX_LINK_JOBS"] = num_link_job
+        
+        if self.options.get_safe("ci_debug_minsize"):
+            tc.cache_variables["CMAKE_CXX_FLAGS_DEBUG"] = ""
```
kexianda: No need to touch this?
CMAKE_CXX_FLAGS_DEBUG flag will controlled by  `conan build -s build_type=xxx`

To use the `Release` dependencies:
`conan install . -sbuild_type=$${DEPENDENCY_BUILD_TYPE:-${BUILD_TYPE}}`
It will work, I guess.

ZacBlanco: In my testing I found CMAKE_CXX_FLAGS_DEBUG is always set to "-g". In order to create a "debug" build but without the debug info, I needed to remove the "-g" argument from the compilation flags. I found if I just set this variable to "" then -g is removed from the compilation command. This variable is set even in non-debug builds. It just only gets used when the build type is set to Debug.

This was the solution I came up with to strip debug info, but I'm happy to do it another way if you have a suggestion

conanfile.py
```diff
@@ -173,6 +173,7 @@ class BoltConan(ConanFile):
         "enable_test": [True, False],
         "build_benchmark": ["off", "basic", "on"],
         "enable_coverage": [True, False],
+        "ci_debug_minsize": [True, False],
```
kexianda: The conan options are the attributes of a library, and these options will exposed to the user (gluten/presto ...)
From my point of view,  "ci_debug_minsize" should not be an `option`. 
It is only for CI.  Use an environment variable?

btw, `build_benchmark` / `enable_coverage`  also should be removed. (I will try to remove them later)

ZacBlanco: Got it. I think that makes sense.

I changed the strategy to modify the cmake variable by instead configuring the profile with the following value in the `[conf]` section:

```
tools.cmake.cmaketoolchain:extra_variables={"CMAKE_CXX_FLAGS_DEBUG": ""}
```

This avoids having any specific logic in the build conanfile for CI, which I feel is a much better solution. Configuring the environment seems better to me anyways.

kexianda (approved):
LGTM

PR #217, opened 2026-02-09
Unify behavior for run-clang-tidy in CI and local

ZacBlanco:

Fixes inconsistent behavior causing failures for clang-tidy across PR, Merge Queue, and Push/Merges on the `main` branch

### Type of Change

- [ ] 📝 Documentation only

### Description

The goal of this change is to introduce new behavior for the clang-tidy pre-commit check so that we can enable merge queue behavior for the bolt repository.

Previously, the clang-tidy pre-commit check behaved in the following way:

1. If run locally, always compared the latest (staged) changes against the most recent commit.
2. If run in CI, would try to detect the latest commit of the base branch, and then compare changes on all files from the base branch up to the tip of the current PR. This ideally would check all files in all commits within a PR. It would accomplish this by trying to calculate the base branch via the github event metadata's "base_ref" field. Some of this logic was split between pre-commit-checks.yml and the run-clang-tidy script.

Unfortunately, this approach still allowed some changes to slip through the cracks and cause clang-tidy checks to fail even after a PR was merged because the logic on which changes/commits to check varied between PRs and push events

In order to unify the behavior, the following changes were implemented:

1. All PR checks use the same `pre-commit run --all-files`. This prevents having a split in logic for PRs/Merge Queue/Main branch in CI actions.
2. Updated the run-clang-tidy script to detect the environment to determine which files and subsections to check. Detection behaves in the following way:

- If inside github actions (`GITHUB_EVENT_PATH` env var defined), check the environment for the GH event type.
    - If a `push`, only compare the current commit to the previous commit. This should cover checks for recently-merged commits.
    - If a `pull_request`, use the `event->base->sha` from GH event metadata
    - If a `merge_group` (merge queue), use the base of the merge group, from GH event metadata
- If GH environment is not detected, then use "local" mode which falls back to searching for the "`git merge-base`" of the current branch `HEAD` which should be equivalent to the PR checks' `event->base->sha` 

### Performance Impact

N/A

### Release Note

N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/.github/workflows/pre-commit-checks.yml
+++ b/.github/workflows/pre-commit-checks.yml
@@ -59,7 +59,6 @@ jobs:
       - name: Generate Compile Database
         run: |
           make compile_db_all
-          cmake --build --preset conan-release --target generate_parquet_thrift
       - name: Run pre-commit hooks
         run: |
           pre-commit run --all-files
--- a/Makefile
+++ b/Makefile
@@ -266,7 +266,8 @@ _compile_db: conan_install
 	   -s "&:build_type=${BUILD_TYPE}" \
 	   -s build_type=$${DEPENDENCY_BUILD_TYPE:-${BUILD_TYPE}} \
 	   --build=missing $${ALL_CONAN_OPTIONS} && \
-	cd -
+	cd - && \
+	cmake --build --preset conan-$$(echo "${BUILD_TYPE}" | tr [A-Z] [a-z]) --target generate_parquet_thrift
 
 compile_db_all:
 	$(MAKE) _compile_db \


.github/workflows/pre-commit-checks.yml
```diff
@@ -59,12 +59,7 @@ jobs:
       - name: Generate Compile Database
         run: |
           make compile_db_all
+          cmake --build --preset conan-release --target generate_parquet_thrift
```
yangzhg: move this to make compile_db_all ?

ZacBlanco: updated

.github/workflows/pre-commit-checks.yml
```diff
@@ -61,10 +61,4 @@ jobs:
...
       - name: Run pre-commit hooks
         run: |
-          if [ "${{ github.event_name }}" == "pull_request" ]; then
-            echo "Detected Pull Request: Checking modified files only."
-            pre-commit run --from-ref origin/${{ github.base_ref }} --to-ref HEAD
-          else
-            echo "Detected Push/Merge: Checking all files."
-            pre-commit run --all-files
-          fi
+          pre-commit run --all-files
```
yangzhg, line 64: While the logic update to run-clang-tidy.py (handling diffs internally to support Merge Queues) is excellent, changing the CI command to pre-commit run --all-files globally introduces a significant regression for other hooks, specifically license-header-check and codespell.

Unlike the updated clang-tidy script (which now sets pass_filenames: false and calculates its own diff), the license-header-check hook still has pass_filenames: true in its config.
By running pre-commit run --all-files, the framework will pass every single file in the repository to the Go script and Codespell. Scanning thousands of files for license headers on every PR is unnecessary and slow.
Since License and Spell checks are robust enough for Merge Queues without special handling, We can split the execution in .github/workflows/pre-commit-checks.yml. We should keep the incremental check for standard hooks and only use the "all-files" (internal diff) mode for clang-tidy.
like 
```
      - name: Run pre-commit hooks
        run: |
          # 1. Run standard hooks (License, Codespell, etc.) on CHANGED files only
          if [ "${{ github.event_name }}" == "pull_request" ]; then
            SKIP=clang-tidy pre-commit run --from-ref origin/${{ github.base_ref }} --to-ref HEAD --skip clang-tidy
          else
            SKIP=clang-tidy pre-commit run --all-files
          fi

      - name: Run Clang-Tidy
        # 2. Run Clang-Tidy separately. 
        run: pre-commit run clang-tidy --all-files
```

yangzhg: Alternatively, you could update check_license.go to implement the same environment detection and git diff logic as run-clang-tidy.py. It would need to ignore the input file list from pre-commit and calculate the incremental changes internally.

ZacBlanco: I feel that the amount of time to check license headers and codespell is too small to worry about introducing this optimization to be more efficient. If it were on the order of minutes I wouldn't be as opposed, but when we introduce this, it makes it harder for developers to reproduce the result of a CI check when relying on `github.base_ref` inside the action script. By keeping it simple with `--all-files` for all checks it makes failures easier to reproduce

```
time pre-commit run license-header-check --all-files
Check for license headers................................................Passed

real    0m11.085s
user    4m36.126s
sys     0m28.680s

time pre-commit run codespell  --all-files
Spell Check..............................................................Passed

real    0m1.192s
user    0m15.746s
sys     0m1.431
```

clang-tidy takes an order of magnitude more time than both of these combined just for a single file in my testing

yangzhg: Thanks for the benchmark data. Since the overhead is relatively low (~12s), I am okay with moving forward with this PR to keep the workflow logic simple.

However, I do want to note that purely from an engineering standpoint, scanning the entire repository for every PR is computationally wasteful. Ideally, lightweight checks like license-check should be near-instant (sub-second) to maximize the "Fail Fast" benefit—canceling expensive jobs (like Clang-Tidy compilation) as early as possible if a simple check fails.

yangzhg (approved):
LGTM

.github/workflows/pre-commit-checks.yml
ZacBlanco: The "fail fast" benefit doesn't really apply here because we haven't configured [`fail_fast`](https://pre-commit.com/#pre-commit-configyaml---top-level) in any of our hooks. pre-commit will always run all hooks and provide the full output with our current configuration. I think the current behavior is reasonable for CI in order to show the author all failures.

Also, from my understanding, the hooks run serially, but parallel within each hook, so we don't really save resources unless `fail_fast` is enabled on every hook

e.g. clang-format will check all files in parallel, but further checks like shfmt, clang-tidy, license-check, etc do not run until all of the previous hook's processes have completed. This is a design decision by the pre-commit authors. Their reasoning being that some pre-commit hooks are expected to edit files, so in order to prevent race conditions on writes to a file, hooks must run serially.

I do agree that running on all files is more wasteful, I'm just more in favor of using all the allocated cores to help simplify checks and guarantee the codebase always fully conforms. This prevents changes which don't pass clang-tidy (or other pre-commit hooks) from somehow slipping through. It ensures the whole codebase _always_ conforms to our style guidelines, rather than letting it potentially go stale over time.

PR #234, opened 2026-02-13
fix: init rows with nulls in Paimon PartialUpdate

ZacBlanco:

In paimon tables, if a particular field is not initialized in any of the files in the split, it should be returned as null. In the previous implementation, fields were not set to null properly, so it cause reads of uninitialized memory which led to wrong query results and crashes.

### Type of Change

- [ ] 📝 Documentation only

### Description

As each new row is initialized, all fields must be assumed to be null. This change explicitly sets each field to null every time a row is added

### Performance Impact

### Release Note

```text
Release Note:
- Fix possible reads of uninitialized memory in Paimon PartialUpdate engine
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/connectors/hive/paimon_merge_engines/PartialUpdateEngine.cpp
+++ b/bolt/connectors/hive/paimon_merge_engines/PartialUpdateEngine.cpp
@@ -46,6 +46,10 @@ vector_size_t PartialUpdateEngine::add(PaimonRowIteratorPtr iterator) {
       }
     }
     result->resize(result->size() + 1);
+    const auto newRowIdx = result->size() - 1;
+    for (int i = 0; i < result->childrenSize(); i++) {
+      result->childAt(i)->setNull(newRowIdx, true);
+    }
     lastSequenceGroupKeyValues =
         std::vector<RowVectorPtr>(sequenceGroups_.size(), nullptr);
   }
--- a/bolt/dwio/paimon/reader/tests/PaimonReaderPartialUpdateTest.cpp
+++ b/bolt/dwio/paimon/reader/tests/PaimonReaderPartialUpdateTest.cpp
@@ -1201,5 +1201,139 @@ TEST_F(PaimonReaderPartialUpdateTest, sequenceGroupDefaultAggregateUpdate) {
   EXPECT_GT(getTableScanRuntimeStats(readTask)["totalMergeTime"].sum, 0);
 }
 
+TEST_F(PaimonReaderPartialUpdateTest, uninitializedMemoryTest) {
+  filesystems::registerLocalFileSystem();
+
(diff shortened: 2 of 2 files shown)


bolt/connectors/hive/paimon_merge_engines/PartialUpdateEngine.cpp
```diff
@@ -46,6 +46,10 @@ vector_size_t PartialUpdateEngine::add(PaimonRowIteratorPtr iterator) {
       }
     }
     result->resize(result->size() + 1);
+    const auto newRowIdx = result->size() - 1;
```
markjin1990, line 49: @ZacBlanco Do we need this fix in other engines?

ZacBlanco: I looked at the other engines, and I don't think this problem applies to them.

For the other paimon engines records are fully created and populated from the file reader. They are only aggregated or deleted based on the engine+primary key. So as long as the underlying reader gives the correct result (e.g. properly initialized rows), then this should not be a problem.

Partial update is special, because we have to aggregate the result of multiple file reads into a single row. So the row creation is actually occurring within the engine, which seems to have been done incorrectly.

So for example from the test in the PR

```
   * Row Schema: k (PK), v1, v2, v3 (never updated)
   *
   * Input 1: k=1, v1=10, v2=NULL, v3=NULL
   * Input 2: k=1, v1=NULL, v2=20, v3=NULL
   *
   * Expected: k=1, v1=10, v2=20, v3=NULL
```

The partial update engine receives multiple files in a single split, so `Input 1` comes from, say, `f1.parquet`. Then input 2 comes from `f2.parquet`. The files are sorted by PK, and then the aggregate engine will try to read from both iterators in tandem.

When the partial update engine gets a new PK for `k=1`, it creates a new row via `resize`, and then sets the children at index `size() - 1` to the values of the new row it just read. However, with partial update it's expected that rows are not "complete". NULLs are basicaly just ignored. Then, when new rows are read from each iterator, it populates the values for each child as long as the child value at that index is non-null.

Previously after reading `f1.parquet`, the result vector would look like:

```
0: k=1, v1=10, v2=<UNINIT MEMORY>, v3=<UNINIT MEMORY>
```

Then after reading `f2.parquet`

```
0: k=1, v1=10, v2=20, v3=<UNINIT MEMORY>
```

After this change, the result row looks like:

After reading f1.parqet:
```
0: k=1, v1=10, v2=NULL, v3=NULL
```

After reading f2.parqet:
```
0: k=1, v1=10, v2=20, v3=NULL
```

PR #243, opened 2026-02-18
Run benchmarks with RelWithDebInfo config instead of Release

erdembilegtbd:
Enabling perf runs to get performance counters.
Issue Number: close #xxx

### Type of Change

- [x ] 🔨 Refactoring (no logic changes)
- [ x] 🔧 Build/CI or Infrastructure changes
- [ ] 📝 Documentation only

### Description

During performance testing using benchmark builds, the release branch does not have symbols, hence perf can't resolve stack symbols.  Our CI system has benchmarks-build checks if that is buildable. So changing it incurred recurring costs.

Meanwhile, benchmark-build with debug info is useless, as benchmarks are not intended to run under debug mode. Hence changing that, and CI is not paying price at all.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Checklist (For Author)

- [ x] I have run clang-format / linters.
- [ x] No need to test or manual test.

### Breaking Changes

- [ x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,16 +337,16 @@ debug_spark_with_test:
 	$(MAKE) conan_build BUILD_TYPE=Debug BOLT_BUILD_TESTING="ON" CONAN_OPTIONS="-o bolt/*:spark_compatible=True -o bolt/*:enable_testutil=True"
 
 benchmarks-basic-build:
-	$(MAKE) conan_build BUILD_TYPE=RelWithDebInfo  BOLT_BUILD_BENCHMARKS_BASIC="ON" CONAN_OPTIONS="-o bolt/*:spark_compatible=True -o bolt/*:enable_testutil=True -o bolt/*:enable_perf=True"
+	$(MAKE) conan_build BUILD_TYPE=Release BOLT_BUILD_BENCHMARKS_BASIC="ON" CONAN_OPTIONS="-o bolt/*:spark_compatible=True -o bolt/*:enable_testutil=True -o bolt/*:enable_perf=True"
 
 benchmarks-build:
-	$(MAKE) conan_build BUILD_TYPE=RelWithDebInfo BOLT_BUILD_BENCHMARKS="ON" CONAN_OPTIONS="-o bolt/*:spark_compatible=False -o bolt/*:enable_testutil=True -o bolt/*:enable_perf=True"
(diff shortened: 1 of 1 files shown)


markjin1990:
@ZacBlanco Have you seen OOM issues before in CI on github?   https://github.com/bytedance/bolt/actions/runs/22160784294/job/64076708040 
image

erdembilegtbd:
@ZacBlanco  The run is going OOM. Any way to add memory to those machines?

ZacBlanco:
I have seem OOMs before, but they tend to be rare. The failure rate of `build-only` jobs is ~2.5% with most of those being due to compilation issues. OOMs are more rare. We can tune down the number of concurrent jobs in CI with `CI_NUM_THREADS`, but the failure rate due to OOM is low enough that I think the better solution is to just re-run it if it does OOM. It looks like these all passed on the 2nd run so it should be ok. If you see it occur more frequently you can let me know and I will take a closer look at tuning the memory limits on the containers

ZacBlanco (changes requested):
I would recommend adding a new line here below the following one:

https://github.com/bytedance/bolt/blob/a2caa7c8d704565ee03856705f049d5c39f1394d/.github/workflows/build-test.yml#L115

```
+           echo 'tools.cmake.cmaketoolchain:extra_variables={"CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O2"}' >> $(conan profile path default) 
```

debug info creates huge binaries and the CI machines can't handle it due to very limited disk bandwidth. It also increases the link time significantly. Since the binaries are in CI, they don't actually need debug info. We just need to verify the build works. Adding this line will just use the same compiler options as normal `RelWithDebInfo`, but without the `-g` so that we save some disk bandwidth.

erdembilegtbd:
Whole idea is to get the debug info so that performance tunings can solve function symbols. I will work on the different ways.

ZacBlanco:
The change will only affect CI builds. We don't make artifacts from CI builds available, so it's OK if they don't have the debug info

erdembilegtbd:
Our CI system has benchmarks-build checks if that is buildable. So changing it incurred recurring costs.

Meanwhile, benchmark-build with debug info is useless, as benchmarks are not intended to run under debug mode. Hence changing that, and CI is not paying price at all. @ZacBlanco

Issue #250, opened 2026-02-22
[Feature][Design] Refactor Bolt connectors to remove Hive coupling and prepare for pluggable connectors (Bolt) Background


yingsu00:
### Feature Category

Connector / Storage (Hive, Iceberg, etc.)

### Problem / Use Case

Bolt currently relies on the Hive connector as the primary storage interface.
However, many parts of Bolt — especially planning utilities, exec helpers, and tests — still directly reference Hive classes such as:

- HiveTableHandle
- HiveColumnHandle
- HiveInsertTableHandle
- HivePartitionFunctionSpec
- LocationHandle
- Hive-specific split and metadata structures

This makes Hive effectively a hard dependency of the engine, not just a connector.
As a result:
- New connectors (Iceberg, Lakehouse formats, internal storage systems) are difficult to integrate
- Exec tests cannot run without Hive
- Planning helpers like PlanBuilder are not connector-agnostic
- Small Hive changes force wide rebuilds and risk regressions
- Bolt’s storage layer cannot evolve independently of Hive
We need to refactor this before introducing additional connectors.

# Goals

1. Make Velox exec/planning code connector-agnostic
2. Remove direct Hive type usage outside the Hive connector
3. Introduce a factory layer so connectors supply their own objects
4. Preserve current behavior for Hive (no functional changes)
5. Prepare the codebase so Iceberg or other connectors can be added later with minimal disruption

### Proposed Solution

# Proposed Refactor
## 1. Introduce ConnectorObjectFactory

Each connector registers a factory responsible for constructing connector-specific objects:
- makeTableHandle()
- makeColumnHandle()
- makeInsertTableHandle()
- makeLocationHandle()
- makePartitionFunctionSpec()
- makeConnectorSplit()

Core code should never instantiate connector classes directly anymore. Instead of directly creating a HiveTableHandle, it calls:
```cpp
auto factory = ConnectorRegistry::getFactory(connectorName);
auto tableHandle = factory->makeTableHandle(...);
```
This moves connector knowledge out of exec and planning layers.

## 2. Refactor PlanBuilder to be connector-aware

Today:

- tableScan() directly constructs Hive handles
- tableWriter() constructs Hive column handles, bucket properties, location handles
- localPartition() directly uses HivePartitionFunctionSpec

We need to:
- Add connectorId to PlanBuilder (default = "hive")
- Replace all Hive object creation with factory calls
- Pass configuration via folly::dynamic options instead of Hive structs

This allows:

[...]

PR #251, opened 2026-02-22
Make WriterOptions serializable

yingsu00:

First PR for https://github.com/bytedance/bolt/issues/250

### Type of Change

- [ ] 📝 Documentation only

### Description

Make WriterOptions serializable. This is required for future refactors

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/.github/actions/bolt-build-base/action.yml
+++ b/.github/actions/bolt-build-base/action.yml
@@ -13,6 +13,7 @@ runs:
         tools.build:exelinkflags=['-fuse-ld=mold', '-Wl,--allow-multiple-definition']
         tools.build:sharedlinkflags=['-fuse-ld=mold', '-Wl,--allow-multiple-definition']
         EOF
+        echo "core.download:parallel=16" >> ~/.conan2/global.conf
 
     - name: Add CI conan remote
       shell: bash
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -1,0 +1,55 @@
+# Copyright (c) ByteDance Ltd. and/or its affiliates.
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Build and Test
+on:
+  push:
+    branches: [ main ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CCACHE_DIR: /data/ccache-data
+  CCACHE_MAX_SIZE: '100G'
+  CI_NUM_THREADS: "16"
+  IN_CI: '1'
+
+jobs:
+  build-only:
+      runs-on: [ self-hosted, medium ]
+      container:
+        image: bolt-registry:5000/bolt-ci:20260114
+        volumes:
+          - /data/ccache-data:/data/ccache-data
+        options: --user root --init
+      services:
+        conanserver:
+          image: bolt-registry:5000/conan-server:latest
+          volumes:
+            - /data/conan-server-data:/var/conan/data
+      steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - id: env-setup
+        uses: ./.github/actions/bolt-build-base
+      - name: Bolt build
+        run: |
+          make release_spark_with_test
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -19,8 +19,6 @@
 
 name: Build and Test
 on:
-  push:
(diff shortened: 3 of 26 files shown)


yingsu00:
@ZacBlanco Could you please review this? Thanks!

.pre-commit-config.yaml
```diff
@@ -77,20 +76,6 @@ repos:
         entry: go run -C .github/license_check check_license.go
         language: golang
         pass_filenames: true
```
ZacBlanco, line 79: Can you revert the changes to this file?

bolt/dwio/common/Options.cpp
```diff
@@ -98,4 +99,117 @@ ColumnReaderOptions makeColumnReaderOptions(const ReaderOptions& options) {
...
+    opts->spillConfig = opts->ownedSpillConfig.get();
+  }
+
+  // pool, nonReclaimableSection, and spillConfig callbacks/executor remain at
+  // default and must be re-injected by the host.
+  return opts;
+}
+
+void WriterOptions::registerSerDe() {
+  bolt::Type::registerSerDe();
```
ZacBlanco, line 195: Why do we call `Type::registerSerDe()` in this method?

bolt/dwio/common/Options.cpp
```diff
@@ -98,4 +99,117 @@ ColumnReaderOptions makeColumnReaderOptions(const ReaderOptions& options) {
...
+
+  // 7) spillConfig
+  if (auto p = obj.get_ptr("spillConfig")) {
+    opts->ownedSpillConfig =
+        ISerializable::deserialize<bolt::common::SpillConfig>(*p);
+    opts->spillConfig = opts->ownedSpillConfig.get();
+  }
+
+  // pool, nonReclaimableSection, and spillConfig callbacks/executor remain at
+  // default and must be re-injected by the host.
```
ZacBlanco: This comment might be better suited to be documentation at the top of the method?

bolt/dwio/common/Options.cpp
```diff
@@ -98,4 +99,117 @@ ColumnReaderOptions makeColumnReaderOptions(const ReaderOptions& options) {
...
+    obj["zlibCompressionLevel"] = static_cast<int64_t>(*zlibCompressionLevel);
+  }
+
+  // 7) spillConfig (value fields only; callbacks/executor re-injected by host)
+  if (spillConfig) {
+    obj["spillConfig"] = spillConfig->serialize();
+  }
+
+  // (We do *not* serialize pool, nonReclaimableSection, or the callbacks /
+  //  executor inside spillConfig—they must be re‐injected by the host.)
```
ZacBlanco: Same thought about location of this comment

bolt/dwio/common/tests/OptionsTests.cpp
```diff
@@ -55,3 +59,127 @@ TEST(OptionsTests, testAppendRowNumberColumnInCopy) {
...
+
+TEST(OptionsTests, writerOptionsDefaultsRoundTrip) {
+  WriterOptions opts;
+
+  folly::dynamic dyn = opts.serialize();
+  auto roundTrip = WriterOptions::deserialize(dyn);
+
+  ASSERT_EQ(nullptr, roundTrip->schema);
+  ASSERT_FALSE(roundTrip->compressionKind.has_value());
+  ASSERT_TRUE(roundTrip->serdeParameters.empty());
```
ZacBlanco, line 133: Should we assert on the rest of the fields/defaults too?

bolt/dwio/common/Options.cpp
yingsu00: > Why do we call `Type::registerSerDe()` in this method?

WriterOptions::deserialize (Options.cpp:157) calls:
```
  opts->schema = ISerializable::deserialize<bolt::Type>(*p);
```
I know other Classes like `HiveColumnHandle::registerSerDe()` didn't add this (HiveColumnHandle also has TypePtr members), but it relies on every caller to call `Type::registerSerDe()` pre-hand. I don't think it's a good pattern to follow. It's a hidden dependency that the caller must know the internal implementation detail that HiveColumnHandle::create calls ISerializable::deserialize<Type>. That leaks the internals and makes the API easy to misuse: forgetting Type::registerSerDe() causes a silent runtime failure, not a compile error. IMHO The better design is for registerSerDe() to chain its own dependencies internally, which is exactly what `WriterOptions::registerSerDe()` does with `bolt::Type::registerSerDe()`. The call is idempotent (it just re-registers the same entry), so there's no harm in calling it redundantly.

I forgot to add `registry.Register("SpillConfig", SpillConfig::deserialize);` last time. In this update it was added too.

ZacBlanco: I'm not a huge fan of this pattern where the serializers and deserializers need to have their dependencies explicitly written in the registration methods. I am open to re-designing it at a later point though. For this PR it is fine.

I would probably prefer if we had something like a static initialization method that was executed at library load time/program start time. Registering the serializers/deserializers automatically makes maintenance a little bit easier on developers. Also, this way as long as the necessary library(s) are loaded/compiled properly all the registry entries should exist. There are some downides to this approach too but we can discuss later. This is fine for now

frankobe:
@yingsu00 FYI I update your branch with `main` to pick up some CI improvement to ensure test stability.

yingsu00:
@ZacBlanco 
> I would probably prefer if we had something like a static initialization method that was executed at library load time/program start time.

I agree. Not only the serializers, but the different file formats, and connectors shall also register themselves when loaded. In https://github.com/facebookincubator/velox/pull/14090 I tried to make file reader/writers register themselves using a static initialization function. This had some issue on Linux, but we can revisit them again.

PR #211, opened 2026-02-06
chore: switch conan index to bytedance/conan-index-center

kexianda:
### What problem does this PR solve?

* it is easier to maintain the conan recipes in a conan recipes repo
* The upstream conan-io/conan-center-index does NOT have any release tag. What's more, it keeps removing the old versions support of dependencies. For Bolt, it is not maintainable. Now we fork conan-io/conan-center-index and will create releases tags for Bolt.

Issue Number: close #210

### Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🚀 Performance improvement (optimization)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🔨 Refactoring (no logic changes)
- [x] 🔧 Build/CI or Infrastructure changes
- [ ] 📝 Documentation only

### Description

Describe your changes in detail.
For complex logic, explain the "Why" and "How".

### Performance Impact

- [ ] **No Impact**: This change does not affect the critical path (e.g., build system, doc, error handling).
- [ ] **Positive Impact**: I have run benchmarks.
    <details>
    <summary>Click to view Benchmark Results</summary>

    ```text
    Paste your google-benchmark or TPC-H results here.
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>
- [ ] **Negative Impact**: Explained below (e.g., trade-off for correctness).

### Release Note

Please describe the changes in this PR

Release Note:

```text
Release Note:
- Fixed a crash in `substr` when input is null.
- optimized `group by` performance by 20%.
```

### Checklist (For Author)

- [ ] I have added/updated unit tests (ctest).
- [ ] I have verified the code with local build (Release/Debug).
- [ ] I have run clang-format / linters.
- [ ] (Optional) I have run Sanitizers (ASAN/TSAN) locally for complex C++ changes.
- [ ] No need to test or manual test.

### Breaking Changes

- [ ] No
- [ ] Yes (Description: ...)
    <details>
    <summary>Click to view Breaking Changes</summary>

    ```text
    Breaking Changes:
    - Description of the breaking change.
    - Possible solutions or workarounds.
    - Any other relevant information.
    ```
    </details>

PR #357, opened 2026-03-06
feat: Add Paimon connector

### LINKED ISSUE #14: Support integration with Paimon C++ SDK
Bolt has already supported Paimon table format at present, which implements reading and writing Paimon by itself. Paimon C++ SDK would release in the near future that would better to be integrated in Bolt.

ZacBlanco:

Issue Number: Fixes #14 

### Type of Change

- [ ] 📝 Documentation only

### Description

This branch adds a **Paimon connector** to Bolt — enabling Bolt's vectorized execution engine to read tables from [Apache Paimon](https://paimon.apache.org/). It bridges two ecosystems: **paimon-cpp** (the C++ implementation at https://github.com/alibaba/paimon-cpp) of the official table format library (https://github.com/apache/paimon) and **Bolt**.

#### How it works

**Read path**: When a query references a Paimon table, the connector receives the table metadata and split information as serialized bytes from the Apache Paimon library (See `PaimonConnectorSplit` - the serialized format is defined by Apache Paimon.). The connector then constructs a paimon-cpp `FileBatchReader` which uses Bolt's Parquet reader under the hood, and streams data back through the Arrow C-data interface. It supports full-schema reads, column projection, and filter pushdown.

**Filter pushdown**: In order to support filter pushdown, we need to translate from bolt expression filters into paimon-cpp-compatible expressions. Then, from Paimon-cpp compatible expressions we must translate back into bolt expressions for the Parquet reader.

We translate paimon-cpp's predicate tree into Bolt's `TypedExpr` system so filters are pushed down to the Parquet column reader level. Supports equality, ranges, IN-lists, IS NULL, NOT, AND/OR.

**HdfsFileSystem**: Added a `stat()` method and `HdfsFileInfo` struct to Bolt's HDFS filesystem interface, returning file metadata (isDirectory, size, modification time). This was needed by the paimon-cpp filesystem bridge (`PaimonBoltHdfsFileSystem`) which implements paimon-cpp's `FileSystem` abstraction on top of Bolt's HDFS support.

A basic diagram is provided here:

```
                              BOLT QUERY ENGINE
                             (planning / execution)
                                      |
                    +-----------------+------------------+
                    |                 |                  |
             Table Handle         Split Data          Filter
             (metadata &         (read region       (predicate
              location)          from paimon)          AST)
                    |                 |                  |
                    +-----------------+------------------+
                                      |
                         +-------------------------+
                         |    PaimonDataSource     |  <--- BOLT
                         | (connector entry point) |
                         +------------+------------+
                                      |
                +---------------------+---------------------+
                |                     |                     |
             Reader                Filters                Config
             Builder              Translator             Options
                |                     |                     |
                v                     v                     v
           +----------------------------------------------------+
           |            PaimonParquetFileBatchReader            |  <--- PAIMON-CPP
           |            (FileBatchReader interface)             |
           +--------------------------+-------------------------+
                                      |
                                      v
           +----------------------------------------------------+
           |                PaimonParquetReader                 |  <--- BOLT
           |             (wraps Bolt ParquetReader)             |
           +--------------------------+-------------------------+
                                      |
                +---------------------+---------------------+
                |                     |                     |
            RowReader              Parquet                Memory
             (Bolt)                Column               Pool Bridge
                                   Reader
                |                     |                     |
                v                     v                     v
             +--+--+               +--+--+               +--+--+
             |     |               |     |               |     |
           Rows  Scan             I/O  Config          Memory I/O
          (Bolt)(Bolt)           (Bolt)(Bolt)          (Bolt)(Bolt)
```

#### Build & dependency changes

Adds a **Conan recipe** for paimon-cpp that patches its vendored dependencies to use Conan-managed ones (arrow, fmt, glog, rapidjson, etc.), and builds it as a shared library. The root `conanfile.py` is updated to include this recipe. A Python package check (`pypaimon`) runs at CMake configure time for integration tests.

#### Test suite

~126 tests across unit, translation, and integration levels:

- **Unit**: Config parsing, filter translator round-trips (every predicate type + cast-wrapped literals), parquet reader round-trips (primitive/complex types, timestamp precision modes, row number tracking)
- **Integration**: Full end-to-end table scans covering append-only tables, PK tables (with/without overwrite), data evolution, partial updates, deduplication, aggregation, and filter pushdown across all predicate types

### Performance Impact

This is a new connector, so we don't affect existing codepaths.

### Release Note

Release Note:

```text
Release Note:
- Add Paimon connector  backed by the Paimon-cpp library
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)


bolt/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp
```diff
@@ -314,4 +314,30 @@ void HdfsFileSystem::rmdir(std::string_view path) {
       impl_->hdfsShim()->GetLastExceptionRootCause());
 }
 
+HdfsFileSystem::HdfsFileInfo HdfsFileSystem::stat(std::string_view path) const {
+  // Only remove the scheme for hdfs path.
+  if (path.find(kScheme) == 0) {
+    path.remove_prefix(kScheme.length());
+    if (auto index = path.find('/')) {
+      path.remove_prefix(index);
```
NicolasDenoyelle, line 322: Feel free to skip (in general) if this is too nitpicky.

I'm wondering if the following example would be a bug: hdfs://path/to where kScheme == 'hdfs:// into pathto?
If kScheme is hdfs: I can see how it becomes /path/to but is this always true that kScheme is hdfs: and that the follow-up string is //?
Perhaps the existing code is doing the same things but it looks fragile.

ZacBlanco:
For reviewers:

This is a large PR since it adds a whole new connector and test suite. Here is my suggestion for going about the review:

I would first focus on reviewing everything **outside** of `bolt/connectors/paimon` - as this is mostly files that are touching existing code and has the highest likelihood to impact existing workloads. The LoC changed outside of the connector is relatively small.

After checking that, I would start to look at the new connector code. 

Let me know if you have questions or would like a walkthrough about the architecture.

scripts/conan/recipes/paimon-cpp/all/patches/paimon-cpp-conan-deps.patch
```diff
@@ -0,0 +1,929 @@
+diff --git a/.gitignore b/.gitignore
```
NicolasDenoyelle, line 1: This patch looks like it is bigger than the initial package.
Should we make a dedicated package instead?

scripts/conan/recipes/paimon-cpp/all/conanfile.py
```diff
@@ -0,0 +1,264 @@
+from conan import ConanFile
```
NicolasDenoyelle, line 15: Shouldn't we host package recipes in our recipe repository and in our hosted artifactory? I am not aware of how we manage packages for OSS so I might be completely wrong.

bolt/connectors/paimon/tests/HdfsContainerMiniCluster.h
```diff
@@ -0,0 +1,51 @@
+/*
```
NicolasDenoyelle, line 1: I wonder if this would suit better a bash script?

bolt/connectors/paimon/PaimonFilterTranslator.h
```diff
@@ -0,0 +1,175 @@
...
+class Literal;
+enum class FieldType : int;
+} // namespace paimon
+
+namespace bytedance::bolt::connector::paimon {
+
+/// Templated result type for filter translation operations.
+/// Carries either a successful value or a diagnostic reason string on failure.
+template <typename T>
+struct TranslationResult {
```
NicolasDenoyelle: If you intend to use this pattern elsewhere, consider using tl/expected: https://github.com/TartanLlama/expected

NicolasDenoyelle (commented):
I looked at the non paimon files and all of the headers so far.
Everything looks clean.

bolt/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp
ZacBlanco: I agree that the code feels a little hacky, but this is the exact same pattern as every other method in this client, so I just opted to maintain the pattern

scripts/conan/recipes/paimon-cpp/all/patches/paimon-cpp-conan-deps.patch
ZacBlanco: Not sure what you mean by "bigger than the initial package"? But yes the patch is large. It was non-trivial to switch the paimon-cpp library to build using conan dependencies

Right now, we don't have an external conan recipes index dedicated to bolt. Xianda set up https://github.com/bytedance/conan-center-index which we will eventually use once #211 is merged but that task hasn't been complete yet, so the only way to make the recipe available to bolt (in our current configuration) is to place the whole recipe in this `scripts/conan/recipes` directory.

scripts/conan/recipes/paimon-cpp/all/conanfile.py
ZacBlanco: See my comment [here](https://github.com/bytedance/bolt/pull/357#discussion_r3155742006) - we store additional bolt recipes under `scripts/conan/recipes` for now

bolt/connectors/paimon/tests/HdfsContainerMiniCluster.h
ZacBlanco: Bash script would definitely make it more portable and faster to iterate. I need to think about the best way to go about it. One concern is the best way to portably look for/find the script when actually running the test, and also ensuring the container is properly cleaned up afterwards

bolt/connectors/paimon/PaimonFilterTranslator.h
ZacBlanco: Oh I really like this. C++ doesn't have enough functional programming concepts in their APIs. I will see if I can incorporate it.

NicolasDenoyelle (commented):
I looked at the CI and I was not sure if your unittests are really running.
For instance I searched `PaimonConfigTest` in both `build-test` steps under "Run tests" and I did not find it.

How can I check that all the tests are running?

scripts/conan/recipes/paimon-cpp/all/patches/paimon-cpp-conan-deps.patch
NicolasDenoyelle: Got it. Sounds good, let's wait to get #211 merged then. Perhaps you can create a follow-up ticket so with low priority so we have it somewhere.

ZacBlanco:
All the tests are linked into one executable, `bolt_paimon_connector_test`: https://github.com/ZacBlanco/bolt/blob/54effb23134999130fbdea1c2e59929d5f2b0f52/bolt/connectors/paimon/tests/CMakeLists.txt#L35-L46

You can search for `bolt_paimon_connector_test` in the test log and see that it passes.

e.g. in https://github.com/bytedance/bolt/actions/runs/25008570740/job/73238030931?pr=357

```
...
Start 391: bolt_paimon_connector_test
...
383/408 Test #391: bolt_paimon_connector_test .......................................................................   Passed    2.32 sec
...
```

You won't see output for the individual test classes unless a test fails

scripts/conan/recipes/paimon-cpp/all/patches/paimon-cpp-conan-deps.patch
ZacBlanco: There are a bunch of other recipes already that would need to be migrated, so I think Xianda will take care of it with #211 . It's just going to be a `cp -r /scripts/recipes $CCI_FORK/recipes`

bolt/connectors/paimon/PaimonFilterTranslator.h
NicolasDenoyelle: I think that starting from C++23, expected has been merged in the c++ standard library.

bolt/connectors/paimon/benchmark/create_benchmark_tables.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env python3
```
NicolasDenoyelle, line 1: I know it may break the nit directory separation between tests and benchmarks, but I can't help to feel that we may be able to regroup the table creation code into a single python script for tests and benchmarks.

NicolasDenoyelle: Since you are using the test libs in the CMakeFile.txt, I thought that you could store this in tests as well.

NicolasDenoyelle (approved):
LGTM

scripts/conan/recipes/paimon-cpp/all/conandata.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+  "0.0.3-bolt":
```
markjin1990, line 16: I see the latest version of paimon-cpp is https://github.com/alibaba/paimon-cpp/tree/v0.1.3. Why we are not referencing this version?

bolt/connectors/paimon/tests/.gitignore
```diff
@@ -0,0 +1 @@
+test_warehouse/
```
markjin1990, line 1: is this necessary change?

bolt/connectors/paimon/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) ByteDance Ltd. and/or its affiliates.
```
markjin1990, line 1: Is this a correct copyright notice?

scripts/conan/recipes/paimon-cpp/all/conandata.yml
ZacBlanco: `0.0.3` is actually just the scheme I came up with. It is the 3rd iteration of the paimon-cpp dep I have built/uploaded to our conan repository. I have iterated over a few versions of paimon-cpp now. When I initially started this project, paimon-cpp did not actually have any tags or versions. And for future builds internally, I don't think we intend to wait for release tags of the paimon-cpp project to pull in updates. 

The current version is based off of a recent commit (maybe ~1 week old?) that I tested with (see `conandata.yml`).

bolt/connectors/paimon/tests/.gitignore
ZacBlanco: It was intentional. Without the gitignore and depending on how you run the test binary the `test_warehouse` directory can pollute the git tree with a bunch of paimon tables and a hundred or so files. To keep `git status` and git client output easier to manage I just added it to gitignore.

bolt/connectors/paimon/benchmark/CMakeLists.txt
ZacBlanco: It passes the license check...so it should be correct. If you think it is wrong I can take a look at the license check configuration

markjin1990: Here is the license header we normally use for new files: https://github.com/bytedance/bolt/blob/main/bolt/benchmarks/paimon/PaimonBenchmark.cpp#L1-L15

ZacBlanco: Thanks for pointing out the issue. License check bug is now fixed and licenses across this PR have been updated

markjin1990 (approved):

bolt/connectors/paimon/PaimonConnector.h
```diff
@@ -0,0 +1,77 @@
...
+      std::shared_ptr<ConnectorQueryCtx> queryCtx,
+      const core::QueryConfig& queryConfig) override;
+
+  std::unique_ptr<DataSink> createDataSink(
+      std::shared_ptr<const RowType> /*inputType*/,
+      std::shared_ptr<ConnectorInsertTableHandle> /*insertHandle*/,
+      ConnectorQueryCtx* /*queryCtx*/,
+      CommitStrategy /*commitStrategy*/,
+      const core::QueryConfig& /*queryConfig*/) override {
+    return nullptr;
```
oarap: Nit. this is a bit odd and risky. If we don't expect this to be called better to add a BOLT_FAIL or something.

bolt/connectors/paimon/PaimonFilterTranslator.cpp
```diff
@@ -0,0 +1,1641 @@
...
+      auto ts = value.value<bytedance::bolt::Timestamp>();
+      int64_t totalNanos = ts.getSeconds() * 1'000'000'000LL + ts.getNanos();
+      int64_t millis = totalNanos / 1'000'000;
+      int32_t nanoOfMillis = static_cast<int32_t>(totalNanos % 1'000'000);
+      auto paimonTs = ::paimon::Timestamp(millis, nanoOfMillis);
+      return ::paimon::Literal(paimonTs);
+    }
+    case ::paimon::FieldType::DECIMAL:
+      // For decimal types, extract as double.
+      return ::paimon::Literal(value.value<double>());
```
oarap: I got this from LLM:

Bolt stores decimals as `int64/int128`, not `double`. `value.value<double>()` will throw on a real DECIMAL variant. Tests don't appear to exercise DECIMAL pushdown. 

Action: drop the case (return nullopt) or implement DECIMAL→paimon::Literal properly.

Suggested change:
```
case ::paimon::FieldType::DECIMAL: {
  // Bolt stores short decimals as int64, long decimals as int128.
  if (value.kind() == TypeKind::BIGINT) {
    return ::paimon::Literal(value.value<int64_t>());  // if paimon::Literal supports it
  }
  // HUGEINT/int128 → would need a paimon-cpp Literal constructor that accepts int128
  return std::nullopt;  // skip pushdown rather than crash
}
```

oarap (approved):
Overall looks good to me, added 2 comments just for you to check.

bolt/connectors/paimon/PaimonConnector.h
ZacBlanco: Good catch. Thanks!

oarap (approved):

PR #359, opened 2026-03-07
[feature] support casting TIMESTAMP to integer

### LINKED ISSUE #358: [Feature] Casting TIMESTAMP to BIGINT is not supported yet
### Feature Category

SQL Function / Operator

### Problem / Use Case

Casting TIMESTAMP to BIGINT is not supported by Bolt yet.
The following SQL
`SELECT 
    original_ts,
    CAST(original_ts AS TIMESTAMP) AS timestamp,
    CAST(CAST(original_ts AS TIMESTAMP) AS BIGINT) AS unix_seconds
FROM (
    VALUES 
        ('1940-01-02 00:00:00'),
) AS t(original_ts);`
will return error `Reason: unsupported type conversion from TIMESTAMP to BIGINT`

### Proposed Solution

Add conversion from TIMESTAMP to integer in `bolt/expression/CastExpr-tpl.cpp`.

### References / Prior Art

_No response_

### Importance

High (Needed for production)

### Willingness to Contribute

Yes, I can submit a PR

markjin1990:
Issue Number: close #358

### Type of Change

- [ ] 📝 Documentation only

### Description
Add support for FROM Timestamp TO integer cases in Converter in `CastExpr-tpl.cpp`

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
- Support casting TIMESTAMP to integer
```

### Checklist (For Author)

### Breaking Changes

- [ ] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/expression/CastExpr-tpl.cpp
+++ b/bolt/expression/CastExpr-tpl.cpp
@@ -25,6 +25,7 @@
 #define bswap_64(x) __builtin_bswap64(x)
 #endif
 
+#include <algorithm>
 #include "bolt/functions/InlineFlatten.h"
 #include "bolt/functions/lib/RowsTranslationUtil.h"
 #include "bolt/functions/lib/StringUtil.h"
@@ -475,24 +476,17 @@ class Converter {
       // Timestamp -> integral seconds, rounding away from zero.
       int64_t secs64 = 0;
       try {
-        __int128_t micros = static_cast<__int128_t>(from.toMicros());
-        __int128_t secs128 =
-            micros / (__int128_t)1'000'000; // trunc toward zero
-        __int128_t rem = micros % (__int128_t)1'000'000;
-        if (rem != 0) {
-          if (micros > 0) {
-            secs128 += 1; // away from zero for positive fraction
-          } else {
-            secs128 -= 1; // away from zero for negative fraction
-          }
-        }
-        if (secs128 > (__int128_t)std::numeric_limits<int64_t>::max()) {
-          secs64 = std::numeric_limits<int64_t>::max();
-        } else if (secs128 < (__int128_t)std::numeric_limits<int64_t>::min()) {
-          secs64 = std::numeric_limits<int64_t>::min();
-        } else {
-          secs64 = static_cast<int64_t>(secs128);
-        }
+        int128_t micros = static_cast<int128_t>(from.toMicros());
+        int128_t secs128 =
+            micros / static_cast<int128_t>(1'000'000); // trunc toward zero
+        int128_t rem = micros % static_cast<int128_t>(1'000'000);
+        // sec128 + 1 when rem != 0 and micros > 0, sec128 - 1 when rem != 0
+        // and micros < 0
+        secs128 += (rem != 0) * ((micros > 0) - (micros < 0));
+        secs64 = std::clamp(
+            secs128,
+            static_cast<int128_t>(std::numeric_limits<int64_t>::min()),
+            static_cast<int128_t>(std::numeric_limits<int64_t>::max()));
       } catch (...) {
         return ConvertStatus::OTHER_FAILURE;
       }


markjin1990:
@guhaiyan0221 @fzhedu please review this PR when you are available.

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -471,6 +471,33 @@ class Converter {
       }
     } else if constexpr (fromBool) {
       return tryToWithFolly(from, to);
+    } else if constexpr (fromKind == PrimitiveKind::TIMESTAMP) {
+      // Timestamp -> integral seconds, rounding away from zero.
+      int64_t secs64 = 0;
+      try {
+        __int128_t micros = static_cast<__int128_t>(from.toMicros());
```
ZacBlanco: I see some places use `__int128_t` and `int128_t`. Was there any particular reason to choose this variant?

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -471,6 +471,33 @@ class Converter {
...
+        __int128_t secs128 =
+            micros / (__int128_t)1'000'000; // trunc toward zero
```
ZacBlanco: I think clang-tidy says we should prefer `static_cast`. I also think it makes it easier to read.
```suggestion
            micros / static_cast<__int128_t>(1'000'000); // trunc toward zero
```

This comment also applies to other uses of C-style casts in this PR

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -471,6 +471,33 @@ class Converter {
...
+        __int128_t rem = micros % (__int128_t)1'000'000;
+        if (rem != 0) {
+          if (micros > 0) {
+            secs128 += 1; // away from zero for positive fraction
+          } else {
+            secs128 -= 1; // away from zero for negative fraction
+          }
+        }
```
ZacBlanco: To avoid branch-misprediction penalty, you could theoretically convert this into a single statement. But keep a comment maybe to describe the desired behavior?

```cpp
secs128 += (rem != 0) * ((micros > 0) - (micros < 0));
```

You can compare the output assembly here:

https://godbolt.org/z/761jxcnr8

There is technically one more instruction in the non-conditional version when `rem != 0` is true, but no branch misprediction penalties. The version above also requires executing all instructions for every input

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -471,6 +471,33 @@ class Converter {
...
+        if (secs128 > (__int128_t)std::numeric_limits<int64_t>::max()) {
+          secs64 = std::numeric_limits<int64_t>::max();
+        } else if (secs128 < (__int128_t)std::numeric_limits<int64_t>::min()) {
+          secs64 = std::numeric_limits<int64_t>::min();
+        } else {
+          secs64 = static_cast<int64_t>(secs128);
+        }
```
ZacBlanco: Could we use `std::clamp` here?

https://en.cppreference.com/w/cpp/algorithm/clamp.html

bolt/expression/CastExpr-tpl.cpp
ZacBlanco: Actually, interestingly enough, compilation with clang using -O2 results in nearly the same assembly for both versions, but GCC seems to still use a `jmp`

bolt/expression/CastExpr-tpl.cpp
markjin1990: Oh, they are the same. You can see the definition "using int128_t = __int128_t;" in `HugeInt.h`. I can switch to  int128_t.

PR #371, opened 2026-03-10
fix(ci): reduce contention on conan download in build

ZacBlanco:

CI sometimes fails with the error that conan fails to acquire a database lock in 20 seconds. This tries to reduce contention to reduce the occurrence of this error.

### Type of Change

- [ ] 📝 Documentation only

### Description

Reduces the number of parallel downloads to just 8 (from 16) to reduce the frequency at which the conan error occurs. It's likely the CI machines are under some higher load when this happens, so by reducing the amount of work the remote server has to do this should hopefully reduce the frequency.


frankobe (approved):
Thx for the effort

PR #373, opened 2026-03-11
[fix] Fix LRU cache eviction UB when entries have external references

frankobe:

Issue Number: https://github.com/bytedance/bolt/discussions/151

The JIT LRU cache eviction loop in `LRUCache::insert()` can call `evict()` on an empty list, causing undefined behavior (SIGABRT/SEGFAULT).

When a cached module has an external `CompiledModuleSP` reference (e.g. held by a concurrent thread via `LookupSymbolsInCache`), eviction removes the cache entry but the `CompiledModuleImpl` isn't destroyed (refcount > 0), so `notifyFreeingObject` is never called and memory usage doesn't decrease. The eviction loop keeps calling `evict()` until the cache is empty, then dereferences `--list_.end()` on an empty list — undefined behavior.

### Type of Change

- [ ] 📝 Documentation only

### Description

**Root cause**: `LRUCache::insert()` has an eviction loop `while (evictPolicy_()) { evict(); }` with no empty-check. The `evictPolicy_` predicate (`CacheEvictPred`) checks `GetMemoryUsage() > GetMemoryLimit()`. When evicted modules still have external shared_ptr references, their destructors don't run, `rt->remove()` isn't called, and `notifyFreeingObject` never fires — so the memory counter doesn't decrease. The loop drains the cache to empty and then calls `evict()` which does `--list_.end()` on an empty `std::list` — UB.

**Fix**: Add `!empty()` guard: `while (!empty() && evictPolicy_()) { evict(); }`.

**Regression test**: `JitEngineTest.cacheEvictionOnEmptyList` holds an external reference to a cached module, then inserts a new module that triggers eviction. Without the fix → crash (exit code 134). With the fix → passes.

### Performance Impact

### Release Note

```text
Release Note:
- Fixed a crash in JIT LRU cache when evicting modules that have external references.
```

### Checklist (For Author)

- [x] (Optional) I have run Sanitizers (ASAN/TSAN) locally for complex C++ changes.

### Breaking Changes

- [x] No

### DIFF
--- a/bolt/jit/tests/ThrustJitEngineTest.cpp
+++ b/bolt/jit/tests/ThrustJitEngineTest.cpp
@@ -266,7 +266,7 @@ TEST_F(JitEngineTest, concurreny) {
 }
 
 TEST_F(JitEngineTest, cacheEvictionOnEmptyList) {
-  // Bug 2: When a cached module has an external reference (CompiledModuleSP
+  // When a cached module has an external reference (CompiledModuleSP
   // held outside the cache), eviction removes the cache entry but the
   // CompiledModuleImpl isn't destroyed (refcount > 0), so notifyFreeingObject
   // is never called and memory usage doesn't decrease. The eviction loop in


kexianda (approved):
+1

PR #397, opened 2026-03-16
[refactor] Introduce ConnectorObjectFactory

yingsu00:

Issue Number: Partially resolves https://github.com/bytedance/bolt/issues/250

### Type of Change

- [ ] 📝 Documentation only

### Description

Introduce ConnectorObjectFactory

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note: None

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/connectors/Connector.h
+++ b/bolt/connectors/Connector.h
@@ -93,11 +93,18 @@ struct ConnectorSplit : public ISerializable {
   virtual std::string toString() const {
     return fmt::format("[split: {}]", connectorId);
   }
+
+  /// Returns the size of this split in bytes, or 0 if unknown.
+  virtual int64_t splitSizeBytes() const {
+    return 0;
+  }
 };
 
 class ColumnHandle : public ISerializable {
  public:
   virtual ~ColumnHandle() = default;
+
+  virtual const std::string& name() const = 0;
 
   folly::dynamic serialize() const override;
 
@@ -125,9 +132,7 @@ class ConnectorTableHandle : public ISerializable {
   /// Returns the connector-dependent table name. Used with
   /// ConnectorMetadata. Implementations need to supply a definition
   /// to work with metadata.
-  virtual const std::string& name() const {
-    BOLT_UNSUPPORTED();
-  }
+  virtual const std::string& name() const = 0;
 
   /// Returns true if the connector table handle supports index lookup.
   virtual bool supportsIndexLookup() const {
@@ -161,6 +166,33 @@ class ConnectorInsertTableHandle : public ISerializable {
   folly::dynamic serialize() const override {
     BOLT_NYI();
   }
+};
+
+class ConnectorLocationHandle : public ISerializable {
+ public:
+  enum class TableType { kNew, kExisting, kTemp };
+
+  ConnectorLocationHandle(std::string connectorId, TableType tableType)
+      : connectorId_{std::move(connectorId)}, tableType_{tableType} {}
+
+  ~ConnectorLocationHandle() override;
+
+  const std::string& connectorId() const {
+    return connectorId_;
+  }
+
+  /// New vs existing vs temp.
+  TableType tableType() const {
+    return tableType_;
+  }
+
+  virtual std::string toString() const = 0;
+
+  folly::dynamic serialize() const override = 0;
+
+ private:
+  const std::string connectorId_;
+  const TableType tableType_;
 };
 
 /// Represents the commit strategy for writing to connector.
@@ -732,6 +764,12 @@ class ConnectorFactory {
 
   // Initialize is called during the factory registration.
   virtual void initialize() {}
+
+  /// Create and register the ConnectorObjectFactory for this connector type.
+  /// The default implementation is a no-op; connectors that support an object
+  /// factory should override this.
+  virtual void registerObjectFactory(const std::string& /*connectorId*/) const {
+  }
 
   const std::string& connectorName() const {
     return name_;
--- a/bolt/connectors/ConnectorObjectFactory.cpp
+++ b/bolt/connectors/ConnectorObjectFactory.cpp
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) International Business Machines Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "bolt/connectors/ConnectorObjectFactory.h"
+
+namespace bytedance::bolt::connector {
+
+ConnectorObjectFactory::~ConnectorObjectFactory() = default;
+
+std::unordered_map<std::string, std::shared_ptr<ConnectorObjectFactory>>&
+connectorObjectFactories() {
+  static std::
+      unordered_map<std::string, std::shared_ptr<ConnectorObjectFactory>>
+          factories;
+  return factories;
+}
+
+bool registerConnectorObjectFactory(
+    const std::shared_ptr<ConnectorObjectFactory>& factory) {
+  bool ok = connectorObjectFactories()
+                .insert({factory->connectorName(), factory})
+                .second;
+  BOLT_CHECK(
+      ok,
+      "ConnectorObjectFactory with name '{}' is already registered",
+      factory->connectorName());
+  return true;
+}
+
+bool hasConnectorObjectFactory(const std::string& connectorName) {
+  return connectorObjectFactories().count(connectorName) == 1;
+}
+
+bool unregisterConnectorObjectFactory(const std::string& connectorName) {
+  auto count = connectorObjectFactories().erase(connectorName);
+  return count == 1;
+}
+
+std::shared_ptr<ConnectorObjectFactory> getConnectorObjectFactory(
+    const std::string& connectorName) {
+  auto it = connectorObjectFactories().find(connectorName);
+  BOLT_CHECK(
+      it != connectorObjectFactories().end(),
+      "ConnectorObjectFactory with name '{}' not registered",
+      connectorName);
+  return it->second;
+}
+
+} // namespace bytedance::bolt::connector
--- a/bolt/connectors/ConnectorObjectFactory.h
+++ b/bolt/connectors/ConnectorObjectFactory.h
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) International Business Machines Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <shared_mutex>
+#include <string>
+
+#include "bolt/connectors/Connector.h"
+#include "bolt/connectors/ConnectorOptions.h"
+
+namespace bytedance::bolt::connector {
+
+/// Factory interface for constructing connector-specific objects from a
+/// generic, serialized representation (ConnectorOptions). Each connector type
+/// registers exactly one ConnectorObjectFactory keyed by its connector name.
+/// A factory instance is bound to a specific connector instance via
+/// connectorId, which is forwarded to objects such as ConnectorSplit and
+/// ConnectorTableHandle that require it.
+///
+/// Subclasses override the makeXXX methods for the object types they support.
+/// Unsupported methods throw by default.
+class ConnectorObjectFactory {
+ public:
+  /// @param name        The connector type name (e.g. "hive").
+  /// @param connectorId The id of the connector instance this factory serves.
+  explicit ConnectorObjectFactory(std::string name, std::string connectorId)
+      : name_(std::move(name)), connectorId_(std::move(connectorId)) {}
+
+  virtual ~ConnectorObjectFactory();
+
+  /// Returns the connector type name this factory is registered under.
+  const std::string& connectorName() const {
+    return name_;
+  }
+
+  /// Returns the connector instance id passed at construction time.
+  const std::string& connectorId() const {
+    return connectorId_;
+  }
+
+  /// Deserializes a ConnectorSplit from @p filePath, @p start, @p length, and
+  /// connector-specific fields encoded in @p options.
+  virtual std::shared_ptr<ConnectorSplit> makeConnectorSplit(
+      const std::string& /*filePath*/,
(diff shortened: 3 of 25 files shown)


yingsu00:
@ZacBlanco can you please review?

bolt/connectors/Connector.h
```diff
@@ -99,6 +99,10 @@ class ColumnHandle : public ISerializable {
  public:
   virtual ~ColumnHandle() = default;
 
+  virtual const std::string& name() const {
+    BOLT_UNSUPPORTED();
+  }
```
ZacBlanco: Would it be better to make this:


```suggestion
  virtual const std::string& name() const = 0
```

This way, we force users to implement this method?

bolt/connectors/Connector.h
ZacBlanco, line 1: Can we add some documentation to the functions/classes/members of this header and ConnectorObjectFactory.h? I think we don't need it for the connector implementation, but I think on SPI-level headers it would be useful for future developers to understand the purpose of these classes for future implementations/authors

bolt/connectors/ConnectorObjectFactory.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace bytedance::bolt::connector {
+
+std::unordered_map<std::string, std::shared_ptr<ConnectorObjectFactory>>&
+connectorObjectFactories() {
+  static std::
+      unordered_map<std::string, std::shared_ptr<ConnectorObjectFactory>>
+          factories;
+  return factories;
+}
```
ZacBlanco, line 29: While this can be useful for maintaining a global list of loaded connectors and to allow connector libraries to automatically register themselves, I wonder about whether it would be better to make this instanced for testing purposes. Similar to how in the Java world, we don't really use `static` on items unless it's strictly required. I am not opposed to this, but would like to hear your thoughts

bolt/connectors/hive/tests/HiveObjectFactoryTest.cpp
```diff
@@ -0,0 +1,239 @@
...
+  std::shared_ptr<connector::ColumnHandle> makeHiveColumnHandle(
+      const std::string& name,
+      HiveColumnHandle::ColumnType columnType,
+      const TypePtr& dataType,
+      const TypePtr& hiveType,
+      std::vector<std::string> requiredSubfields) {
+    folly::dynamic opts = folly::dynamic::object;
+
+    // columnType would be serialized as int32_t
+    opts["columnType"] = (int32_t)columnType;
```
ZacBlanco: nit: use c++-style cast

bolt/connectors/ConnectorObjectFactory.h
```diff
@@ -0,0 +1,93 @@
...
+class ConnectorObjectFactory {
+ public:
+  virtual ~ConnectorObjectFactory() = default;
+
+  const std::string& connectorName() const {
+    return name_;
+  }
+
+  virtual std::shared_ptr<ConnectorSplit> makeConnectorSplit(
+      const std::string& connectorId,
```
ZacBlanco: What's the reason for having the connectorId as a parameter to all these methods? I feel like the ConnectorObjectFactory instance should have the connector ID already stored internally as a class member rather than requiring it to be passed on every call?

bolt/connectors/ConnectorObjectFactory.h
```diff
@@ -0,0 +1,93 @@
...
+      const std::string& /*filePath*/,
+      uint64_t /*start*/,
+      uint64_t /*length*/,
+      const folly::dynamic& /*options*/) const {
```
ZacBlanco: I know bolt/velox depend on folly, but I think the public API would ideally not _also_ rely on folly. 

In order to reduce the need for the folly dependency, it might be better to have an empty virtual class called `ConnectorOptions` or something similar as the argument type here, and then force the implementers to create a concrete class with all the desired option parameters such that (1) you don't need to be checking for key/value presence in the `dynamic` object, (2) allows connectors to implement their specific set of options in a strongly-typed way.

Then, in the implementations, connectors can do a dynamic cast to check if the option type passed is correct?

```c++
// HiveObjectFactory.cpp
std::shared_ptr<ConnectorSplit> HiveObjectFactory::makeConnectorSplit(
    const std::string& connectorId,
    const std::string& filePath,
    uint64_t start,
    uint64_t length,
    const ConnectorOptions& options) const {

    // ...
    HiveConnectorOptions* hiveOptions = std::dynamic_ptr_cast<HiveConnectorOptions>(options);
    BOLT_CHECK_NOT_NULL(hiveOptions, "could not cast to HiveConnectorOptions!");
    // ...


    }
```

This is what I'm thinking

```c++

virtual class ConnectorOptions {
ConnectorOptions() = default;
~ConnectorOptions() = default;
}

class HiveConnectorOptions : bolt::ConnectorOptions {
    
    private:
        std::string param1;
        std::string param2;
        // ...
}

```


What are your thoughts?

ZacBlanco (commented):
A few small nits and a question or two on the API design. Otherwise I am excited for this!

bolt/connectors/ConnectorObjectFactory.h
yingsu00: @ZacBlanco 
Thanks — this is a good point, and I agree with the direction of keeping the public API independent of Folly and moving toward stronger typing.

That said, given where we are right now (early refactor, adding connector abstraction, no Iceberg yet), I’d like to propose a **hybrid approach**:

We can introduce the lightweight base class as you recommended:

```cpp
class ConnectorOptions {
public:
  virtual ~ConnectorOptions() = default;
};
```

And provide a default implementation:

```cpp
class DynamicConnectorOptions : public ConnectorOptions {
public:
  folly::dynamic options;
};
```

The factory APIs would take `const ConnectorOptions&`, and connectors can internally downcast:

```cpp
auto* opts = dynamic_cast<const DynamicConnectorOptions*>(&options);
VELOX_CHECK_NOT_NULL(opts);
const auto& dynamic = opts->options;
```

This gives us a clean public API (no direct dependency on folly::dynamic) while making minimal changes to current code (we can reuse existing dynamic-based logic). We are still open to a migration path toward strongly-typed options later.

Once the API stabilizes and we see repeated patterns, we can gradually introduce typed options where it adds value. For example:

```cpp
class HiveTableOptions : public ConnectorOptions {
public:
  bool filterPushdownEnabled{true};
  std::unordered_map<std::string, std::string> tableParameters;
};
```

The ObjectFactories in each connector can then support both:

```cpp
if (auto* typed = dynamic_cast<const HiveTableOptions*>(&options)) {
  // use typed version
} else if (auto* dyn = dynamic_cast<const DynamicConnectorOptions*>(&options)) {
  // fallback
}
```

I can also add helpers so that the callers (e.g. TableScanTest) can compose these options easily:
```
DynamicConnectorOptions makeOptions(
    std::initializer_list<std::pair<std::string, folly::dynamic>> kvs) {
  DynamicConnectorOptions opts;
  opts.options = folly::dynamic::object;
  for (auto& [k, v] : kvs) {
    opts.options[k] = v;
  }
  return opts;
}
```
Then the callers can
```
auto opts = makeOptions({
    {"filterPushdownEnabled", true},
    {"sessionTimezoneName", "UTC"}
});
```
If you agree, I'll make the above changes soon.

bolt/connectors/ConnectorObjectFactory.h
yingsu00: > What's the reason for having the connectorId as a parameter to all these methods? I feel like the ConnectorObjectFactory instance should have the connector ID already stored internally as a class member rather than requiring it to be passed on every call?

The convention of this code base uses fixed connector name(Class) but variable connectorId(Instance). For example, in the tests (e.g. bolt/connectors/hive/storage_adapters/s3fs/tests/S3MultipleEndpointsTest.cpp) you could see
```
static const std::string_view kConnectorId1 = "test-hive1";
static const std::string_view kConnectorId2 = "test-hive2";
```
I agree this is counter intuitive. We can discuss options of changing this in the future, but IMHO it's better to keep using it this way in this refactor PR. Since "ConnectorId" actually mean the ID of the connector instances, it's necessary to have its instance id (here as connectorId) when constructing connector objects for that instance.

bolt/connectors/ConnectorObjectFactory.cpp
yingsu00: > While this can be useful for maintaining a global list of loaded connectors and to allow connector libraries to automatically register themselves, I wonder about whether it would be better to make this instanced for testing purposes. Similar to how in the Java world, we don't really use `static` on items unless it's strictly required. I am not opposed to this, but would like to hear your thoughts

@ZacBlanco This is a function-local static (also called a Meyer's singleton) —   the idiomatic C++ pattern for global registries. The `static` local variable in C++ is subtly different than Java. It's wrapped in a function, not a raw global or class-static, which gives it two important properties:                                    
                                                                  
  1. Lazy, thread-safe initialization — since C++11, initialization of function-local statics is guaranteed to happen exactly once, on first call, and is thread-safe by the standard.
  2. No static initialization order disaster — the map is only constructed when first accessed, avoiding the classic C++ problem of one global  depending on another that hasn't been initialized yet.

Unlike Java, C++ has global free functions like connectorFactories() or connectors() that can be called anywhere in the code base:
```
std::unordered_map<std::string, std::shared_ptr<ConnectorFactory>>&
connectorFactories() {
  static std::unordered_map<std::string, std::shared_ptr<ConnectorFactory>>
      factories;
  return factories;
}

std::unordered_map<std::string, std::shared_ptr<Connector>>& connectors() {
  static std::unordered_map<std::string, std::shared_ptr<Connector>> connectors;
  return connectors;
}
```
There is also a free global function unregisterConnectorObjectFactory() , which lets tests clean up after themselves.

While in Java, everything lives inside a class, and there is no way for it to call a global function to obtain this static map. If you have
```
class ConnectorRegistry {      
      private static Map<String, ConnectorFactory> factories = new HashMap<>();                                                          
      // ...                                                            
  }                                                                     
 ```
the problem occurs: If test A registers a factory and test B runs next, test B sees it.  THere is no way for you to call a free global  unregisterConnectorObjectFactory() to unregister it. So you have to instance it in Java.

The free global function in C++ mitigates this issue perfectly. The users can call the wrapping free function anywhere in the code base as long as it's running in the same process to register or unregister a connector. The tests can clean up the registry when it's done easily. Also, C++ tests typically run in isolated processes (each _test binary is its own process), so even without explicit cleanup, state doesn't leak between test binaries. That's a much smaller impact than the JVM case.  

If instancing were desired in C++, you'd pass a ConnectorObjectFactoryRegistry through the call chain that needs to create a split or column handle. That's significant coupling for a component whose concrete implementation (HiveConnector vs TpchConnector vs future connectors) is meant to be opaque to the engine. The global registry keeps that coupling out of the hot path.

So that being said, the pattern used here is an established convention in C++. I hope this explanation helps.

File: bolt/connectors/ConnectorObjectFactory.h (hunk 5 above)
Comment (ZacBlanco) [reply bolt/connectors/ConnectorObjectFactory.h]: This sounds reasonable to me. I am good with this approach

File: bolt/exec/TableScan.cpp [hunk 6]
```diff
@@ -502,14 +497,7 @@ void TableScan::checkPreload() {
...
 
-            auto hiveSplit = std::dynamic_pointer_cast<
-                const connector::hive::HiveConnectorSplit>(split);
-            // Handle the default max value for length by treating it as 0
-            int64_t preloadBytes = 0;
-            if (hiveSplit &&
-                hiveSplit->length != std::numeric_limits<uint64_t>::max()) {
-              preloadBytes = static_cast<int64_t>(hiveSplit->length);
-            }
+            int64_t preloadBytes = split->splitSizeBytes();
```
ZacBlanco, line 499: I think with this change we need to implement the `splitSizeBytes()` method inside of HiveConnectorSplit?

bolt/connectors/hive/HiveConnector.cpp
```diff
@@ -189,6 +193,13 @@ void HiveConnectorFactory::initialize() {
   }();
 }
 
+void HiveConnectorFactory::registerObjectFactory() const {
```
ZacBlanco: Should this method be called when the connector factory is registered too? What's the expected sequence that registers the ObjectFactory?

Connector.cpp calls `connectorFactory->initialize()`, but I don't see where `registerObjectFactory` would normally be called?

bolt/connectors/hive/HiveDataSink.h
```diff
@@ -49,24 +50,18 @@ class LocationHandle;
 using LocationHandlePtr = std::shared_ptr<const LocationHandle>;
 
 /// Location related properties of the Hive table to be written.
-class LocationHandle : public ISerializable {
+class LocationHandle : public ConnectorLocationHandle {
```
ZacBlanco, line 53: Should the serialization/deserialization implementation for classes which inherit from this one be updated to serialize the connectorId_ and tableType_ fields from ConnectorLocationHandle?

bolt/connectors/ConnectorObjectFactory.cpp
ZacBlanco: Makes enough sense to me. Thanks for your explanation.

bolt/connectors/ConnectorObjectFactory.h
ZacBlanco: I would prefer to have instances of connectorObjectFactory associated already with a connectorId, rather than having to pass connectorId to every method. Maybe we would need to add a method like `getConnectorId()` to the objectFactory interface to make this 

I am ok to push changing some of this to (yet another) later refactor

bolt/connectors/Connector.h
yingsu00: @ZacBlanco I added documentation for the ConnectorObjectFactories. As to the Connector class, I do plan to clean it up and add doc later.

bolt/connectors/ConnectorObjectFactory.h
yingsu00: > I would prefer to have instances of connectorObjectFactory associated already with a connectorId, rather than having to pass connectorId to every method. Maybe we would need to add a method like `getConnectorId()` to the objectFactory interface to make this
> 
> I am ok to push changing some of this to (yet another) later refactor

I might have misunderstood you last time. Were you saying to pass the connectorId to the factory's constructor? Yes I have made this change and removed the connectorId parameter from the makeXXX functions:
```
explicit ConnectorObjectFactory(std::string name, std::string connectorId)
      : name_(std::move(name)), connectorId_(std::move(connectorId)) {}
```

bolt/exec/TableScan.cpp
yingsu00: done

bolt/connectors/hive/HiveConnector.cpp
yingsu00: nice catch! Yes it should be called in newConnector().
```
  std::shared_ptr<Connector> newConnector(
      const std::string& id,
      std::shared_ptr<const config::ConfigBase> config,
      folly::Executor* executor = nullptr) override {
    registerObjectFactory(id);
    return std::make_shared<HiveConnector>(id, config, executor);
  }
```

bolt/connectors/hive/HiveDataSink.h
yingsu00: done

yingsu00:
> I would prefer to have instances of connectorObjectFactory associated already with a connectorId, r

Done.

@ZacBlanco Thank you for your thorough review. I've addressed your comments. Would you please review again?

bolt/connectors/ConnectorObjectFactory.h
ZacBlanco: Yes, this is what I meant. Thank you for making these changes!

PR #458, opened 2026-03-31
build: support compiling with clang 19

ZacBlanco:

Brings our project's compatibility with later compilers up to more modern standards. Clang 19 was released in Septemer of 2024.

### Type of Change

- [ ] 📝 Documentation only

### Description

Updates project to bring compatibility with clang 19. This is important for a number of reasons

1. Clang is much less resource-intensive than GCC, specifically 50% less memory usage compiling many of the units in this project. Allows more developers access to the project.
2. Clang generally has better compile times than GCC
3. More modern compilers mean we can slowly bump the cpp standard used, meaning access to new C++ language features from C++20 and beyond.

Specific changes are outlined here:

1. duckdb fails to compile on clang 19 until the 1.1.X line. The duckdb dependency and related plan parsing code has been updated to support the duckdb API changes.
2. Some template code was erroring on clang 19. The code has been fixed to prevent the errors.
3. Added a new clang-based dev container + dockerfile to support developers in setting up an environment to build with clang
4. Adds a necessary rt-lib=compiler-rt compilation argument when building on aarch64 with clang
5. Upgrades libunwind to 1.8.3 across all build configurations to allow building on aarch64
6. Forces arrow recipe to use rapidjson/cci.20250205 which fixes an issue with incorrectly-resolved rapidjson dependency compared to the previous version specified.
7. Changes clang-tidy to only compile `.c**` files that are in the compilation DB to prevent header include errors. The line filters for headers are should still be propagated to users as long as the headers are used in a modified compilation unit.

### Performance Impact

### Release Note

Release Note:

```text
Release Note:
- Add support for building with clang 19
- Upgrade DuckDB to 1.1.3
- Upgrade rapidjson to cci.20250205
- Upgrade libunwind to 1.8.3
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/duckdb/conversion/DuckParser.cpp
+++ b/bolt/duckdb/conversion/DuckParser.cpp
@@ -35,7 +35,7 @@
 #include "bolt/parse/Expressions.h"
 #include "bolt/type/Variant.h"
 
-#include <sstream>
+#include <cmath>
 
 #include <duckdb.hpp> // @manual
 #include <duckdb/parser/expression/between_expression.hpp> // @manual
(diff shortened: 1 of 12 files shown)


bolt/duckdb/conversion/DuckParser.cpp
```diff
@@ -251,6 +253,116 @@ std::shared_ptr<const core::ConstantExpr> tryParseInterval(
...
+    if (constInput->type()->isBigint() && !constInput->value().isNull()) {
+      value = constInput->value().value<int64_t>();
+    }
+  } else if (
+      const auto* castInput =
+          dynamic_cast<const core::CastExpr*>(input.get())) {
+    if (castInput->type()->isBigint()) {
+      if (const auto* constInput = dynamic_cast<const core::ConstantExpr*>(
+              castInput->getInput().get())) {
+        if (constInput->type()->isBigint() && !constInput->value().isNull()) {
```
NicolasDenoyelle: The way I understand it is: "If the expression is a cast from bigint to bigint".
Does it matter to know if the cast output type is bigint?

bolt/duckdb/conversion/DuckParser.cpp
```diff
@@ -251,6 +253,116 @@ std::shared_ptr<const core::ConstantExpr> tryParseInterval(
...
+  if (unitName == "hour" || unitName == "hours") {
+    multiplier = 60 * 60 * 1'000;
+  } else if (unitName == "minute" || unitName == "minutes") {
+    multiplier = 60 * 1'000;
+  } else if (unitName == "second" || unitName == "seconds") {
+    multiplier = 1'000;
+  } else if (unitName == "millisecond" || unitName == "milliseconds") {
+    multiplier = 1;
+  } else {
+    return nullptr;
```
NicolasDenoyelle: Should we have the duplicated code in one static (`namespace {}`) function?

conanfile.py
```diff
@@ -352,6 +350,7 @@ def configure(self):
             self.options[llvm_core].with_z3 = False
             self.options[llvm_core].with_zstd = False
             self.options[llvm_core].with_ffi = False
+            self.options[llvm_core].targets = "AArch64;ARM;X86"
```
NicolasDenoyelle: Should that be gated similarly to line 320, i.e with:
```
if str(self.settings.arch) in ["armv8", "arm", "armv9"]:
```

NicolasDenoyelle (commented):
Hi Zack, I have a couple minor comments.

That looks great overall and a welcome improvement.


I am not well versed in the tests you changed so I would probably put another pair of eyes on it if we have someone who knows that code. I didn't see anything shocking there.

conanfile.py
ZacBlanco: I thought about it and went with this to keep things simple at first. I think we could gate it and do X86/ARM builds separately though for even more savings. (why would an X86 build need to generate ARM for bolt anyways?). I will play around and see if I can get this to work

bolt/duckdb/conversion/DuckParser.cpp
ZacBlanco: Actually, I think a lot of this code might be incorrect. We shouldn't be trying to manually parse these intervals. DuckDB should be taking care of parsing the tree correctly here. I will try to re-implement this part to get tests passing.

ZacBlanco: I reverted all the changes in the parser. The new DuckDB actually handles parsing all the intervals into CALL expressions anyways, so we don't actually need all this extra code. I think the intern added it to pass the tests, but I felt modifying the test expectation instead was reasonable.

conanfile.py
ZacBlanco: This is now updated to only use the respective targets for each LLVM build

NicolasDenoyelle (approved):
LGTM on the latest changes.

kexianda (approved):
LGTM

PR #459, opened 2026-03-31
build: Reduce cost of heavy compilation units

ZacBlanco:

Reduces compilation cost (CPU and peak RSS) of most expensive compilation units: MinMaxByAggregates.cpp and CastExpr-tpl.cpp

### Type of Change

- [ ] 📝 Documentation only

### Description

- Reduces CastExpr-tpl compilation cost by ~67% by reducing the number of required template instantiations
- Parallelizes MinMaxBy registration to a MinBy/MaxBy unit to cut cost and memory consumption in half for each. Total cost is not really reduced here but splitting lowers peak memory and helps improve compilation wall time

### Performance Impact

### Release Note

N/A (changes only affect compilation)

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/expression/CastExpr-tpl.cpp
+++ b/bolt/expression/CastExpr-tpl.cpp
@@ -823,6 +823,38 @@ template <PrimitiveKind fromKind, PrimitiveKind toKind>
 class VectorConverter : public ConverterBase {
  public:
   virtual ~VectorConverter() = default;
+
+ private:
+  static constexpr bool isIntegerKind(PrimitiveKind kind) {
+    return (
+        kind == PrimitiveKind::TINYINT || kind == PrimitiveKind::SMALLINT ||
+        kind == PrimitiveKind::INTEGER || kind == PrimitiveKind::BIGINT);
+  }
+
+  static constexpr bool isFloatKind(PrimitiveKind kind) {
+    return kind == PrimitiveKind::REAL || kind == PrimitiveKind::DOUBLE;
+  }
+
+  static constexpr bool isDecimalKind(PrimitiveKind kind) {
+    return (
+        kind == PrimitiveKind::SHORT_DECIMAL ||
+        kind == PrimitiveKind::LONG_DECIMAL);
+  }
+
+  static constexpr bool kLegacySensitive = toKind == PrimitiveKind::STRING &&
+      (fromKind == PrimitiveKind::TIMESTAMP || isFloatKind(fromKind));
+
+  static constexpr bool kTruncateSensitive =
+      // integer -> boolean
+      (toKind == PrimitiveKind::BOOLEAN && isIntegerKind(fromKind)) ||
+      // string/float/decimal -> integer
+      (isIntegerKind(toKind) &&
+       (fromKind == PrimitiveKind::STRING || isFloatKind(fromKind) ||
+        isDecimalKind(fromKind))) ||
+      // float -> float
+      (isFloatKind(toKind) && isFloatKind(fromKind));
+
+ public:
   template <bool legacy, bool truncate>
   FLATTEN void convertWithPolicy(
       const SelectivityVector& rows,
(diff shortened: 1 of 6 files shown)


bolt/expression/CastExpr-tpl.cpp
```diff
@@ -880,20 +912,44 @@ class VectorConverter : public ConverterBase {
...
+      } else if constexpr (kTruncateSensitive) {
+        bool truncate = queryConfig.isCastToIntByTruncate();
+        if (truncate) {
+          convertWithPolicy<false, true>(
+              rows, input, context, result, errorPolicy);
+        } else {
+          convertWithPolicy<false, false>(
+              rows, input, context, result, errorPolicy);
+        }
       } else {
```
NicolasDenoyelle, line 952: FYI, my intern noticed a compiler bug on `if constexpr`.
I don't remember exactly which it was between:
* The `else` part was always running
* `else if` from `else if constexpr` was not running as `else if` but as `if`.

I would suggest trying a little code snippet to see if that could be an issue here.

NicolasDenoyelle (approved):
LGTM, If you can spare a few minutes, check the comment I left.

bolt/expression/CastExpr-tpl.cpp
ZacBlanco: Hmm. My intern came up with this to showcase that the extra branches shouldn't be executed. If they are, we would expect a failure:

```c++
#include <iostream>
#include <cassert>
#include <type_traits>

template <typename T>
constexpr int chain_test() {
  int result = 0;

  if constexpr (std::is_same_v<T, int>) {
    result = 1;
  } else if constexpr (std::is_same_v<T, float>) {
    result = 2;
  } else {
    result = 3;
  }

  return result;
}

template <typename T>
int return_test() {
  int result = 0;
  if constexpr (std::is_same_v<T, int>) {
    std::cout << "return_test branch: int\n";
    result = 10;
  } else if constexpr (std::is_same_v<T, float>) {
    std::cout << "return_test branch: float\n";
    result = 20;
  } else {
    std::cout << "return_test branch: else\n";
    result = 30;
  }
  return result;
}

int main() {
  std::cout << "chain_test<int>   = " << chain_test<int>() << "\n";
  std::cout << "chain_test<float> = " << chain_test<float>() << "\n";
  std::cout << "chain_test<char>  = " << chain_test<char>() << "\n";

  std::cout << "return_test<int>   = " << return_test<int>() << "\n";
  std::cout << "return_test<float> = " << return_test<float>() << "\n";
  std::cout << "return_test<char>  = " << return_test<char>() << "\n";

  static_assert(chain_test<int>() == 1);
  static_assert(chain_test<float>() == 2);
  static_assert(chain_test<char>() == 3);

  assert(return_test<int>() == 10);
  assert(return_test<float>() == 20);
  assert(return_test<char>() == 30);
}

```

And then compiled with `clang++ -std=c++17 -O3 if_constexpr_repro.cpp -o if_constexpr_repro`.

NicolasDenoyelle: Great, thanks for trying. We did something similar but it was on the old gcc-9 or 10. It's unlikely to be an issue anymore.

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -823,6 +823,38 @@ template <PrimitiveKind fromKind, PrimitiveKind toKind>
 class VectorConverter : public ConverterBase {
  public:
   virtual ~VectorConverter() = default;
+
+ private:
+  static constexpr bool isIntegerKind(PrimitiveKind kind) {
```
NicolasDenoyelle, line 828: It might be out of scope for this PR but `PrimitiveKind` sounds redundant with `TypeKind` + `TypeTraits`. I would think that these function could live in `Type.h` and be either a type trait or a `consteval` function.

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -880,20 +912,44 @@ class VectorConverter : public ConverterBase {
...
-      } else if (legacy && !truncate) {
-        convertWithPolicy<true, false>(
-            rows, input, context, result, errorPolicy);
-      } else if (!legacy && truncate) {
-        convertWithPolicy<false, true>(
-            rows, input, context, result, errorPolicy);
+      const auto& queryConfig = context.execCtx()->queryCtx()->queryConfig();
+
+      if constexpr (kLegacySensitive && kTruncateSensitive) {
+        bool legacy = queryConfig.isLegacyCast();
```
NicolasDenoyelle, line 918: Can `legacy` and `truncate` value be changed to `constexpr`? That would simplify the enumeration of all the template combinations into one.

bolt/functions/prestosql/aggregates/MinMaxByAggregateRegistration.h
```diff
@@ -0,0 +1,108 @@
...
+  std::vector<std::shared_ptr<exec::AggregateFunctionSignature>> signatures;
+  signatures.push_back(exec::AggregateFunctionSignatureBuilder()
+                           .typeVariable("V")
+                           .orderableTypeVariable("C")
+                           .returnType("V")
+                           .intermediateType("row(V,C)")
+                           .argumentType("V")
+                           .argumentType("C")
+                           .build());
+  const std::vector<std::string> supportedCompareTypes = {
```
NicolasDenoyelle: Can this be a `constexpr std::array`? (It might be c++20)

NicolasDenoyelle (approved):
I added minor and optional comments.
That looks ready to merge.

bolt/expression/CastExpr-tpl.cpp
ZacBlanco: since `legacy` and `truncate` come from queryConfig, they can be controlled by the user. I don't think it's possible to make them `constexpr`

bolt/functions/prestosql/aggregates/MinMaxByAggregateRegistration.h
ZacBlanco: Good point. std::string doesn't work AFAIK, but I can do the following:

```
static constexpr std::array<const char*, 10> kSupportedCompareTypes = { ...
```

bolt/expression/CastExpr-tpl.cpp
ZacBlanco: I agree that there doesn't seem to be a good reason here to have `PrimitiveKind` defined within this file. It should probably be using Type.h. However, as you mentioned I think changing this is out of scope for the PR. But adding some additional traits to Type.h would be a good improvement. Things like `isIntegral`, `isFloatingPoint`, etc might be useful for the future.

I created https://github.com/bytedance/bolt/issues/492 to log this task

oarap (approved):

PR #460, opened 2026-03-31
docs: Add bolt-build skill

ZacBlanco:

Allows AI agents to gain more autonomy by instructing them how to compile bolt efficiently. 

### Type of Change

- [x] 📝 Documentation only

### Description

Adds a bolt-build agent skill, compatible with tools like Codex or Claude to instruct agents how to compile bolt properly. It gives AI agents more autonomy to iterate and test changes without user input.

### Performance Impact

### Release Note

N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/.agents/skills/bolt-build/SKILL.md
+++ b/.agents/skills/bolt-build/SKILL.md
@@ -10,10 +10,9 @@ tests, or benchmarks
 
 ## Workflow
 
-- Configure the repository using instructions from the "Configure" section ONLY if necessary
-- Compile the project using instructions from "Compile" section
+- **Configure** the repository using instructions from the "Configure" section ONLY if necessary
+- **Compile** the project using instructions from "Compile" section
   - Wait for all units to finish. If compilation exits early, check the error
-  - If syntax error, consider the users' previous instructions and fix it
   - If killed (e.g. due to OOM), restart the build, but lower the number of jobs
   by adding the `--parallel` flag and specify a lower number than the machine's available cores.
   If the build is repeatedly killed to due OOMs, decrease it after each failure.
@@ -84,17 +83,14 @@ Use this to decide whether you can reuse the existing preset or need to reconfig
 
 - Prefer the narrowest target that validates the change.
 - Reuse the current configured preset whenever possible.
-- If a compile fails, fix the smallest coherent issue and re-run the same narrow build first.
-- In status updates and final summaries, state whether you reused an existing configured preset or had to reconfigure.
 
 ## Finding the narrowest target
 
 To identify the smallest target that contains an edited compilation unit:
 
 1. Locate the edited `*.cpp` file's directory.
-2. Find the `CMakeLists.txt` in that directory (or the nearest parent directory).
-3. Search for the target that lists the edited `*.cpp` file in its source list.
-4. Build that target.
+2. Find `CMakeLists.txt` in that directory (or the nearest parent directory).
+3. Find the target with the edited `*.cpp` file in its source list.
 
 Example:
 - Edited file: `bolt/core/execution/ExecutionPlan.cpp`
--- a/.agents/skills/bolt-build/SKILL.md
+++ b/.agents/skills/bolt-build/SKILL.md
@@ -1,0 +1,80 @@
+---
+name: bolt-build
+description: Use when working on the Bolt repository and you need to configure or compile code correctly and efficiently. This skill defines the canonical Bolt build workflow, including when to run `make <TARGET> BOLT_CONAN_CONFIGURE_ONLY=1`, when to switch to `cmake --build --preset conan-<build type> --target <TARGET>`, and when reconfiguration is required because the build type or test inclusion has changed.
+---
+
+# Bolt Build
+
+Use this skill for any Bolt task that requires configuring or compiling the repository.
+
+## Core rule
+
+Do not use `make` for routine incremental compilation once the build is configured.
+
+## Configure
+
+To configure a build, run:
+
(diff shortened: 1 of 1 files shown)


.agents/skills/bolt-build/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+---
+
+# Bolt Build
+
+Use this skill for any Bolt task that requires configuring or compiling the repository for binaries
+tests, or benchmarks
+
+## Workflow
+
+- Configure the repository using instructions from the "Configure" section ONLY if necessary
```
NicolasDenoyelle: Sometimes when I ask AI to make a skill file, the items in the list like this one would have a bold title e.g `- **Configure**`
I don't know if it will impact the actual AI output but it looks more striking when you read each item and it starts with the keyword in bold font.

.agents/skills/bolt-build/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+- Compile the project using instructions from "Compile" section
+  - Wait for all units to finish. If compilation exits early, check the error
+  - If syntax error, consider the users' previous instructions and fix it
```
NicolasDenoyelle: I am thinking that altering the code in a skill that's about building might be surprising the user because the fix may be larger than expected. Perhaps, we should "show fix diff and prompt the user before applying it".

.agents/skills/bolt-build/SKILL.md
```diff
@@ -0,0 +1,105 @@
+---
+name: bolt-build
```
NicolasDenoyelle, line 2: Should we add parameters (e.g debug=true)?

NicolasDenoyelle (commented):
I added a few suggestions.
I looks really good overall.

.agents/skills/bolt-build/SKILL.md
ZacBlanco: A bit confused by what you mean here. Can you elaborate?

.agents/skills/bolt-build/SKILL.md
ZacBlanco: I added it. I don't think it will have much impact though. Personally I think the section titles should be enough to emphasize certain parts, but I also haven't written many skills before, so it's hard to comment on the efficacy 🤷 

I think if anything, it would be best to condense the skill as much as possible so that it takes up as little of the context window while still providing the necessary information. Adding the `*` characters might add a token or two

.agents/skills/bolt-build/SKILL.md
ZacBlanco: This is a good point, we can probably leave it out. I designed it in the context of allowing the AI to become more autonomous, so this felt reasonable at the time. I agree that it actually feels out of scope if the skill is specifically for building bolt. I think any existing context is probably enough for the AI to determine whether it should make changes itself or discuss with the user.

.agents/skills/bolt-build/SKILL.md
NicolasDenoyelle: That's a fair point. Your call.

.agents/skills/bolt-build/SKILL.md
NicolasDenoyelle: I saw that later in the skill file you mention the user requesting a build with debug. I wonder how that happens.

For instance, when I use a skill, I type `/<skill> <args>...` to run it.
If you only have `/<skill>` how do you ask for a debug build? Can you type input text after the command?

.agents/skills/bolt-build/SKILL.md
NicolasDenoyelle: AI taking initiatives with side effects can be a good or a bad surprise.
Yesterday, I was trying to make it review my code and it started pushing code upstream. 
For that reason, I would remain cautious about side effects and if it is not expected, I would add an explicit statement that no file should be modified until the next input.

.agents/skills/bolt-build/SKILL.md
ZacBlanco: Oh interesting, I actually don't do anything with the `/skill` command. My understanding is that the tool (codex, claude code, coco, etc) as a harness loads in all skills in the `.agents/*/SKILL.md` files, and it's up to the AI to actually "choose" when to use the skill.

However, looking at the Claude code documentation, it looks like they've merged skills and commands: https://code.claude.com/docs/en/skills

In codex, it's similar: https://developers.openai.com/codex/skills

> Codex can activate skills in two ways:
> 
>     Explicit invocation: Include the skill directly in your prompt. In CLI/IDE, run /skills or type $ to mention a skill.
>     Implicit invocation: Codex can choose a skill when your task matches the skill description.
> 
> Because implicit matching depends on description, write descriptions with clear scope and boundaries.


It looks like the skill can be activated both explicitly or implicitly. I've actually only ever used the implicit method with this skill. It would be interesting to add the explicit invocation, but I need to think about what that looks like

.e.g

```
/skill bolt-build <makefile target> <cmake target>?
```

Should we always require the makefile target, or only the cmake target. Maybe we can have an offline discussion about explicit skill invocation

NicolasDenoyelle: Thanks for looking into the details.
I am thinking we should keep it simple. Whatever is the most common usage that comes to mind. Perhaps it could a string describing what should be built, e.g "the unittest <test> with debug options".

NicolasDenoyelle (approved):

PR #472, opened 2026-04-01
ci: reduce test runtime in JIT cmp tests

ZacBlanco:

Reduce the total test runtime for some of the JIT comparison tests.

### Type of Change

- [ ] 📝 Documentation only

### Description

These tests generally take the bulk of the time
during the testing phase in CI. They take a long
time due to performing O(n^2) operations on
randomly-generated vectors. The default vector 
size is 1000, but reducing it to 256 reduces the 
amount of work by a factor of 95% while still
maintaining a large amount of comparisons

### Performance Impact

### Release Note
 
N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/jit/tests/RowEqRowIRTest.cpp
+++ b/bolt/jit/tests/RowEqRowIRTest.cpp
@@ -39,7 +39,7 @@ using namespace bytedance::bolt;
 using namespace bytedance::bolt::exec::test;
 
 static constexpr int32_t kNumVectors = 1;
-static constexpr int32_t kRowsPerVector = 1'000;
+static constexpr int32_t kRowsPerVector = 256;
 
 #define DEBUG 0
 namespace {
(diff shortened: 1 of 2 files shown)


ZacBlanco:
FYI @kexianda @frankobe - I think we can speed up most build jobs by about 5 min (~25-30%) if we reduce the runtime of these tests from the JIT suite. Technically these tests are just generating a bunch of values with the fuzzer and doing a bunch of comparisons. I think lowering the value to 256 here is reasonable and would still give good signal, but if you think otherwise please let me know. 

Full job suite runtime drops from ~10 minutes to just 4m30s with these changes.

frankobe (approved):
Should be good imo, 1000 never catch any bugs in CI

PR #475, opened 2026-04-02
[feat][cp] Enhance floating point value retrieval with type-specific handling in parquet reader

guhaiyan0221:
Summary:
This PR enables the Parquet `FloatingPointColumnReader` to support reading `FLOAT` columns as `DOUBLE`, allowing for schema evolution where the requested type is wider than the file type to avoid result mismatch. It also enforces the bulk path (fast path) for these operations and adds comprehensive test coverage.

1.  **Type Promotion Support**: Updated `ParquetColumnReader::build` to instantiate `FloatingPointColumnReader<float, double>` when the file type is `REAL` but the requested type is `DOUBLE`.
2.  **Always Enable Bulk Path**: Overridden `hasBulkPath()` in `FloatingPointColumnReader` to always return `true`.
3.  **Safe Bulk Path Selection**: Updated `useFastPath` in `DecoderUtil.h` to disable the fast path when there is a mismatch between the filter type and the data type (e.g., `FloatingPointRange<double>` on `float` data).
4.  **Comprehensive Testing**: Added `floatToDoubleComprehensive_*` tests covering various combinations of nulls, dictionary encoding, filters, and deletions.

In this PR, we explicitly set `kHasBulkPath = true` for `FloatingPointColumnReader`, even when `TData` (file type, e.g., `float`) differs from `TRequested` (output type, e.g., `double`).

**Reasoning:**
The Parquet decoder architecture decouples the decoding phase from the type conversion phase. The buffer management in `SelectiveColumnReader` relies on the file type `TData`:

*   **Buffer Allocation**: `prepareRead<TData>()` initializes the `values_` buffer with `valueSize_ = sizeof(TData)`.
*   **Decoding**: The decoding process utilizes the fast path (bulk path) mechanisms if possible:
    *   **Dictionary Encoding**: `RleBpDataDecoder` executes `bulkScan`, which calls `DictionaryColumnVisitor::processRun` or `processRle`. These methods copy values from the dictionary (which holds `TData`) directly into the `values_` buffer (casted to `TData*`).
    *   **Direct Encoding**: Uses `fixedWidthScan` to read `TData` values directly.
    *   **Null Handling**: `dwio::common::nonNullRowsFromDense` is used to identify valid rows for decoding.
    Crucially, all these operations write `sizeof(TData)` bytes per value, matching the buffer allocation.
*   **Bulk Path Validity**: Since the decoder and the buffer both operate in the `TData` domain, the bulk path (fast path) logic—which efficiently decodes batches of data—remains valid and safe.
*   **Deferred Conversion**: Type conversion to `TRequested` is deferred until the `getValues()` call, where `upcastScalarValues<TData, TRequested>()` handles the promotion (e.g., allocating a new 8-byte buffer for doubles and converting the 4-byte floats).

This design allows us to leverage the performance benefits of the bulk path for `float` -> `double` reads without modification to the decoders.

`FloatingPointRange` explicitly forbids cross-type SIMD evaluation. For instance, calling `FloatingPointRange<float>::testValues` with a `double` batch triggers a `VELOX_FAIL` at runtime.

*   **The Fix**: `useFastPath` now returns `false` if:
    *   The filter is `FloatingPointRange<double>` but the data type is not `double`.
    *   The filter is `FloatingPointRange<float>` but the data type is not `float`.
*   **Impact**: For Float-to-Double promotion, the bulk path is used when there is no filter or the filter is compatible (e.g., `IsNull`, `IsNotNull`). If a `double` range filter is provided, the reader safely falls back to the
bulk path.

**Before this PR:**
`TRequested` served three main purposes:
1.  **Bulk Path Determination**: Used to decide if the fast path could be enabled (`std::is_same_v<TData, TRequested>`).
2.  **Result Container**: Defined the type of the output vector.
3.  **Null Handling**: Used as the type for default null values in `filterNulls`.

**After this PR:**
1.  **Bulk Path Determination**: No longer used for this check in Parquet. We explicitly enable bulk path for floating-point conversions because the decoder operates on `TData`.
2.  **Result Container**: Remains the primary type for the output vector.
3.  **Null Handling**: The usage of `filterNulls<TRequested>` remains unchanged.

**Why `filterNulls<TRequested>` was not changed to `filterNulls<TData>`:** Although `filterNulls<TRequested>` writes `sizeof(TRequested)` bytes (e.g., 8 bytes for double) into a buffer allocated for `TData` (e.g., 4 bytes for float), we decided not to change this because:
1.  **Safety via Padding**: The `values_` buffer allocation includes `simd::kPadding`, which provides sufficient buffer space to prevent memory corruption even when writing larger null values at the end of the buffer.
2.  **Data Integrity**: While writing a larger null value might overwrite adjacent bytes, those bytes correspond to subsequent values. Since the current value is null, its specific bit pattern is irrelevant. If the subsequent va is valid, it will be written correctly afterwards, overwriting the "tail" of the previous null value.
3.  **DWRF Compatibility**: This code path is shared with DWRF, which may rely on the existing behavior. **Before this PR:**
`TRequested` served three main purposes:
1.  **Bulk Path Determination**: Used to decide if the fast path could be enabled (`std::is_same_v<TData, TRequested>`).
2.  **Result Container**: Defined the type of the output vector.
3.  **Null Handling**: Used as the type for default null values in `filterNulls`.

**After this PR:**
1.  **Bulk Path Determination**: No longer used for this check in Parquet. We explicitly enable bulk path for floating-point conversions because the decoder operates on `TData`.
2.  **Result Container**: Remains the primary type for the output vector.
3.  **Null Handling**: The usage of `filterNulls<TRequested>` remains unchanged.

**Why `filterNulls<TRequested>` was not changed to `filterNulls<TData>`:** Although `filterNulls<TRequested>` writes `sizeof(TRequested)` bytes (e.g., 8 bytes for double) into a buffer allocated for `TData` (e.g., 4 bytes for float), we decided not to change this because:
1.  **Safety via Padding**: The `values_` buffer allocation includes `simd::kPadding`, which provides sufficient buffer space to prevent memory corruption even when writing larger null values at the end of the buffer.
2.  **Data Integrity**: While writing a larger null value might overwrite adjacent bytes, those bytes correspond to subsequent values. Since the current value is null, its specific bit pattern is irrelevant. If the subsequent va is valid, it will be written correctly afterwards, overwriting the "tail" of the previous null value.
3.  **DWRF Compatibility**: This code path is shared with DWRF, which may rely on the existing behavior.

Added a comprehensive suite of tests in `ParquetReaderTest.cpp` to verify the correctness of float-to-double promotion under various scenarios. The test matrix covers:

*   **Basic Scenarios**: Simple float-to-double conversion, handling of nulls, parent struct nulls, and row deletions.
*   **Dictionary Encoding**: Verified both with and without dictionary encoding enabled.
*   **Filters**: Tested `IsNull`, `IsNotNull` and value filters to ensure they work correctly with type promotion.
*   **Comprehensive Matrix**: Refactored individual tests into a parameterized test suite `FloatToDoubleEvolutionTest` using GTest's `TEST_P`. This systematically covers all combinations of:
    *   **Nulls**: Presence vs. absence of null values.
    *   **Dictionary**: Enabled vs. disabled.
    *   **Filters**: No filter, `IsNull`, `IsNotNull`, and `>=`
    *   **Density**: Dense reading (contiguous) vs. Non-dense reading (with deletions/sparse reading).

This ensures that the bulk path logic holds up across different decoding paths (RLE/BP vs. Direct) and filtering strategies, while keeping the test code concise and maintainable.

Corresponding PR: https://github.com/facebookincubator/velox/pull/15657
Corresponding PR: https://github.com/facebookincubator/velox/pull/13317

Issue Number: close #191 

### Type of Change

- [ ] 📝 Documentation only

### Description

see Summary

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>


Weixin-Xu (approved):
LGTM

PR #478, opened 2026-04-03
[Feature] Restore safe cast and map-subscript filter pushdown with parser fallback normalization

### LINKED ISSUE #479: [Feature] Unified Architecture for Complex Expression and Filter Pushdown
### Feature Category

SQL Function / Operator

### Problem / Use Case

As a user running queries on datasets with complex filter conditions (like nested map lookups or upcasting), I expect the Parquet reader to efficiently evaluate these operations natively and prune row groups. Currently, pushing down complex expressions like map subscript (`m['key'] = value`) or cast filters (`CAST(col AS type) = value`) relies heavily on frontend engines (like Gluten and Presto) implementing their own custom expression parsers (e.g., `SparkExprToSubfieldFilterParser`).

This disjointed architecture causes:
- Massive code duplication across different compute plugins.
- Lack of feature parity across engines natively executed in Bolt.
- High barrier to entry for adding pushdown support for new functions.
- Unsafe statistics evaluation, where numeric boundaries might be incorrectly evaluated against string `Min/Max` statistics (yielding false negatives).

### Proposed Solution

I propose introducing a new long-term, unified architecture in Bolt to parse generic, engine-specific expressions for filter pushdown and natively apply them to Parquet row group pruning. While the immediate implementation focuses on **Cast** and **Map Subscript** filters, this architecture lays the foundation for any future complex function pushdown (e.g., `substring`, `abs`, etc.).

The design involves:
1. **Generic Expression Parser (`ExprToSubfieldFilterParser`)**: Centralizing parsing logic within Bolt and introducing an alias registration system (`registerAlias()`). Frontends can dynamically map dialect-specific functions (e.g., Spark's `element_at` -> Bolt's generic `subscript`), removing the need for frontend-specific parsers.
2. **Extensible Function Extraction (`FunctionInfo`)**: Modifying `SingleSubfieldExtractor` to generically traverse nested expressions, casts, and dereferences to locate the underlying Parquet column and operation signature.
3. **Dedicated Control Knobs**: Splitting the configuration into granular knobs like `map_subscript_filter_pushdown` and `cast_filter_pushdown`, and enabling them by default (`true`).
4. **Safe Statistics Evaluation**: Modifying `MetadataFilter` and `HiveDataSource` to pass down the new flags. We will enforce that extracted operations (like `common::CastFilter`) are safely evaluated against Parquet `Min/Max` statistics (e.g., only safely pruning when upcasting `INT` -> `BIGINT`).
5. **Observability**: Adding explicit telemetry in `ParquetReader.cpp` that logs the count of retained row groups vs total row groups, allowing users to verify pruning effectiveness.

### References / Prior Art

_No response_

### Importance

High (Needed for production)

### Willingness to Contribute

Yes, I can submit a PR

oarap:
This PR restores cast and map-subscript filter pushdown on top of `oss/main` and makes the multi-parser flow safer.

The main change is that Bolt now:
- tries the built-in Bolt parser first,
- lets fallback parsers normalize only the outer call name,
- retries Bolt parsing on that normalized call,
- and only then falls back to full fallback-parser handling.

This keeps nested Bolt-recognized expressions such as `element_at(...)` and `cast(...)` eligible for Bolt-side subfield-filter conversion even when the outer operator comes
from another dialect.

#### Expression-to-subfield filter parsing
- Restore cast and map-subscript pushdown in `ExprToSubfieldFilter`.
- Add a generic parser hook to normalize the outer call before retrying Bolt parsing.
- Change parser chaining order to:
  1. Bolt parser
  2. normalized-call retry through Bolt
  3. fallback parser full parse

#### Correctness and safety fixes
- Preserve cast/map-subscript wrappers for `IS NULL` / `IS NOT NULL` pushdown instead of collapsing to the inner field.
- Reject non-constant map keys so expressions like `element_at(a, b) = ...` do not produce invalid metadata filters.
- Avoid pushing unsupported repeated-column map filters into Hive metadata filtering; keep those expressions in the remaining filter instead.

#### Hive / Parquet integration
- Wire restored filters through Hive and Parquet metadata pruning paths.
- Update related metadata filter handling and query config plumbing for cast/map-subscript pushdown.

#### Tests
- Add regression coverage for:
  - wrapped null predicates on cast expressions,
  - non-constant map-subscript keys,
  - fallback normalization for outer dialect-specific operator names while still letting Bolt parse nested map/cast expressions,
  - Hive remaining-filter extraction behavior for map-subscript expressions.

### Why

Without this change, mixed-dialect expressions can lose Bolt-side subfield pushdown opportunities because the fallback parser tries to parse the whole expression itself. That
is especially problematic for cases where only the outer operator name is dialect-specific, but the nested expression is something Bolt already understands well, such as
`element_at(...)` or `cast(...)`.

This PR keeps those nested expressions on the Bolt path while also avoiding unsafe pushdown for unsupported repeated-column filters.

Issue Number: close #479

### Type of Change

- [ ] 📝 Documentation only

### Description

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [ ] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/connectors/hive/HiveConnectorUtil.cpp
+++ b/bolt/connectors/hive/HiveConnectorUtil.cpp
@@ -757,6 +757,26 @@ bool isNotExpr(
       exec::FunctionCanonicalName::kNot;
 }
 
+bool containsMapSubscriptExpr(const core::TypedExprPtr& expr) {
+  auto* call = dynamic_cast<const core::CallTypedExpr*>(expr.get());
+  if (call == nullptr) {
+    return false;
+  }
+
+  if ((call->name() == "subscript" || call->name() == "element_at") &&
+      !call->inputs().empty() && call->inputs()[0]->type()->isMap()) {
+    return true;
+  }
+
+  for (const auto& input : call->inputs()) {
+    if (containsMapSubscriptExpr(input)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 core::TypedExprPtr extractFiltersFromRemainingFilter(
     const core::TypedExprPtr& expr,
     core::ExpressionEvaluator* evaluator,
@@ -772,6 +792,11 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
             exec::ExprToSubfieldFilterParser::getInstance()
                 ->leafCallToSubfieldFilter(*call, evaluator, negated)) {
       auto& [subfield, filter] = subfieldAndFilter.value();
+      if (containsMapSubscriptExpr(expr) ||
+          filter->kind() == common::FilterKind::kCast ||
+          filter->kind() == common::FilterKind::kMapSubscript) {
+        return expr;
+      }
       if (auto it = filters.find(subfield); it != filters.end()) {
         oldFilter = it->second.get();
         filter = filter->mergeWith(oldFilter);
--- a/bolt/dwio/common/MetadataFilter.cpp
+++ b/bolt/dwio/common/MetadataFilter.cpp
@@ -45,8 +45,7 @@ struct MetadataFilter::Node {
       const core::ITypedExpr&,
       core::ExpressionEvaluator*,
       bool negated,
-      bool enableMapSubscriptFilter,
-      bool enableCastFilter);
+      MetadataFilter::Options options);
   virtual ~Node() = default;
   virtual void addToScanSpec(ScanSpec&) const = 0;
   virtual uint64_t* eval(LeafResults&, int size) const = 0;
@@ -174,51 +173,25 @@ std::unique_ptr<MetadataFilter::Node> MetadataFilter::Node::fromExpression(
     const core::ITypedExpr& expr,
     core::ExpressionEvaluator* evaluator,
     bool negated,
-    bool enableMapSubscriptFilter,
-    bool enableCastFilter) {
+    MetadataFilter::Options options) {
   auto* call = asCall(&expr);
   if (!call) {
     return nullptr;
   }
   if (call->name() == "and") {
-    auto lhs = fromExpression(
-        *call->inputs()[0],
-        evaluator,
-        negated,
-        enableMapSubscriptFilter,
-        enableCastFilter);
-    auto rhs = fromExpression(
-        *call->inputs()[1],
-        evaluator,
-        negated,
-        enableMapSubscriptFilter,
-        enableCastFilter);
+    auto lhs = fromExpression(*call->inputs()[0], evaluator, negated, options);
+    auto rhs = fromExpression(*call->inputs()[1], evaluator, negated, options);
     return negated ? OrNode::create(std::move(lhs), std::move(rhs))
                    : AndNode::create(std::move(lhs), std::move(rhs));
   }
   if (call->name() == "or") {
-    auto lhs = fromExpression(
-        *call->inputs()[0],
-        evaluator,
-        negated,
-        enableMapSubscriptFilter,
-        enableCastFilter);
-    auto rhs = fromExpression(
-        *call->inputs()[1],
-        evaluator,
-        negated,
-        enableMapSubscriptFilter,
-        enableCastFilter);
+    auto lhs = fromExpression(*call->inputs()[0], evaluator, negated, options);
+    auto rhs = fromExpression(*call->inputs()[1], evaluator, negated, options);
     return negated ? AndNode::create(std::move(lhs), std::move(rhs))
                    : OrNode::create(std::move(lhs), std::move(rhs));
   }
   if (call->name() == "not") {
-    return fromExpression(
-        *call->inputs()[0],
-        evaluator,
-        !negated,
-        enableMapSubscriptFilter,
-        enableCastFilter);
+    return fromExpression(*call->inputs()[0], evaluator, !negated, options);
   }
   try {
     auto subfieldAndFilter =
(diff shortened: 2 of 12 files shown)


bolt/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1441,6 +1441,12 @@ class ParquetRowReader::Impl {
       VLOG(1) << "Total row groups: " << rowGroups_.size();
       VLOG(1) << "Total rows: " << rowNumber;
     }
+
+    if (rowGroupIds_.size() < rowGroups_.size()) {
+      LOG(INFO) << "Bolt Parquet Scan: Retained " << rowGroupIds_.size()
```
markjin1990: Replace `LOG(INFO)` with `VLOG(1)`

bolt/dwio/common/MetadataFilter.cpp
```diff
@@ -227,12 +252,14 @@ MetadataFilter::MetadataFilter(
     ScanSpec& scanSpec,
     const core::ITypedExpr& expr,
     core::ExpressionEvaluator* evaluator,
-    bool enableMapSubscriptFilter)
+    bool enableMapSubscriptFilter,
+    bool enableCastFilter)
```
markjin1990: Is it a good practice that we keep adding new argument to this function? Or maybe we pass something like this
```
struct MetadataFilterOptions {
  bool enableMapSubscriptFilter{true};
  bool enableCastFilter{true};
  // future toggles here
};
```

bolt/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -757,6 +757,26 @@ bool isNotExpr(
       exec::FunctionCanonicalName::kNot;
 }
 
+bool containsMapSubscriptExpr(const core::TypedExprPtr& expr) {
+  auto* call = dynamic_cast<const core::CallTypedExpr*>(expr.get());
+  if (call == nullptr) {
+    return false;
+  }
+
+  if ((call->name() == "subscript" || call->name() == "element_at") &&
```
markjin1990, line 766: Will `presto.default.subscript` and `presto.default.element_at` be ignored here?

bolt/expression/tests/ExprToSubfieldFilterTest.cpp
```diff
@@ -50,12 +51,58 @@ void validateSubfield(
...
 
+core::CallTypedExprPtr renameCall(
+    const core::CallTypedExprPtr& call,
+    const std::string& name) {
+  return std::make_shared<core::CallTypedExpr>(
+      call->type(),
+      std::vector<core::TypedExprPtr>(
+          call->inputs().begin(), call->inputs().end()),
+      name);
+}
```
ZacBlanco: This function is essentially the same as renameRootCall - It would be better if we could combine and use the same header/implementation of this method in HiveConnectorTest.cpp

bolt/expression/ExprToSubfieldFilter.cpp
```diff
@@ -30,10 +30,95 @@
...
+      "presto.default.lte",
+      "presto.default.gte",
+      "presto.default.between",
+      "in",
+      "is_null",
+      "cast",
+      "subscript",
+      "presto.default.subscript",
+      "element_at",
+      "presto.default.element_at"};
```
ZacBlanco, line 84: pedantic: Either put all the presto ones together or have them before or directly after the corresponding


```suggestion
      "eq",
      "presto.default.eq",
      "neq",
      "presto.default.neq",
      "lt",
      "presto.default.lt",
      "gt",
      "presto.default.gt",
      "lte",
      "presto.default.lte",
      "gte",
      "presto.default.gte",
      "between",
      "presto.default.between",
      "in",
      "is_null",
      "cast",
      "subscript",
      "presto.default.subscript",
      "element_at",
      "presto.default.element_at"};
```

bolt/expression/ExprToSubfieldFilter.cpp
```diff
@@ -532,16 +713,147 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeOrFilter(
...
+    if (functionInfo->functionName == "cast" && functionInfo->fieldDepth == 1) {
+      filter = common::createCastFilter(
+          functionInfo->sourceType,
+          functionInfo->targetType,
+          std::move(filter));
+    } else if (
+        (functionInfo->functionName == "subscript" ||
+         functionInfo->functionName == "presto.default.subscript" ||
+         functionInfo->functionName == "element_at" ||
+         functionInfo->functionName == "presto.default.element_at") &&
```
ZacBlanco: This list of function names is repeated in conditionals quite a few times. Can we extract this to some common inlined function? or just put them in a `static std::unordered_set` or something and check for existence in the set?

ZacBlanco (commented):
A few minor things, otherwise lgtm

ZacBlanco (approved):

markjin1990 (approved):

markjin1990:
Found an issue related to this fix: https://github.com/bytedance/bolt/issues/504

PR #484, opened 2026-04-04
[refactor] Move HiveConnectorSplitBuilder to HiveConnectorSplit

yingsu00:

    In the upcoming refactor to decouple Hive from exec tests, we will need
    to make HiveConnectorTestBase connector agnostic. However, it contains
    HiveConnectorSplitBuilder, and this is an obstacle.

### Type of Change

- [ ] 📝 Documentation only

### Description

The HiveConnectorSplitBuilder should go back to the Hive connector
    in connectors/hive/HiveConnectorSplit.cpp. This commit makes this move
    and updates HiveObjectFactory to use it to create Hive splits.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:
N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>


yingsu00:
@ZacBlanco Can you please review this PR? Thanks!

bolt/connectors/hive/HiveConnectorSplit.h
```diff
@@ -205,4 +205,122 @@ struct HiveConnectorSplit : public connector::ConnectorSplit {
   static void registerSerDe();
 };
 
+/// Builder for HiveConnectorSplit. Allows constructing splits incrementally
+/// without specifying every constructor argument.
+class HiveConnectorSplitBuilder {
+ public:
+  explicit HiveConnectorSplitBuilder(std::string filePath)
```
ZacBlanco, line 213: Should connectorId also be a required field on the constructor? What happens if there is no connectorId provided by the user?

ZacBlanco (changes requested):
Change overall LGTM. One other thing though - I know the CI doesn't build all configurations now, but I think there might be some more tests which use the old HiveSplitBuilder in modules that we don't run in CI - for example, the GCS and S3 storage tests. Can you check for usages there to see if they also need to be updated?

You can build locally with something like

```
make <target> CONAN_OPTIONS=" -o bolt/*:enable_gcs=True -o bolt/*:enable_s3=True "
```

I'm going to add an item to my backlog to enable some more of these options in CI, but I'm not sure when I'll get around to it.

yingsu00:
> Change overall LGTM. One other thing though - I know the CI doesn't build all configurations now, but I think there might be some more tests which use the old HiveSplitBuilder in modules that we don't run in CI - for example, the GCS and S3 storage tests. Can you check for usages there to see if they also need to be updated?
> 
> You can build locally with something like
> 
> ```
> make <target> CONAN_OPTIONS=" -o bolt/*:enable_gcs=True -o bolt/*:enable_s3=True "
> ```
> 
> I'm going to add an item to my backlog to enable some more of these options in CI, but I'm not sure when I'll get around to it.

@ZacBlanco bolt/exec/tests/utils/HiveConnectorTestBase.cpp is using the moved HiveConnectorSplitBuilder by:
```
using connector::hive::HiveConnectorSplitBuilder;
```
Same for bolt/exec/tests/HashJoinTest.cpp. In the next PR I will remove the direct reference to any Hive objects in HiveConnectorTestBase.cpp, but I believe it's ok to do so in this PR.

I verified bolt_exec_test builds fine and tests using it run fine too:
```
/Users/yingsu/repo/bytedance1/bolt/_build/debug/bolt/exec/tests/bolt_exec_scan_test --gtest_filter=TableScanTest.allColumns:TableScanTest/*.allColumns:TableScanTest.allColumns/*:*/TableScanTest.allColumns/*:*/TableScanTest/*.allColumns --gtest_color=no
Testing started at 23:21 ...
I0407 23:21:51.900535 23769091 Compression.cpp:643] Initialized zstd compressor with compression level 7
Process finished with exit code 0

/Users/yingsu/repo/bytedance1/bolt/_build/debug/bolt/exec/tests/bolt_exec_infra_test --gtest_filter=AssertQueryBuilderTest.basic:AssertQueryBuilderTest/*.basic:AssertQueryBuilderTest.basic/*:*/AssertQueryBuilderTest.basic/*:*/AssertQueryBuilderTest/*.basic --gtest_color=no
Testing started at 23:19 ...
Process finished with exit code 0

/Users/yingsu/repo/bytedance1/bolt/_build/debug/bolt/exec/tests/bolt_exec_hash_join_test --gtest_filter=MultiThreadedHashJoinTest.bigintArray:MultiThreadedHashJoinTest/*.bigintArray:MultiThreadedHashJoinTest.bigintArray/*:*/MultiThreadedHashJoinTest.bigintArray/*:*/MultiThreadedHashJoinTest/*.bigintArray --gtest_color=no
Testing started at 23:24 ...
Process finished with exit code 0
``` 
The CI would fail if the tests weren't able to find HiveConnectorSplitBuilder. It's nice to have more coverage in the CI, but for the concern you had I think the CI already covers it.

bolt/connectors/hive/HiveConnectorSplit.h
yingsu00: > Should connectorId also be a required field on the constructor? What happens if there is no connectorId provided by the user?

@ZacBlanco Good catch, but I think it's better not to add it to the constructor parameter list.
The HiveConnectorSplitBuilder is not a public general-purpose API, and it’s used in a controlled path (factory):
```
std::shared_ptr<ConnectorSplit> HiveObjectFactory::makeConnectorSplit(...)
  HiveConnectorSplitBuilder builder(filePath);
  builder.connectorId(connectorId()).start(start).length(length);
```
So anybody using the ConnectorObjectFactory would have the connectorId set because it's passed in to its constructor
```
explicit ConnectorObjectFactory(std::string name, std::string connectorId)
      : name_(std::move(name)), connectorId_(std::move(connectorId)) {}
```
Our goal is that anybody outside of the connectors module who wants to create a connector object shall call ConnectorObjectFactory.makeXXX(), so the connectorId will always be enforced. Any connector within the connectors module should not call into another connector directly either. So I think what I can do is to give it a default connectorId in HiveConnectorSplitBuilder:
```
private:
  std::string filePath_;
  std::string connectorId_{kDefaultHiveConnectorId};
  dwio::common::FileFormat fileFormat_{dwio::common::FileFormat::PARQUET};
  uint64_t start_{0};
  uint64_t length_{std::numeric_limits<uint64_t>::max()};
```
What do you think?

ZacBlanco: It makes sense. I am just trying to prevent incorrect usage. I am ok to keep it.

ZacBlanco (approved):

Issue #502, opened 2026-04-14
[Feature] Pre-built JIT IR: compile C++ kernels to bitcode for JIT inlining


zhangxffff:
### Feature Category

SQL Function / Operator

### Problem / Use Case

Writing JIT code in bolt currently requires using the LLVM IRBuilder API directly, which is verbose (e.g., RowContainerCodeGenerator is ~400 lines for row comparison) and hard to maintain.

### Proposed Solution

Introduce a pre-built JIT IR framework that compiles C++ kernel functions to LLVM bitcode at build time and inlines them into JIT-composed modules at runtime.

Key components:
  1. Build pipeline: kernels.cpp → clang -emit-llvm -c -O2 → xxd -i → embedded byte array. Compile flags are extracted from ninja build commands via gen_kernel_flags.sh to support bolt headers (e.g., DecodedVector.h).
  2. Runtime linking: PrebuiltIR::load() parses embedded bitcode, ThrustJITv2::linkPrebuiltModule() links it into JIT modules with internal linkage to avoid symbol conflicts.
  3. IR optimization: AlwaysInlinerLegacyPass + InstCombine/GVN/CFGSimplification in ThrustJITv2's IR transform layer.

### Importance

High (Needed for production)

### Willingness to Contribute

Yes, I can submit a PR

PR #513, opened 2026-04-17
feat: add clang support to llvm-core Conan recipe

zhangxffff:
Add with_clang option to llvm-core recipe that builds clang alongside LLVM from the same source. This ensures exact version match between clang and LLVM for bitcode compatibility in the pre-built JIT IR framework.

Changes:
- scripts/conan/patches/llvm-core-add-clang-support.patch: adds with_clang option, clang source download, LLVM_ENABLE_PROJECTS, and clang library exclusion from components
- conanfile.py: enables with_clang=True when JIT is enabled

Issue Number: related #502 

### Type of Change

- [ ] 📝 Documentation only

### Description

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [ ] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>


scripts/conan/patches/llvm-core-add-clang-support.patch
```diff
@@ -0,0 +1,72 @@
...
+         graphviz_options = textwrap.dedent(f"""
+             set(GRAPHVIZ_EXECUTABLES OFF)
+@@ -547,3 +556,7 @@ class LLVMCoreConan(ConanFile):
+         else:
+             self.cpp_info.set_property("cmake_target_name", "LLVM")
+             self.cpp_info.libs = collect_libs(self)
++
++        if self.options.with_clang:
++            bindir = os.path.join(self.package_folder, "bin")
++            self.buildenv_info.prepend_path("PATH", bindir)
```
ZacBlanco: From the conan documentation at https://docs.conan.io/2/reference/conanfile/methods/package_info.html

> It is not necessary to add bindirs to the PATH environment variable, this will be automatically done by the consumer VirtualBuildEnv and VirtualRunEnv generators. Likewise, it is not necessary to add includedirs, libdirs or any other dirs to environment variables, as this information will be typically managed by other generators.

It looks like this is not necessary?

ZacBlanco (commented):
Just one question

scripts/conan/patches/llvm-core-add-clang-support.patch
ZacBlanco: Actually, I am mistaken as `bindir` here is just the variable. But maybe LLVM should export it's bindir to the conan package so that the binaries are propagated?

On the same page, the example is

```
self.cpp_info.bindirs = ['bin']  # Directories where executables and shared libs can be found
```

ZacBlanco:
@zhangxffff Have you tested that the new recipe works for PR #501 ? Also, I think something is not correct with the patch, because I would expect that adding this patch would cause the CI to rebuild the llvm-core recipe, but it seems like it's not rebuilding it. Can you look into that?

zhangxffff:
> @zhangxffff Have you tested that the new recipe works for PR #501 ? Also, I think something is not correct with the patch, because I would expect that adding this patch would cause the CI to rebuild the llvm-core recipe, but it seems like it's not rebuilding it. Can you look into that?

For the question before, we use `LLVM_TOOLS_BINARY_DIR` in `CMakeLists.txt` to locate `clang`, so you should be right, adding it to PATH is not necessary.
https://github.com/conan-io/conan-center-index/blob/f7e534b5a37d22627b73f7a5e009a5eb20a1d4ed/recipes/llvm-core/all/conanfile.py#L468

The reason `llvm-core` does not rebuild is that Conan appears to treat the remote version and the newly patched version as having the same recipe ID, even though we added a `with_clang` option. So in #501, I renamed the `llvm-core` version to `19.1.7-clang` to force a rebuild. I will test in #501 to make sure the latest changes work. After this patch is merged, I will rename it back to `19.1.7`.

zhangxffff:
> @zhangxffff Have you tested that the new recipe works for PR #501 ? Also, I think something is not correct with the patch, because I would expect that adding this patch would cause the CI to rebuild the llvm-core recipe, but it seems like it's not rebuilding it. Can you look into that?

OK, it seems that modify patch does not retigger CI, I will look at it.

zhangxffff:
@ZacBlanco The reason we cannot trigger a rebuild for `llvm-core` is that the remote Conan server was added with `--index=0`, giving it the highest priority when resolving `llvm-core/19.1.7`. Once Conan finds a matching version on the remote server, it uses that instead of the recipe we patched in `bolt-cci-local`.

```
  CI adds its remote at index 0 (highest priority):
  0: ci              ← pre-built binaries (stale, unpatched recipes)
  1: bolt-local      ← custom recipes
  2: bolt-cci-local  ← patched CCI checkout
  3: conancenter
```

I tried adding `--index=0` to `bolt-cci-local`, but that would cause all recipes to be built from source, since `bolt-cci-local` contains no prebuilt binaries.

So what we do now is add a step in the GitHub Action to extract modified recipes from the changed code and move them into a separate local Conan server `bolt-patched` with the highest priority. This way, only the modified recipes will be rebuilt.


  Result for a PR that modifies llvm-core:
```
  0: bolt-patched    ← only llvm-core (different revision → rebuild)
  1: ci              ← everything else (cached binaries → reused)
  2: bolt-local
  3: bolt-cci-local
  4: conancenter
```

Action output shows that llvm-core was successfully rebuilt.
```
Run ./.github/actions/bolt-build-base/override-changed-recipes.sh "${GITHUB_BASE_REF:-main}"
✅ Override recipes: llvm-core
```

.github/actions/bolt-build-base/override-changed-recipes.sh
ZacBlanco, line 1: I think this script is too complex to be worth committing and maintaining. I would rather have the patch update the LLVM-core recipe version to `19.1.7-bolt` instead, and then just have bolt's conanfile reference `19.1.7-bolt`. No external conan remote will have that version, so it will be forced to get picked from the CCI local recipe index.

ZacBlanco (changes requested):
The recipe changes themselves LGTM, but I think the patch should just update the llvm-core recipe version as well so we don't conflict with the OSS conan recipe version.

.github/actions/bolt-build-base/override-changed-recipes.sh
zhangxffff: I agree with updating the version for the `llvm-core` recipe. However, without this script, changes to `scripts/conan/patches` or `scripts/conan/recipes` would not trigger a rebuild in CI unless the version is also updated. Since we always upload used conan recipes to the `ci` server, and it is always searched first, this overrides the CCI local recipe index and prevents us from picking up recipes from CCI local even we have a version that differs from official conan server.

https://github.com/bytedance/bolt/blob/17a2d9b1a0f669a7d29cc7f5696af23e845deb5f/.github/workflows/conan-upload.yml#L75-L81

If we remove this script, we would need to establish a convention that any modification to a recipe must be accompanied by a corresponding version update to ensure CI rebuilds the changed recipe.

ZacBlanco: I think having a convention where we update the recipe version is not a bad, but is somewhat tedious. I guess I am ok with this script for now then. If it presents issues in the future, we might want to look at other solutions.

Anyways, we can merge this then. Can you test on your other PR to make sure that this recipe works before merging though?

ZacBlanco (approved):

ZacBlanco:
@zhangxffff once you can confirm the recipe works for #501 we can merge this, upload the new recipe to CI conan server, and then you can rebase 501

PR #536, opened 2026-04-30
fix: license check bugs

ZacBlanco:

- Disallow SPDX-Identifier header to enforce full license in header
- fix issue in check_license.go which silently passed non-compliant headers
- Automatically apply correct license header to non-compliant files. This is the same as the clang-format check.
- Update pre-commit-config ignore list with files who were vendored into bolt without modifications and retain their Apache 2.0 SPDX header or don't require a header at all

### Type of Change

- [ ] 📝 Documentation only

### Description

The license header check was not behaving correctly.

### DIFF
--- a/.github/license_check/check_license.go
+++ b/.github/license_check/check_license.go
@@ -99,5 +99,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	for file := range result.Failure {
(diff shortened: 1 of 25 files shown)


markjin1990 (approved):
LGTM!

PR #543, opened 2026-05-08
[Expr][Gluten] Fix decimal ceil remainder overflow

### LINKED ISSUE #544: [Bug] ceil(long decimal) returns incorrect result
### Component Selection

- [ ] Build
- [ ] Other

## Problem

  `ceil` and `floor` share decimal rounding logic that computes a quotient and a fractional remainder. The old implementation stored the remainder using the result type `R`.

  This is incorrect when the input is a long decimal but the result is a short decimal. In that case, the fractional remainder can be wider than `int64_t`, even though the final rounded result fits in
  the short decimal result type.

  For example, `ceil(100.10010010010010010010010010)` could incorrectly return `100` instead of `101`.

  ## Root Cause

  When R is int64_t, a large-scale fractional remainder from an int128_t decimal input may overflow or truncate before the positive-remainder check. As a result, ceil may fail to detect that the value
  has a non-zero fractional part.

### Reproduction Steps

`ceil(100.10010010010010010010010010)` could incorrectly return `100` instead of `101`

### Bolt Version / Commit ID

main@e2c93b3107efd69979819e7d43e8e2c7e41f4928

### System Configuration

```markdown
system irrelevant
```

### Logs / Stack Trace

```shell

```

### Expected Behavior

_No response_

### Additional context

_No response_

wangxinshuo-bolt:

Issue Number: close https://github.com/bytedance/bolt/issues/544

### Type of Change

- [ ] 📝 Documentation only

### Description

  Store the remainder using the divisor type B, which is wide enough for the decimal scale involved in the operation:
```
  B remainder = unsignedDividendRescaled % unsignedDivisor;
```
  This keeps the rounding decision based on the full decimal-scale remainder while preserving the original result type for the final quotient.

  Added regression coverage for the long-decimal ceil case and moved/refactored related decimal vector function tests so the Spark decimal test target includes the relevant coverage.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>


bolt/functions/sparksql/tests/DecimalVectorFunctionTest.cpp
```diff
@@ -1,162 +0,0 @@
-/*
```
markjin1990, line 1: @wangxinshuo-bolt Is deleting this test file intentional or by accident?

XinShuoWang: > [@wangxinshuo-bolt](https://github.com/wangxinshuo-bolt) Is deleting this test file intentional or by accident?

It was intentional to delete this. If not, we would have two test files (`DecimalVectorFunctionTest.cpp` and `DecimalVectorFunctionsTest.cpp`) with similar names and functionalities, which would be confusing. All the UTs have been preserved, just moved to more appropriate locations.

markjin1990 (approved):

PR #553, opened 2026-05-13
refactor: align PrimitiveKind enum values with TypeKind in CastExpr-tpl

mikasa1110:
#### Summary

Replace the internal `PrimitiveKind` enum in `CastExpr-tpl.cpp` with `CastKindType<TypeKind, Decoration>` — a type-level construct that eliminates the duplicated enum by directly using `TypeKind` from `Type.h`.

The old `PrimitiveKind` enum mirrored `TypeKind` with different integer values, requiring two 30+ line switch statements (`getPrimitiveKind`, `getInnerKind`) to map between them. By switching from a non-type template parameter (enum value) to a type template parameter (`CastKindType`), the mapping collapses to a single compact `getCastKind()` function with a fall-through switch.

`Decoration` carries the extra semantic info for three "virtual types" — `ShortDecimal`, `LongDecimal`, `Date` — that share physical `TypeKind` storage with `BIGINT`/`HUGEINT`/`INTEGER` but require distinct casting behavior.

### Type of Change

- [ ] 📝 Documentation only

### Description

- **Delete** `PrimitiveKind` enum, 11 `static_assert`s, `alignedWith`, `isSupportedTypeKind`.
- **Add** `Decoration` enum, `CastKindType<TypeKind Storage, Decoration Deco>` struct, `CastKindKey` runtime struct, and 13 type aliases (`BooleanKind`, `TinyintKind`, ..., `DateKind`).
- **Replace** `getPrimitiveKind()` + `getInnerKind()` with a single `getCastKind()`.
- **Convert** all template parameters from `PrimitiveKind kind` (non-type) to `typename FromKind`/`typename ToKind` (type), with classification helpers (`isInteger`, `isFloat`, `isDecimal`) keyed off `K::storage` and `K::deco`.
- **Rewrite** registration infrastructure: delete `ConverterRegister`, `get_last`, `get_front`; replace with `KindList` + fold expressions (`registerForOneFrom`, `RegisterAllPairs`)

### Performance Impact

    <details>
    </details>

### Release Note

Release Note:

N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)


bolt/expression/CastExpr-tpl.cpp
```diff
@@ -57,92 +57,51 @@ enum class ConvertStatus : int8_t {
...
+  TINYINT = static_cast<int8_t>(TypeKind::TINYINT),
+  SMALLINT = static_cast<int8_t>(TypeKind::SMALLINT),
+  INTEGER = static_cast<int8_t>(TypeKind::INTEGER),
+  BIGINT = static_cast<int8_t>(TypeKind::BIGINT),
+  REAL = static_cast<int8_t>(TypeKind::REAL),
+  DOUBLE = static_cast<int8_t>(TypeKind::DOUBLE),
+  STRING = static_cast<int8_t>(TypeKind::VARCHAR),
+  BINARY = static_cast<int8_t>(TypeKind::VARBINARY),
+  TIMESTAMP = static_cast<int8_t>(TypeKind::TIMESTAMP),
+  SHORT_DECIMAL = 11,
```
zhangxffff: These values introduce an implicit numeric contract between PrimitiveKind and TypeKind that nothing enforces. Any future change to TypeKind can silently break the cast layer with no compiler warning.

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -57,92 +57,51 @@ enum class ConvertStatus : int8_t {
...
-      return PrimitiveKind::LONG_DECIMAL;
-    case TypeKind::UNKNOWN:
-      return PrimitiveKind::UNKNOWN;
-    default:
-      BOLT_FAIL("Unknown type {}", type->toString());
+  auto kind = type->kind();
+  if (kind == TypeKind::HUGEINT) {
+    return PrimitiveKind::LONG_DECIMAL;
   }
+  return static_cast<PrimitiveKind>(kind);
```
zhangxffff: The default: `BOLT_FAIL("Unknown type {}", ...)` in getPrimitiveKind is gone. Unsupported kinds now silently produce out-of-range enum values instead of failing explicitly.

bolt/expression/CastExpr-tpl.cpp
mikasa1110: thx for pointing out. I will try to fix these two problems

mikasa1110: I've made the following fixes: Added static_asserts and isSupportedTypeKind util for BOLT_FAIL guard. Please let me know if you'd like me to adjust anything

zhangxffff:
@mikasa1110 Thanks you very much for your commit, but it seems simply aligning `PrimitiveKind`'s enum values with `TypeKind` doesn't solve the real problem. `PrimitiveKind` still exists and introduces more implicit coupling between the two enums. Maybe we should rethink this approach — rather than just changing the enum values, we should replace `PrimitiveKind` with `TypeKind` along with additional decorators to convey the extra type information, and add sufficient guards to assert against unexpected combinations.

In C++20, we can use `struct` as `NTTP(Non-Type Template Parameter)`, so we can define something like
```C++
enum class Decoration : int8_t { None, ShortDecimal, LongDecimal, Date };

struct CastKind {
  TypeKind   storage;
  Decoration deco = Decoration::None;
  constexpr bool operator==(const CastKind&) const = default;
};
```

and then replace `PrimitiveKind` with `CastKind`.

But now bolt is still C++17, we can only use type parameter in template to annotate TypeKind with Decoration. like:
```C++
enum class Decoration : int8_t { None, ShortDecimal, LongDecimal, Date };

template <TypeKind Storage, Decoration Deco = Decoration::None>
struct CastKindType {
  static constexpr TypeKind   storage = Storage;
  static constexpr Decoration deco    = Deco;
};
```

And then replace all Non-Type template parameter `PrimitiveKind` to type template parameter `CastKindType`.

I think this is closer to what we described in issue #492. What are your thoughts? @mikasa1110 @ZacBlanco

mikasa1110:
Thanks!  I will carefully rethink this problem based on your valuable feedback. I will try to completely eliminate the duplicated PrimitiveKind enum by replacing it with CastKindType. And I will proceed with this approach and update the PR accordingly.

mikasa1110:
@zhangxffff I've taken your suggestion and updated the pr. Please help to review again.  Really appreciate your detailed feedback and time on this!

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -978,110 +957,66 @@ class VectorConverter : public ConverterBase {
...
-  static_assert(N > 0, "Array cannot be empty");
-  return arr[N - 1];
-}
-
-template <typename T, size_t N, size_t... Indices>
-constexpr auto get_front_impl(
-    const std::array<T, N>& arr,
-    std::index_sequence<Indices...>) {
-  return std::array<T, sizeof...(Indices)>{arr[Indices]...};
+std::map<std::pair<CastKindKey, CastKindKey>, std::shared_ptr<ConverterBase>>
```
zhangxffff, line 958: The `CastKindKey` declaration can be moved here, since it is only used as a key in the runtime map. Placing it closer to its usage makes the code clearer.

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -56,127 +56,108 @@ enum class ConvertStatus : int8_t {
...
-using PrimitiveNativeType =
-    typename TypeTraits<getInnerKind<type>()>::NativeType;
+template <typename CK>
+using PrimitiveNativeType = typename TypeTraits<CK::storage>::NativeType;
 
-template <PrimitiveKind type>
+template <typename CK>
 constexpr bool isStringLikeKind =
-    (type == PrimitiveKind::STRING || type == PrimitiveKind::BINARY);
+    CK::storage == TypeKind::VARCHAR || CK::storage == TypeKind::VARBINARY;
```
zhangxffff: Since we already have `BooleanKind` here, it would be better to use `std::is_same_v<CK, BooleanKind>` instead of comparing storage type kinds, as comparing storage type kinds could be error-prone for types like BIGINT/SHORT_DECIMAL/DATE.

zhangxffff (commented):
@mikasa1110 Excellent work on this refactor, thanks a lot for your effort.

bolt/expression/CastExpr-tpl.cpp
mikasa1110: Thanks for pointing this out. Updated and changed similar codes

bolt/expression/CastExpr-tpl.cpp
mikasa1110: done

mikasa1110:
Thanks @zhangxffff, I resovled mentioned comments above. Please take a look when you have free time.

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -296,67 +232,63 @@ std::optional<bool> sparkStringToBoolean(const folly::StringPiece& str) {
...
-  template <PrimitiveKind K = toKind> \
-  FOLLY_ALWAYS_INLINE std::enable_if_t<cond<K>, ConvertStatus>
+#define TO_IF(cond)               \
+  template <typename CK = ToKind> \
+  FOLLY_ALWAYS_INLINE std::enable_if_t<cond<CK>, ConvertStatus>
 
  public:
   Converter(exec::EvalCtx& context, TypePtr fromType, TypePtr toType) {
-    if (fromKind == PrimitiveKind::SHORT_DECIMAL) {
+    if (FromKind::deco == Decoration::ShortDecimal) {
```
zhangxffff: these 4 equal can also replace with std::is_same_v

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -601,14 +533,14 @@ class Converter {
             from, fromScale_, fromDecimalMaxSize_, cached);
         to.set(std::string_view(cached, strSize));
       }
-    } else if constexpr (fromKind == PrimitiveKind::DATE) {
+    } else if constexpr (FromKind::deco == Decoration::Date) {
```
zhangxffff: This equal can also replace with std::is_same_v

bolt/expression/CastExpr-tpl.cpp
```diff
@@ -694,23 +626,23 @@ class Converter {
         return nullOutput ? ConvertStatus::OTHER_FAILURE
                           : ConvertStatus::SUCCESS;
       }
-    } else if constexpr (fromKind == PrimitiveKind::DATE) {
+    } else if constexpr (FromKind::deco == Decoration::Date) {
```
zhangxffff: And this.

zhangxffff (commented):
Just a minor comment: some of the equality checks could also be replaced with `std::is_same_v`. Everything else looks great.

bolt/expression/CastExpr-tpl.cpp
mikasa1110: updated

mikasa1110:
Thx a lot. I removed direct type comparisions using deco and fixed a compiling problem triggered by C++20 default operator==.

zhangxffff (approved):
LGTM! Thanks very much for your contribution.

PR #557, opened 2026-05-13
[fix] limit skewed hash build spill file rows

### LINKED ISSUE #556: [Bug] HashBuild OOM caused by  oversized spill files
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

In HashBuild row-based spill with skew partition support enabled, a skewed partition can still produce oversized spill files. During restore, this can lead to excessive memory pressure and potentially OOM.

  A typical case is:

    skew partition 6, rowCount = 8167
    desired restore granularity ~ 2546 rows per range partition

  However, skew splitting is currently file-granular. If the skewed partition only ends up with 2 spill files, it can only be split into 2 range partitions, even when row-count-based estimation suggests 4. This
  leaves each restored partition too large.

### Reproduction Steps

run bytedance internal sql

### Bolt Version / Commit ID

main

### System Configuration

```markdown
```

### Logs / Stack Trace

```shell

```

### Expected Behavior

_No response_

### Additional context

_No response_

guhaiyan0221:

Issue Number: close #556 

### Type of Change

- [ ] 📝 Documentation only

## Problem

  In HashBuild row-based spill with skew partition support enabled, a skewed partition can still produce oversized spill files. During restore, this can lead to excessive memory pressure and potentially OOM.

  A typical case is:

    skew partition 6, rowCount = 8167
    desired restore granularity ~ 2546 rows per range partition

  However, skew splitting is currently file-granular. If the skewed partition only ends up with 2 spill files, it can only be split into 2 range partitions, even when row-count-based estimation suggests 4. This
  leaves each restored partition too large.

  --------

  ## Root Cause

  In the row-container spill path, fillSpillRuns() keeps accumulating rows into per-partition spillRuns_ until maxSpillRunRows_ (or EOF). For skewed data, one partition may accumulate too many rows in a single
  run.

  runSpill() then writes that run to disk. Although prepareForRangPartitionIfNeeded() can detect skew and flush files, it runs after writing, which is too late to prevent an oversized file from being created.

  Also, HashBuild’s true restore-bound (maxRowsInMemory_) is only set near finishSpill(), i.e., after spill writing has already happened. So it cannot be used as an early file-size limiter during write.

  --------

  ## Solution

  This change introduces an early run-level guard in Spiller::spill(const RowContainerIterator*) for the row-container path, using current average rows per partition:

    maxRowsPerPartition = container_->numRows() / state_.maxPartitions()

  ### Strategy

  1. Enable this only when supportSkewPartition_ is true.
  2. On the first row-container spill, compute maxRowsPerPartition from current container rows and partition count.
  3. In fillSpillRuns(), stop filling the current run once any partition reaches this per-partition row target.
  4. runSpill() writes the bounded run as usual.
  5. Existing prepareForRangPartitionIfNeeded() logic is reused to detect skew victim and flush current file.
  6. After a skew victim is identified, subsequent row-container spill calls continue to enforce the bound using maxRowsPerFile_, preventing later oversized skew files.

  Additionally, skewedVictim_ is migrated from sentinel -1 to std::optional<uint32_t> for clearer semantics.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/exec/Spiller.cpp
+++ b/bolt/exec/Spiller.cpp
@@ -813,13 +813,19 @@ void Spiller::spill(const RowContainerIterator* startRowIter) {
   BOLT_CHECK(
       type_ != Type::kHashJoinProbe && type_ != Type::kOrderByOutput &&
       type_ != Type::kLocalMergeInput);
-  const bool doPotentialRangePartitionCheck =
-      (supportSkewPartition_ && container_ &&
-       container_->numRows() >
-           std::max(maxSpillRunRows_, kDefaultSpillRunRows));
   uint64_t totalTimeUs{0};
   {
     MicrosecondTimer timer(&totalTimeUs);
+
+    uint64_t maxRowsPerPartition{0};
+    if (supportSkewPartition_) {
+      if (skewedVictim_.has_value() && maxRowsPerFile_ > 0) {
+        maxRowsPerPartition = maxRowsPerFile_;
+      } else if (!isAnySpilled()) {
+        maxRowsPerPartition =
+            bits::divRoundUp(container_->numRows(), state_.maxPartitions());
+      }
+    }
     markAllPartitionsSpilled();
 
     RowContainerIterator rowIter;
@@ -829,9 +835,9 @@ void Spiller::spill(const RowContainerIterator* startRowIter) {
 
     bool lastRun{false};
     do {
-      lastRun = fillSpillRuns(&rowIter);
+      lastRun = fillSpillRuns(&rowIter, maxRowsPerPartition);
       runSpill(lastRun);
-      if (doPotentialRangePartitionCheck || skewedVictim_ >= 0) {
+      if (supportSkewPartition_ || skewedVictim_.has_value()) {
         prepareForRangPartitionIfNeeded();
       }
     } while (!lastRun);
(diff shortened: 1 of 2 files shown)


bolt/exec/Spiller.cpp
```diff
@@ -829,9 +835,9 @@ void Spiller::spill(const RowContainerIterator* startRowIter) {
 
     bool lastRun{false};
     do {
-      lastRun = fillSpillRuns(&rowIter);
+      lastRun = fillSpillRuns(&rowIter, maxRowsPerPartition);
       runSpill(lastRun);
-      if (doPotentialRangePartitionCheck || skewedVictim_ >= 0) {
+      if (supportSkewPartition_ || skewedVictim_.has_value()) {
```
fzhedu, line 840: is the condition equal to `doPotentialRangePartitionCheck`, as it considers `container_`

guhaiyan0221: we do not need an extra container_ null check here. The Spiller constructor already enforces that only HashJoinProbe / HashJoinProbeMatchFlag / LocalMergeInput spillers can have a null container_.
  This path is the row-container spill path, and fillSpillRuns() already directly uses container_->listRows/hash/rowSize. Therefore, a non-null container_ is a precondition of this path

fzhedu (approved):
LGTM

Issue #107, opened 2026-01-12
[Feature]  Iceberg Connector for Bolt


yingsu00:
Apache Iceberg is a modern, open table format with snapshot isolation, schema evolution, partition evolution, and efficient metadata handling. Iceberg is widely adopted for analytic workloads across engines (Spark, Trino, Flink) and is increasingly required for Bolt’s analytical query use cases. This issue describes the design of a new Iceberg connector that supports reading and writing Iceberg tables on all Iceberg versions. The work consists of several areas:

1. Support reading Iceberg tables with positional deletes (for backword support reading tables written by legacy engines)
2. Support reading Iceberg tables with equality deletes 
3. Support inserting into Iceberg tables
4. Support writing delete files defined in Iceberg spec V2 and V3

## 1. Design Goals

Because of some historical reasons, our previous Iceberg contributions to Velox have been tied to the Hive connector, reusing Hive abstractions (splits, table handles, column handles) and implementation (HiveDataSource, HiveSplitReader, etc.). This coupling leads to:

- No connector level configurations
- Hive-specific abstractions polluting the generic table format code,
- Inability to cleanly represent Iceberg-specific concepts like evolving PartitionSpecs or Iceberg metadata fields,
- Difficulty in expressing Iceberg semantics (e.g., multi-snapshot reads, delete files) without bending Hive types.

Thus, Bolt needs a standalone Iceberg connector, decoupled from Hive, to cleanly support Iceberg reads and writes, being composable and embeddable in the Bolt execution stack. The design goals of the new Iceberg connectors include:

- Native modular and pluggable Iceberg support . The new Iceberg connector should be a standalone library component, usable without pulling in Hive connector code.
- First-class Iceberg types: table handle, partition spec, split metadata, delete file handling.
- Extensible for reads/writes and future features (e.g., multi-engine compatibility, Iceberg catalog integration).
- Performance: preserve and improve on existing read performance by utilizing Bolt's existing filter pushdown, introducing metadata and result caching.

# 2. Architecture
## 2.1 Iceberg Connector 
We will add a new Icberg folder under connectors/
```
bolt/connectors/ 
├─ connector.h/cpp ← connector interfaces
├─ common/
|    ├─ DataSourceBase
|    ├─ DataSinkBase
|    ├─ SplitReaderBase
|    ├─ TableHandleBase
|    ├─ ConnectorSplitBase
|    ├─ TableHandleBase
|    ├─ ColumnHandleBase
├─ hive/
[...]

PR #567, opened 2026-05-18
refactor(tests): Make exec tests connector-agnostic

yingsu00:

Decouple exec tests from the Hive-specific HiveConnectorTestBase so they can
  run against any connector via parameterization, and fix the layering so
  connector test scaffolding no longer depends on exec.
  
Issue Number: https://github.com/bytedance/bolt/issues/107

### Type of Change

- [ ] 📝 Documentation only

### Description
- Move generic helpers to OperatorTestBase — writeToFile, makeVectors,
  makeFilePaths had no Hive specifics; they belong on the generic exec fixture,
   not the connector one.
  - Explicit connector factory registration — add
  register{Hive,Tpch}ConnectorFactories() so tests can bootstrap connectors by
  name instead of relying on static-init side effects.
  - Introduce connector::test::ConnectorTestBase (bolt/connectors/tests/utils/,
   lib bolt_connector_test_lib) — a connector-agnostic fixture that inherits
  only ::testing::Test + VectorTestBase. No dependency on bolt_exec_test_lib or
   any specific connector. Legacy HiveConnectorTestBase becomes a thin wrapper
  that reuses its registration helpers.
  - Add ConnectorTestParam + paramFor/paramsFor — parameterization plumbing so
  exec tests select connectors by name string, never including
  connector-specific headers.
  - Relocate TempFilePath/TempDirectoryPath to bolt/common/testutil/ (namespace
   bytedance::bolt::test) — they're generic FS test utilities; keeping them
  under exec/tests/utils/ forced a layering inversion. (~119-file mechanical
  sweep.)
  - Migrate 12 read-side exec tests to OperatorTestBase + 
  WithParamInterface<ConnectorTestParam> (TEST_F→TEST_P,
  makeHiveConnectorSplit→connector::test::makeConnectorSplit, Hive-only
  sub-cases extracted to bolt/connectors/hive/tests/).
  
 #### TODO
  
  HashJoinTest/TableScanTest (heaviest), write-side tests, PlanBuilder,
  dwio/common/tool migrations, and removal of the legacy HiveConnectorTestBase
  — these are follow-up commits in the same series.
  
 #### Notes
  
  - No exec test source includes bolt/connectors/hive/* headers anymore.
  - Each commit builds independently (CI hooks bypassed locally only due to a
  pre-existing unrelated LLVM/JIT compile break in bolt/jit/ThrustJITv2.cpp).
### Performance Impact

### Release Note

Release Note:
N/A

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/.github/workflows/pre-commit-checks.yml
+++ b/.github/workflows/pre-commit-checks.yml
@@ -60,6 +60,18 @@ jobs:
       - name: Generate Compile Database
         run: |
           make compile_db_all
+      - name: Ensure generated headers exist
+        run: |
+          # _compile_db (invoked by compile_db_all) is supposed to build these,
+          # but make it explicit so clang-tidy in the next step never races
+          # against a missing dwrf_schema.pb.h / parquet thrift header. If the
+          # underlying chain already produced them, this is a no-op rebuild.
+          cmake --build --preset conan-release \
+            --target generate_parquet_thrift \
+            --target bolt_dwio_dwrf_proto
+          test -f _build/Release/bolt/dwio/dwrf/proto/dwrf_schema.pb.h \
+            || { echo "::error::dwrf_schema.pb.h missing after proto build" >&2; \
+                 ls -la _build/Release/bolt/dwio/dwrf/proto/ || true; exit 1; }
       - name: Run pre-commit hooks
         run: |
           pre-commit run --all-files
--- a/bolt/connectors/tests/utils/ConnectorTestBase.cpp
+++ b/bolt/connectors/tests/utils/ConnectorTestBase.cpp
@@ -1,0 +1,196 @@
+/*
+ * Copyright (c) ByteDance Ltd. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "bolt/connectors/tests/utils/ConnectorTestBase.h"
+
+#include <folly/dynamic.h>
+
+#include "bolt/common/file/FileSystems.h"
+#include "bolt/common/serialization/Serializable.h"
+#include "bolt/type/filter/FilterBase.h"
+
+namespace bytedance::bolt::connector::test {
+
+namespace {
+
+std::string toEffectivePath(const std::string& filePath) {
+  return filePath.find('/') == 0 ? "file:" + filePath : filePath;
+}
+
+uint64_t fileSize(const std::string& filePath) {
+  auto file =
+      filesystems::getFileSystem(filePath, nullptr)->openFileForRead(filePath);
+  return static_cast<uint64_t>(file->size());
+}
+
+} // namespace
+
+void registerTestConnector(
+    const std::string& connectorName,
+    const std::string& connectorId,
+    folly::Executor* ioExecutor,
+    const std::shared_ptr<const config::ConfigBase>& config,
+    const FactoryRegistrar& factoryRegistrar) {
+  if (factoryRegistrar && !connector::hasConnectorFactory(connectorName)) {
+    factoryRegistrar();
+  }
+  auto factory = connector::getConnectorFactory(connectorName);
+  factory->registerObjectFactory(connectorId);
+  connector::registerConnector(
+      factory->newConnector(connectorId, config, ioExecutor));
+}
+
+void unregisterTestConnector(
+    const std::string& connectorName,
+    const std::string& connectorId) {
+  connector::unregisterConnector(connectorId);
+  connector::unregisterConnectorObjectFactory(connectorName);
+}
+
+ConnectorTestBase::ConnectorTestBase(
+    std::string connectorName,
+    std::string connectorId,
+    FactoryRegistrar factoryRegistrar)
+    : connectorName_(std::move(connectorName)),
+      connectorId_(std::move(connectorId)),
+      factoryRegistrar_(std::move(factoryRegistrar)) {}
+
+ConnectorTestBase::~ConnectorTestBase() = default;
+
+void ConnectorTestBase::SetUp() {
+  Type::registerSerDe();
+  common::Filter::registerSerDe();
+  core::ITypedExpr::registerSerDe();
+  ioExecutor_ = std::make_unique<folly::IOThreadPoolExecutor>(3);
+  auto emptyConfig = std::make_shared<config::ConfigBase>(
+      std::unordered_map<std::string, std::string>());
+  registerTestConnector(
+      connectorName_,
+      connectorId_,
+      ioExecutor_.get(),
+      emptyConfig,
+      factoryRegistrar_);
+  filesystems::registerLocalFileSystem();
+}
+
+void ConnectorTestBase::TearDown() {
+  // Make sure all pending loads finish or are cancelled before unregistering
+  // the connector.
+  ioExecutor_.reset();
+  unregisterTestConnector(connectorName_, connectorId_);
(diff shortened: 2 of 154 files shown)


.github/workflows/pre-commit-checks.yml
```diff
@@ -60,6 +60,18 @@ jobs:
...
+          # _compile_db (invoked by compile_db_all) is supposed to build these,
+          # but make it explicit so clang-tidy in the next step never races
+          # against a missing dwrf_schema.pb.h / parquet thrift header. If the
+          # underlying chain already produced them, this is a no-op rebuild.
+          cmake --build --preset conan-release \
+            --target generate_parquet_thrift \
+            --target bolt_dwio_dwrf_proto
+          test -f _build/Release/bolt/dwio/dwrf/proto/dwrf_schema.pb.h \
+            || { echo "::error::dwrf_schema.pb.h missing after proto build" >&2; \
+                 ls -la _build/Release/bolt/dwio/dwrf/proto/ || true; exit 1; }
```
ZacBlanco, line 74: This shouldn't be necessary as it's already taken care of in the makefile

bolt/connectors/tests/utils/ConnectorTestBase.cpp
```diff
@@ -0,0 +1,364 @@
...
+  auto file =
+      filesystems::getFileSystem(filePath, nullptr)->openFileForRead(filePath);
+  const int64_t fileSize = file->size();
+  const auto splitSize =
+      static_cast<uint64_t>((fileSize + splitCount - 1) / splitCount);
+  std::vector<std::shared_ptr<connector::ConnectorSplit>> splits;
+  splits.reserve(splitCount);
+  auto factory = connector::getConnectorObjectFactory(connectorName);
+  const auto effectivePath = toEffectivePath(filePath);
+  for (uint32_t i = 0; i < splitCount; ++i) {
```
ZacBlanco, line 141: Correct me if I am wrong, but can we actually do this? Don't files need to be split by row groups? or can they be split by an arbitrary number of records?

ZacBlanco: Also, what about the case if the file can't be split evenly?

bolt/connectors/tests/utils/ConnectorTestBase.cpp
```diff
@@ -0,0 +1,364 @@
...
+  std::vector<std::shared_ptr<connector::ConnectorSplit>> splits;
+  splits.reserve(splitCount);
+  auto factory = connectorObjectFactory();
+  const auto effectivePath = toEffectivePath(filePath);
+  for (uint32_t i = 0; i < splitCount; ++i) {
+    splits.emplace_back(factory->makeConnectorSplit(
+        effectivePath,
+        i * splitSize,
+        splitSize,
+        connector::makeOptions({{"fileFormat", static_cast<int>(format)}})));
```
ZacBlanco, line 298: This seems to be a copy of logic from an earlier implementation. Can we combine the implements to prevent duplication?

bolt/core/Expressions.cpp
```diff
@@ -123,6 +124,9 @@ TypedExprPtr ConstantTypedExpr::create(
   std::istringstream dataStream(serializedData);
 
   auto* pool = static_cast<memory::MemoryPool*>(context);
+  if (pool == nullptr) {
+    pool = memory::MemoryManager::getInstance()->tracePool();
```
ZacBlanco, line 128: Can you explain why this change is needed?

bolt/connectors/tests/utils/ConnectorTestBase.cpp
```diff
@@ -0,0 +1,364 @@
...
+    uint64_t start,
+    uint64_t length) {
+  return connector::getConnectorObjectFactory(connectorName)
+      ->makeConnectorSplit(
+          toEffectivePath(filePath),
+          start,
+          length,
+          connector::makeOptions(
+              {{"fileFormat",
+                static_cast<int>(dwio::common::FileFormat::DWRF)}}));
```
ZacBlanco, line 163: nit: can we make all the default file formats in this Parquet?

scripts/run-clang-tidy.py
```diff
@@ -472,6 +472,37 @@ def _compute_changed_lines() -> Optional[Multimap]:
...
+                    for _o in _orphans:
+                        print(f"  {_o}")
+                files_to_process = _kept
+            except (json.JSONDecodeError, KeyError, OSError) as _e:
+                print(
+                    f"Warning: could not parse {db_path} for orphan filter: {_e}",
+                    file=sys.stderr,
+                )
+
         if not files_to_process:
```
ZacBlanco, line 506: My understanding of the issue is that this stems from headers not being in the compile-commands.json file. I have run into similar issues before. Maybe we can just simply filter out files with header extensions instead? e.g. `.h`/`.hpp`/etc

ZacBlanco (changes requested):
Thanks for this great refactor @yingsu00 ! Decoupling the hive connector from the test infrastructure is a big step forward for the project supporting more data connectors.

I have a few minor comments, but otherwise I am happy with this change.

Below are also a few items pointed out by an AI reviewer that seem like reasonable asks to me

----

Tests that migrate from `HiveConnectorTestBase` to `OperatorTestBase + WithParamInterface<ConnectorTestParam>` implement their own `SetUp()`/`TearDown()` but **do not drain `ioExecutor_`** before calling `unregisterTestConnector`. Compare:

- `HiveConnectorTestBase::TearDown()` (line 58-64): drains `ioExecutor_.reset()` before unregistering — **correct**.
- `ConnectorTestBase::TearDown()` (line 258-263): drains `ioExecutor_.reset()` before unregistering — **correct**.
- `AssertQueryBuilderTest::TearDown()` (in migrated test): does **NOT** drain `ioExecutor_` before `unregisterTestConnector`.

This pattern is repeated in:
- `AssertQueryBuilderTest::TearDown()`
- `GroupedExecutionTest::TearDown()`
- `TaskTest::TearDown()`
- `MergeJoinTest::TearDown()` (via `MergeJoinTestBase`)
- `LocalPartitionTest::TearDown()`

---


### 8.3 `HiveObjectFactory::makeTableHandle` — `dataColumns` Deserialization

**File:** `bolt/connectors/hive/HiveObjectFactory.cpp:216-230`

The old code always built `dataColumns` from `columnHandles`. The new code first checks for `dataColumns` in options and deserializes it:
```cpp
if (const auto* dc = options.get_ptr("dataColumns")) {
  dataColumns = std::dynamic_pointer_cast<const RowType>(
      ISerializable::deserialize<Type>(*dc, nullptr));
}
```
Deserializing with `nullptr` as the memory pool parameter may create allocations in a default/global pool. If `dataColumns` is used in a context where a specific pool is required, this could be an issue. The `remainingFilter` deserialization above it uses `memory::MemoryManager::getInstance()->tracePool()` — inconsistency worth noting.

bolt/connectors/tests/utils/ConnectorTestBase.cpp
yingsu00: > nit: can we make all the default file formats in this Parquet?

Yes I do want to make that change but thought it might be better to do it in a separate PR/commit in the near future, just to isolate the possibilities tests failing because we changed the file format.

yingsu00: > nit: can we make all the default file formats in this Parquet?

Also, the DWIO interface needs similar refactor to the connector one. The exec tests should be file format agnostic too.

ZacBlanco: Thanks, we can move to later if DWRF is the current default.

bolt/connectors/tests/utils/ConnectorTestBase.cpp
yingsu00: > Correct me if I am wrong, but can we actually do this? Don't files need to be split by row groups? or can they be split by an arbitrary number of records?

Yes it was designed this way so that schedulers can split the files easily. The readers in the (Java and Native) workers would round up to RowGroup boundaries and no worker reads partial RowGroups.

ZacBlanco: Thanks for clarifying

PR #585, opened 2026-05-22
[fix] tighten HashBuild probe admission under memory pressure

### LINKED ISSUE #577: [Bug] HashBuild OOM caused by
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

In the current HashJoin flow, HashBuild::finishHashBuild() is the only point where the final merged hash table is built and published to probe. However, there is no effective pre-probe admission check to
  determine whether the built hash table can still be safely admitted into the probe phase under the task's current memory pressure.

  This creates a risky scenario for large build-side tables. HashBuild may barely succeed in constructing the final table, but once the table is handed over to probe, the query still needs additional memory for
  probe processing, output batches, and other concurrent operator activity. If the task is already near or beyond the memory level at which reclaim was previously triggered, admitting the table directly into
  probe can push the query into OOM instead of spilling earlier.

  This timing is especially important because the build-to-probe handoff is the last safe opportunity to spill the join table. After probe starts, HashJoinBridge::tableOrFuture() marks probe as started and the
  table spill callback is no longer available, which means the join table can no longer be spilled through the normal reclaim path. At that point, a later memory reclaim request may no longer be able to free the
  build table memory, making OOM much more likely.

  The root problem is that the existing build-side memory fit check is focused on whether the hash table can be built, but it does not sufficiently account for the extra admission pressure of entering probe. As
  a result, large hash tables may be published even when the task has already shown clear signs of sustained memory pressure.

  A better behavior would be to apply a pre-probe admission check before publishing the final hash table, and to increase admission pressure when the current HashBuild footprint is already large or has exceeded
  the task's historical reclaim watermark. This would make ensureTableFits() more likely to trigger spill before probe begins, reducing the chance of OOM in large-table HashJoin workloads.

### Reproduction Steps

run internal sql

### Bolt Version / Commit ID

main

### System Configuration

```markdown
```

### Logs / Stack Trace

```shell

```

### Expected Behavior

_No response_

### Additional context

_No response_

guhaiyan0221:

Issue Number: close #577 

Reserve additional probe-side admission budget before publishing HashBuild tables when task-level memory pressure is observed, so the build side can still spill before probe starts.

Move memory pressure helpers into MemoryPressureUtils to keep HashBuild away from SparkSQL ExecutionMemoryPool internals, and make ExecutionMemoryPool::debugString() take the pool lock before formatting shared state.

### Type of Change

- [ ] 📝 Documentation only

#### Background

When `HashBuild` finishes building the hash table and is about to publish it to the probe side, the existing admission logic may only reserve memory for the build-side hash table itself. If the executor is already under memory pressure, the probe side may still need additional memory for output batches and join processing after the table is published. At that point, the build side has fewer opportunities to reduce memory usage via spill, which increases the risk of OOM during probe.

##### Reserve additional probe admission budget in HashBuild under memory pressure
Before publishing the built hash table, `HashBuild` now computes an extra reservation based on task-level memory pressure signals. This allows the build side to trigger spill earlier, before the probe side starts consuming the table.
The new logic considers:
    - task memory reclaim watermark
    - SparkSQL `ExecutionMemoryPool` borrow-from-RSS watermark
    - current hash table memory usage
    - whether build-side spill is still available
If a memory pressure watermark exists and current build memory exceeds the threshold, `HashBuild` applies a stronger admission reservation so that it can spill before the probe phase starts.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/common/memory/sparksql/ExecutionMemoryPool.cpp
+++ b/bolt/common/memory/sparksql/ExecutionMemoryPool.cpp
@@ -33,6 +33,11 @@ ExecutionMemoryPoolPtr ExecutionMemoryPool::instance() {
   return instance_;
 }
 
+std::string ExecutionMemoryPool::debugString() {
+  return inited() ? instance()->toString()
+                  : "ExecutionMemoryPool(not initialized)";
+}
+
 // Call ExecutionMemoryPool::init does not mean that BoltMemoryManager is
 // enabled, but enabling BoltMemoryManager definitely means that
 // ExecutionMemoryPool::init has been executed.
@@ -106,6 +111,52 @@ int64_t ExecutionMemoryPool::getMemoryUsageForTask(int64_t taskAttemptId) {
   return (ans == memoryForTask_.end()) ? 0L : ans->second;
 }
 
+ExecutionMemoryPool::ScopedDisableDynamicMemoryQuotaManager
+ExecutionMemoryPool::scopedDisableDynamicMemoryQuotaManagerForTask(
+    int64_t taskAttemptId) {
+  auto self = shared_from_this();
+  disableDynamicMemoryQuotaManagerForTask(taskAttemptId);
+  return folly::makeGuard(
+      std::function<void()>([self = std::move(self), taskAttemptId]() {
+        self->enableDynamicMemoryQuotaManagerForTask(taskAttemptId);
+      }));
+}
+
+std::optional<ExecutionMemoryPool::ScopedDisableDynamicMemoryQuotaManager>
+ExecutionMemoryPool::maybeScopedDisableDynamicMemoryQuotaManagerForTask(
+    const std::optional<int64_t>& taskAttemptId) {
+  if (!taskAttemptId.has_value() || !inited()) {
+    return std::nullopt;
+  }
+  return instance()->scopedDisableDynamicMemoryQuotaManagerForTask(
+      *taskAttemptId);
+}
+
+void ExecutionMemoryPool::disableDynamicMemoryQuotaManagerForTask(
+    int64_t taskAttemptId) {
+  MemoryMutexGuard guard(lock_);
+  ++dynamicMemoryQuotaManagerDisableCounts_[taskAttemptId];
+}
+
+void ExecutionMemoryPool::enableDynamicMemoryQuotaManagerForTask(
+    int64_t taskAttemptId) {
+  MemoryMutexGuard guard(lock_);
+  auto it = dynamicMemoryQuotaManagerDisableCounts_.find(taskAttemptId);
+  if (it == dynamicMemoryQuotaManagerDisableCounts_.end()) {
+    LOG(WARNING)
+        << "TID " << taskAttemptId
+        << " is not disabled by dynamic memory quota manager, cannot be enabled";
+    return;
+  }
+  BOLT_CHECK(
+      it->second > 0,
+      "Expect dynamic memory quota disable count is greater than 0 for task {}",
+      taskAttemptId);
+  if (--it->second == 0) {
+    dynamicMemoryQuotaManagerDisableCounts_.erase(it);
+  }
+}
+
 int64_t ExecutionMemoryPool::acquireMemory(
     int64_t numBytes,
     int64_t taskAttemptId) {
@@ -127,10 +178,12 @@ int64_t ExecutionMemoryPool::acquireMemory(
     int64_t maxPoolSize = internalPoolSize();
     int64_t maxMemoryPerTask = maxPoolSize / numActiveTasks;
     int64_t minMemoryPerTask = internalPoolSize() / (2 * numActiveTasks);
+    int64_t configureMaxMemoryPerTask = poolSize_.value_or(0) / numActiveTasks;
+    int64_t freeMemory = internalMemoryFree();
 
     int64_t maxToGrant =
         std::min(numBytes, std::max<int64_t>(0L, maxMemoryPerTask - curMem));
-    int64_t toGrant = std::min(maxToGrant, internalMemoryFree());
+    int64_t toGrant = std::min(maxToGrant, freeMemory);
 
     if (toGrant < numBytes && curMem + toGrant < minMemoryPerTask) {
       LOG(INFO) << "TID " << taskAttemptId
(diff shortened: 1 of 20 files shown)


bolt/exec/tests/TaskTest.cpp
```diff
@@ -528,6 +528,52 @@ class TaskTest : public HiveConnectorTestBase {
...
+                  .values({makeRowVector(ROW({"c0"}, {BIGINT()}), 1)})
+                  .planFragment();
+  auto task = Task::create(
+      "query_0_TID_123_ATTEMPT_4",
+      std::move(plan),
+      0,
+      core::QueryCtx::create(driverExecutor_.get()),
+      Task::ExecutionMode::kParallel);
+
+  EXPECT_EQ(task->sparkTaskAttemptId(), 123);
```
wangxinshuo-bolt, line 542: `AttemptId` should be `4`?

bolt/common/memory/sparksql/ExecutionMemoryPool.cpp
```diff
@@ -106,6 +111,52 @@ int64_t ExecutionMemoryPool::getMemoryUsageForTask(int64_t taskAttemptId) {
   return (ans == memoryForTask_.end()) ? 0L : ans->second;
 }
 
+ExecutionMemoryPool::ScopedDisableDynamicMemoryQuotaManager
+ExecutionMemoryPool::scopedDisableDynamicMemoryQuotaManagerForTask(
```
wangxinshuo-bolt, line 115: I think the current functionality of `DynamicMemoryQuotaManager` is tightly coupled with `ExecutionMemoryPool`, and it would be better to extract the switch logic for `DynamicMemoryQuotaManager` functionality and the logic of `DynamicMemoryQuotaManager` itself into separate `.cpp/.h` files, which would be more elegant. This can be a long-term optimization goal to prevent code decay.

bolt/exec/HashBuild.cpp
```diff
@@ -1243,6 +1245,52 @@ bool HashBuild::finishHashBuild() {
...
+                  .memoryPressureWatermarkRatio()) {
+    return pressureWatermarkBytes;
+  }
+
+  const uint64_t probeReservationLimit =
+      operatorCtx_->driverCtx()->queryConfig().preferredOutputBatchBytes() *
+      operatorCtx_->driverCtx()
+          ->queryConfig()
+          .outputBatchMemoryReservationMultiple();
+  return std::min<uint64_t>(currentBytes / 2, probeReservationLimit);
```
wangxinshuo-bolt, line 1291: Why divide by `2`?

bolt/exec/HashBuild.cpp
```diff
@@ -1280,11 +1330,29 @@ void HashBuild::ensureTableFits(uint64_t numRows) {
...
   // parallel table build operation.
-  const uint64_t bytesToReserve = table_->estimateHashTableSize(numRows) * 1.1;
+  const uint64_t tableBytesToReserve =
+      table_->estimateHashTableSize(numRows) * 1.1;
+  const uint64_t bytesToReserve = tableBytesToReserve + extraReservationBytes;
+
   {
     Operator::ReclaimableSectionGuard guard(this);
     BOLT_TEST_ADJUST("bytedance::bolt::exec::HashBuild::ensureTableFits", this);
+    const auto scopedDisableMemoryExpansion =
```
wangxinshuo-bolt, line 1340: Is this the most core modification in the entire PR? To remove the impact of DynamicMemoryQuotaManager when maybeReserve.

bolt/exec/MemoryPressureUtils.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "bolt/exec/MemoryPressureUtils.h"
+
+#include "bolt/common/memory/sparksql/ExecutionMemoryPool.h"
+
+#include <exception>
+
+namespace bytedance::bolt::exec::memorypressure {
+
+std::optional<int64_t> extractSparkTaskAttemptId(std::string_view taskId) {
```
wangxinshuo-bolt, line 24: This logic was also present in the previous code. I feel like it could be extracted into a dedicated utility class to handle this

bolt/exec/tests/TaskTest.cpp
guhaiyan0221: 123 is intentional here. This value needs to match the taskAttemptId key used by ExecutionMemoryPool / TaskMemoryManager. In the current task id format query_0_TID_123_ATTEMPT_4, _TID_123 is the Spark
taskAttemptId used as the memory-pool key, while _ATTEMPT_4 is the attempt number. If we change this to 4, borrowFromRssWatermarkBytes(taskAttemptId) would not look up the watermark for the same task.

bolt/common/memory/sparksql/ExecutionMemoryPool.cpp
guhaiyan0221: Agree this is a cleaner long-term direction. In this PR, I kept the scoped disable entry inside ExecutionMemoryPool because the disable counter, borrow-from-RSS watermark, and acquireMemory quota decision are all protected by ExecutionMemoryPool's lock and internal state. Splitting DynamicMemoryQuotaManager here would turn this fix into a larger structural refactor.

bolt/exec/HashBuild.cpp
guhaiyan0221: reserving the full currentBytes would be too aggressive and could cause unnecessary spill or admission failures. So this uses 50% of the current build footprint as a heuristic, and it is still capped by preferredOutputBatchBytes * outputBatchMemoryReservationMultiple.

bolt/exec/HashBuild.cpp
guhaiyan0221: Yes, this is one of the core changes in this PR. ensureTableFits reserves memory right before the final build table is published to the probe side. If maybeReserve triggers DynamicMemoryQuotaManager expansion at this point, it may hide the fact that the task is already under memory pressure, and the probe admission check may not get a strong enough spill/admission signal.

bolt/exec/MemoryPressureUtils.cpp
guhaiyan0221: there is only one implementation of extractSparkTaskAttemptId, which is memorypressure::extractSparkTaskAttemptId, and the previous logic also calls this function.  I kept it as namespace-level utility functions rather than a class because there is no independent state or lifecycle.

wangxinshuo-bolt: In file `bolt/connectors/hive/IgnoreCorruptFile.h` extractAttemptNumber

wangxinshuo-bolt (approved):
LGTM

PR #613, opened 2026-06-01
[fix] Fix oversized BYTE_ARRAY parquet data pages

### LINKED ISSUE #612: [BUG] Parquet writer can produce oversized BYTE_ARRAY data pages for large VARCHAR/VARBINARY batches
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

The Parquet writer can generate oversized data pages for `BYTE_ARRAY` columns, including `VARCHAR` and `VARBINARY`.

For plain `BYTE_ARRAY` encoding, the writer currently appends a whole write batch into the encoder first and only checks `EstimatedDataEncodedSize() >= data_pagesize_` afterward. If a single batch contains enough encoded bytes, the page size limit is checked too late: the writer may already have buffered a data page larger than the configured page size.

In extreme cases, the generated page can exceed the Parquet `PageHeader` `int32` size fields (`compressed_page_size` / `uncompressed_page_size`), causing integer truncation or invalid Parquet output.

This can also happen after dictionary fallback, when a `VARCHAR` / `VARBINARY` column switches from dictionary encoding to plain encoding.

### Reproduction Steps

1. Create a Parquet writer for a `VARCHAR` or `VARBINARY` column.
2. Disable dictionary encoding, or force dictionary fallback with a small dictionary page limit.
3. Set a small data page size for the column, for example:

   ```cpp
   writerOptions.enableDictionary = false;
   writerOptions.columnDataPageSizeMap["c0"] = 100;

### Bolt Version / Commit ID

95bf0f9a9

### System Configuration

```markdown
```

### Logs / Stack Trace

```shell

```

### Expected Behavior

The Parquet writer should split plain `BYTE_ARRAY` writes by encoded byte size before appending them to the encoder, while respecting record boundaries when required by page index or DataPageV2.

The writer should also validate data and dictionary page sizes before assigning them to Parquet `PageHeader` int32 fields.

### Additional context

Affected types:
- VARCHAR
- VARBINARY
- other Parquet BYTE_ARRAY-backed columns

Affected paths:
- plain BYTE_ARRAY encoding
- dictionary fallback to plain BYTE_ARRAY encoding

A robust fix should avoid buffering an oversized page first. For flat Arrow binary arrays, the writer can use Arrow offsets to cheaply estimate encoded byte size before writing the batch, and only fall back to per-value splitting when the batch may exceed the page limit.

Weixin-Xu:
This PR fixes oversized Parquet data pages for `BYTE_ARRAY` columns, including `VARCHAR` and `VARBINARY`.

Previously, plain `BYTE_ARRAY` writes appended a full write batch into the encoder before checking the data page size. If one batch contained enough encoded bytes, the writer could produce a data page much larger than the configured page size. In extreme cases, page sizes could exceed the Parquet `PageHeader` int32 fields.

This PR:
- Splits plain `BYTE_ARRAY` writes by encoded byte size before appending to the encoder.
- Uses Arrow binary offsets as a fast path for flat batches, avoiding per-value scanning unless the batch may exceed the page limit.
- Respects record boundaries when page boundaries must align with records.
- Adds int32 guards for data and dictionary page header sizes.
- Adds writer tests for `VARCHAR`, `VARBINARY`, and dictionary fallback.

Issue Number: close #612

### Type of Change

- [ ] 📝 Documentation only

### Description

This PR fixes oversized BYTE_ARRAY Parquet data pages by splitting plain-encoded BYTE_ARRAY writes before the page size can exceed the Parquet `PageHeader` int32 limit.

Changes include:

- Validate dictionary and data page sizes before writing them into Parquet `PageHeader` int32 fields.
- Split plain-encoded BYTE_ARRAY chunks by encoded byte size for VARCHAR / VARBINARY.
- Preserve dictionary encoding behavior and apply BYTE_ARRAY page splitting after dictionary fallback to plain encoding.
- Use Arrow offsets to cheaply estimate whether the current batch can fit in the current page.
- Fall back to precise per-level splitting only when the batch is close to the data page byte limit.
- Add tests for VARCHAR, VARBINARY, and dictionary fallback cases.
- Fix the Parquet writer benchmark harness to honor folly benchmark iterations and avoid optimized-away results.

### Performance Impact

- [] **No Impact**: This change does not affect the critical path (e.g., build system, doc, error handling).
    <details>

    ```text
    Benchmark target:
    bolt_dwio_parquet_writer_benchmark

    Build:
    cmake --build --preset conan-release --target bolt_dwio_parquet_writer_benchmark

    Baseline:
    implementation before the BYTE_ARRAY page-splitting refactor/optimization

    Current:
    this PR

    Note:
    The benchmark harness was fixed to honor folly benchmark iterations.
    Without this, some fast cases reported 0.00fs / Infinity and were not reliable.

    Targeted BYTE_ARRAY non-dictionary path:
    --bm_regex='VarcharPlain.*no_dict|VarcharListPlain.*no_dict'
    --bm_max_secs=1
    --bm_min_usec=100000

    VarcharPlain 4k null0          15.90ms -> 13.55ms   -14.8%
    VarcharPlain 32k null0        116.61ms -> 95.79ms   -17.8%
    VarcharPlain 256k null0         1.04s  -> 870.75ms  -15.9%

    VarcharPlain 4k null20         18.66ms -> 15.82ms   -15.2%
    VarcharPlain 32k null20       141.76ms -> 120.16ms  -15.2%
    VarcharPlain 256k null20        1.20s  -> 1.03s     -14.1%

    VarcharListPlain 4k null0     114.51ms -> 101.46ms  -11.4%
    VarcharListPlain 32k null0    906.03ms -> 854.19ms  -5.7%
    VarcharListPlain 256k null0     7.82s  -> 7.12s     -9.0%

    VarcharListPlain 4k null20    105.21ms -> 94.35ms   -10.3%
    VarcharListPlain 32k null20   839.07ms -> 821.86ms  -2.1%
    VarcharListPlain 256k null20    6.99s  -> 6.30s     -9.8%

    Additional existing writer benchmark coverage:
    --bm_regex='Varchar_batch|BigInt_batch|Double_batch|ShortDecimalType_batch|LongDecimalType_batch|Map_batch|List_batch'
    --bm_max_secs=1
    --bm_min_usec=100000

    Covered existing benchmark types:
    - Varchar dictionary encoding
    - BigInt
    - Double
    - ShortDecimalType
    - LongDecimalType
    - Map
    - List

    Result:
    No clear performance regression was observed in non-BYTE_ARRAY-targeted paths.
    Some large complex-type cases showed several-percent run-to-run variance, so suspicious cases were rerun individually.

    List-only rerun:
    List_batch_256k_dict    3972.93ms -> 3932.61ms   -1.0%
    List_batch_1M_dict        17.13s  -> 16.69s      -2.5%

    The page-header size check fast path is kept lightweight:
    - normal path is inline
    - overflow branch is guarded by ARROW_PREDICT_FALSE
    - throw helper is marked [[noreturn]]
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
- Fixed oversized BYTE_ARRAY Parquet data pages by splitting plain-encoded VARCHAR / VARBINARY pages before they exceed Parquet PageHeader int32 size limits.
- Optimized the common BYTE_ARRAY write path to avoid per-level splitting checks unless the batch is close to the page size limit.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/arrow/ColumnWriter.cpp
+++ b/bolt/dwio/parquet/arrow/ColumnWriter.cpp
@@ -36,8 +36,11 @@
 #include <algorithm>
 #include <cstdint>
 #include <cstring>
+#include <limits>
 #include <map>
 #include <memory>
+#include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -92,6 +95,17 @@ namespace bytedance::bolt::parquet::arrow {
 using util::CodecOptions;
 
 namespace {
+
+int32_t CheckPageHeaderSize(std::string_view sizeName, int64_t size) {
+  if (size < 0 || size > std::numeric_limits<int32_t>::max()) {
+    throw ParquetException(
+        std::string(sizeName),
+        " page size cannot be represented in a Parquet PageHeader int32 "
+        "field: ",
+        size);
+  }
+  return static_cast<int32_t>(size);
+}
 
 // Visitor that extracts the value buffer from a FlatArray at a given offset.
 struct ValueBufferSlicer {
@@ -350,25 +364,30 @@ class SerializedPageWriter : public PageWriter {
     dict_page_header.__set_is_sorted(page.is_sorted());
 
     const uint8_t* output_data_buffer = compressed_data->data();
-    int32_t output_data_len = static_cast<int32_t>(compressed_data->size());
+    int64_t output_data_len = compressed_data->size();
+    const int32_t uncompressed_page_size =
+        CheckPageHeaderSize("Uncompressed dictionary", uncompressed_size);
+    int32_t compressed_page_size =
+        CheckPageHeaderSize("Compressed dictionary", output_data_len);
 
     if (data_encryptor_.get()) {
       UpdateEncryption(encryption::kDictionaryPage);
       PARQUET_THROW_NOT_OK(encryption_buffer_->Resize(
-          data_encryptor_->CiphertextSizeDelta() + output_data_len, false));
+          data_encryptor_->CiphertextSizeDelta() + compressed_page_size,
+          false));
       output_data_len = data_encryptor_->Encrypt(
           compressed_data->data(),
-          output_data_len,
+          compressed_page_size,
           encryption_buffer_->mutable_data());
       output_data_buffer = encryption_buffer_->data();
+      compressed_page_size =
+          CheckPageHeaderSize("Compressed dictionary", output_data_len);
     }
 
     format::PageHeader page_header;
     page_header.__set_type(format::PageType::DICTIONARY_PAGE);
-    page_header.__set_uncompressed_page_size(
-        static_cast<int32_t>(uncompressed_size));
-    page_header.__set_compressed_page_size(
-        static_cast<int32_t>(output_data_len));
+    page_header.__set_uncompressed_page_size(uncompressed_page_size);
+    page_header.__set_compressed_page_size(compressed_page_size);
     page_header.__set_dictionary_page_header(dict_page_header);
     if (page_checksum_verification_) {
       uint32_t crc32 =
@@ -452,24 +471,29 @@ class SerializedPageWriter : public PageWriter {
     const int64_t uncompressed_size = page.uncompressed_size();
     std::shared_ptr<Buffer> compressed_data = page.buffer();
     const uint8_t* output_data_buffer = compressed_data->data();
-    int32_t output_data_len = static_cast<int32_t>(compressed_data->size());
+    int64_t output_data_len = compressed_data->size();
+    const int32_t uncompressed_page_size =
+        CheckPageHeaderSize("Uncompressed data", uncompressed_size);
+    int32_t compressed_page_size =
+        CheckPageHeaderSize("Compressed data", output_data_len);
 
     if (data_encryptor_.get()) {
       PARQUET_THROW_NOT_OK(encryption_buffer_->Resize(
-          data_encryptor_->CiphertextSizeDelta() + output_data_len, false));
+          data_encryptor_->CiphertextSizeDelta() + compressed_page_size,
+          false));
       UpdateEncryption(encryption::kDataPage);
       output_data_len = data_encryptor_->Encrypt(
           compressed_data->data(),
-          output_data_len,
+          compressed_page_size,
           encryption_buffer_->mutable_data());
       output_data_buffer = encryption_buffer_->data();
+      compressed_page_size =
+          CheckPageHeaderSize("Compressed data", output_data_len);
     }
 
     format::PageHeader page_header;
-    page_header.__set_uncompressed_page_size(
-        static_cast<int32_t>(uncompressed_size));
-    page_header.__set_compressed_page_size(
-        static_cast<int32_t>(output_data_len));
+    page_header.__set_uncompressed_page_size(uncompressed_page_size);
+    page_header.__set_compressed_page_size(compressed_page_size);
 
     if (page_checksum_verification_) {
       uint32_t crc32 =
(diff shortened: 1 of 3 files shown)


bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -2790,6 +2860,76 @@ Status TypedColumnWriterImpl<ByteArrayType>::WriteArrowDense(
     value_offset += batch_num_spaced_values;
   };
 
+  auto WriteChunk = [&](int64_t offset, int64_t batch_size, bool check_page) {
```
luozenglin, line 2872: When `int32_t` overflows, the row group size has already far exceeded the default 128 MB limit, and the modified `WriteChunk` still cannot solve this problem.

In addition, the Parquet writer expects the written batch size to be 40 MB, as referenced here: https://github.com/bytedance/bolt/blob/95bf0f9a99acaf932170d30682b7688ff6226d44/bolt/dwio/parquet/writer/Writer.h#L200

Can we solve this issue by modifying the logic in [`splitWriteRecordBatch`](https://github.com/bytedance/bolt/blob/95bf0f9a99acaf932170d30682b7688ff6226d44/bolt/dwio/parquet/writer/Writer.cpp#L590)? This would avoid both excessively large page sizes and excessively large row group sizes, while making the change more controlled.

Weixin-Xu: The problem fixed here is mainly about the encoded data page size. writeBatchBytes limits the input batch before Parquet encoding/compression, while this issue happens after plain BYTE_ARRAY values are appended into the encoder. Therefore, even with a 40MB write batch limit, a single encoded data page can still become oversized.
The int32 guard here is only a safety check for invalid Parquet PageHeader sizes. The main fix is to split plain BYTE_ARRAY writes before appending too much encoded data into the page encoder. We can improve splitWriteRecordBatch separately for row group/input batch control, but I think the page-level split is still needed.

luozenglin: Could you check whether this has any performance impact and provide benchmark results?

Weixin-Xu: Sure, I can check the performance impact for normal cases.

For the overflow case, benchmark results may not be very meaningful since the original behavior could already write corrupted data when the page size exceeds int32_t. I will mainly verify that regular BYTE_ARRAY writes are not noticeably affected.

luozenglin: Yes, we need to verify the performance when the page size is normal and the `WriteChunk` method does not hit the fast path but enters the `while` loop.

Weixin-Xu: The current change keeps a fast path for the common flat/non-null case when the estimated encoded size is within the page limit.

I will add benchmark results for the normal page-size case where WriteChunk cannot take this fast path and falls back to the while-loop logic, e.g. with def levels / nulls.

Weixin-Xu: I added benchmarks for the normal page-size cases, including plain-encoded VARCHAR and nested ARRAY<VARCHAR>.

The benchmark results did reveal unexpected overhead in the common case, so I moved the fast-path check earlier: we now first use Arrow offsets to conservatively estimate whether the whole batch can fit in the current page. If it can, we write the batch directly and avoid entering the per-level while loop. The precise per-level splitting loop is only used when the batch is close to the page-size limit.

This keeps the oversized-page fix from affecting unrelated/common write paths while still preserving the correctness guard for pages that may exceed the Parquet PageHeader int32 size limit.

luozenglin: I still see some residual performance risk in the cases where the fast path is not applicable, especially nullable VARCHAR/VARBINARY and nested ARRAY<VARCHAR>. In those paths, WriteChunk enters the per-level loop and then WriteSubChunk scans levels again via MaybeCalculateValidityBits / WriteLevelsSpaced, so the no-split normal case may pay an extra full scan.

Current Bolt Parquet writer has no performance advantage over the Java writer for string types. I think checking whether `data_pagesize_` is exceeded at every level is too strict and may introduce performance issues.

We can refer to Arrow’s implementation: [https://github.com/apache/arrow/blob/main/cpp/src/parquet/column_writer.cc](https://github.com/apache/arrow/blob/main/cpp/src/parquet/column_writer.cc). Arrow checks whether `int32_t` overflows, but it does not check the page size level by level.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -2790,6 +2869,76 @@ Status TypedColumnWriterImpl<ByteArrayType>::WriteArrowDense(
...
+    const bool split_by_byte_size =
+        check_page && !IsDictionaryEncoding(current_encoder_->encoding());
+    if (!split_by_byte_size) {
+      WriteSubChunk(offset, batch_size, check_page);
+      return;
+    }
+
+    if (def_levels == nullptr || level_info_.def_level == 0) {
```
luozenglin, line 2880: This fast path does not appear to cover the common Bolt VARCHAR/VARBINARY write path. The Velox-to-Arrow bridge marks fields as nullable, so a normal top-level string leaf usually has def_level > 0 and non-null def_levels, even when the current batch contains no nulls.
As a result, every plain-encoded nullable string batch enters the per-level loop. That loop scans definition levels, validity and value lengths, and then WriteSubChunk() scans the levels again through MaybeCalculateValidityBits() and WriteLevelsSpaced(). It also calls the virtual EstimatedDataEncodedSize() method once per level even though the encoder is not modified inside the loop.
Could we extend the offsets fast path to flat columns without parent nulls, for example using !maybe_parent_nulls && rep_levels == nullptr? Including null slots in the offsets range and 4 * batch_size would be a conservative upper bound, so it should remain safe. Also, please cache EstimatedDataEncodedSize() once per subchunk instead of calling it for every level.

bolt/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -675,6 +675,123 @@ TEST_F(ParquetWriterTest, columnPageSize) {
...
+      if (stats.encoding == thrift::Encoding::PLAIN) {
+        plainDataPageCount += stats.count;
+      }
+    }
+  }
+  ASSERT_EQ(1, dictionaryPageCount);
+  ASSERT_GT(dataPageCount, 1);
+  ASSERT_GT(plainDataPageCount, 1);
+}
+
```
luozenglin, line 794: Could we add a correctness test for ARRAY<VARCHAR> with parent nulls, element nulls and multiple values per row?
The current tests cover flat VARCHAR, VARBINARY and dictionary fallback, but the nested per-level splitting path is the most complex part of this change. A DataPageV2/page-index case would also verify that page splitting continues to respect record boundaries.

bolt/dwio/parquet/tests/writer/ParquetWriterBenchmark.cpp
```diff
@@ -131,6 +136,41 @@ void run(
...
+      run,                                                       \
+      _name_##_batch_256k_no_dict_null##_null_,                  \
+      #_name_,                                                   \
+      _type_,                                                    \
+      _null_,                                                    \
+      262144,                                                    \
+      true);                                                     \
+  BENCHMARK_DRAW_LINE();
+
+#define PARQUET_BENCHMARKS_NO_DICT(_type_, _name_)    \
```
luozenglin, line 170: Thanks for adding nullable VARCHAR and ARRAY<VARCHAR> benchmark coverage. However, the reported numbers combine the new page layout, page count, writer construction and filesystem I/O, so they do not isolate the overhead of the additional per-level scan.
Could we also benchmark a no-split case where both versions produce the same page layout, and report the output file size and page count? Short nullable strings would be the most sensitive case because the added per-value work is less likely to be hidden by value copying.

PR #620, opened 2026-06-02
[fix] avoid int32 overflow in presto serializer offsets

taiyang-li:
## Background

The Presto serializer back-fills the page header / CRC during `flush`. However, several places in the implementation prematurely narrow stream offsets to `int32_t`. When the cumulative output position exceeds `INT32_MAX`, computations such as `offset + size` overflow, eventually causing `seekp` to seek to a wrong (negative) position.

## Changes

- `bolt/serializers/PrestoSerializer.cpp`
  - Keep `offset` / `endOffset` / `endSize` as the stream-native position type instead of narrowing to `int32_t` early.
  - Only narrow to `int32_t` when actually writing the on-wire size field, with explicit overflow checks.
  - In `flushUncompressed`, validate that `uncompressedSize = size - kHeaderSize <= INT32_MAX`.
  - In `flushCompressed`, keep the protocol-level `uncompressedSize <= INT32_MAX` check.
- `bolt/serializers/tests/PrestoSerializerTest.cpp`
  - Add a regression test that exercises a large absolute offset scenario, verifying that flushing across `INT32_MAX` does not produce a negative `seekp`.

## Verification

Added a unit test `PrestoSerializerTest.largeAbsoluteOffsetDoesNotOverflowWhenFlushing` that asserts no negative `seekp` is observed and that `tellp()` ultimately exceeds `INT32_MAX`.

### DIFF
--- a/bolt/serializers/PrestoSerializer.cpp
+++ b/bolt/serializers/PrestoSerializer.cpp
@@ -30,6 +30,8 @@
 
 #include "bolt/serializers/PrestoSerializer.h"
 #include <folly/compression/Zstd.h>
+
+#include <limits>
 #include "bolt/common/base/Crc.h"
 #include "bolt/common/memory/ByteStream.h"
 #include "bolt/common/memory/RawVector.h"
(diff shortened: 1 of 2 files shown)


zhangxffff (approved):
LGTM!

PR #625, opened 2026-06-02
fix(parquet): lazily load pending repdefs during page seek

### LINKED ISSUE #624: [Bug] Parquet reader can fail seeking past known repdefs for repeated columns
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

The native Parquet reader can fail when reading non-top-level repeated columns with chunk-level repetition/definition levels.

The failure happens when the value reader advances to a later physical data page before the corresponding repeated/definition-level page metadata has been materialized into `numLeavesInPage_`.

In this state, `PageReader::setPageRowInfo()` increments `pageIndex_` and checks that the page index is already covered by `numLeavesInPage_`:

  ```cpp
  BOLT_CHECK_LT(
      pageIndex_,
      numLeavesInPage_.size(),
      "Seeking past known repdefs for non top level column page {}",
      pageIndex_);
```

However, this state can be recoverable. More rep/def metadata may already be staged in `preloadedRepDefs_`, but not yet decoded into `numLeavesInPage_`.

As a result, sparse/selective reads over repeated Parquet columns can throw:

Seeking past known repdefs for non top level column page N even though the reader already has pending rep/def batches available and could continue by materializing them.

### Reproduction Steps

A focused unit test can reproduce the failure by constructing the relevant `PageReader` state directly:

  1. Create a non-top-level repeated leaf `PageReader`.
  2. Set chunk-level rep/defs state:
     - `hasChunkRepDefs_ = true`
     - `pageIndex_ = 0`
     - `numLeavesInPage_ = {1}`
  3. Add one pending rep/def batch to `preloadedRepDefs_`.
  4. Call `setPageRowInfo(false)`.

  Without the fix, `setPageRowInfo(false)` increments `pageIndex_` to `1`, sees that `numLeavesInPage_.size()` is still `1`, and fails with:

  (1 vs. 1) Seeking past known repdefs for non top level column page 1

  The regression test could be added for this case is:

  TEST_F(ParquetPageReaderTest, loadsPendingRepDefsBeforePageRowInfoCheck)

  The issue can also be triggered by sparse/selective reads over repeated Parquet columns, especially when rep/def decoding is batched and the value reader reaches a later physical page while more rep/def metadata
  remains pending in preloadedRepDefs_.

  Targeted verification command:

  cmake --build _build/Release --target bolt_dwio_parquet_reader_test

  _build/Release/bolt/dwio/parquet/tests/reader/bolt_dwio_parquet_reader_test \
    --gtest_filter=ParquetPageReaderTest.loadsPendingRepDefsBeforePageRowInfoCheck

  Observed result without the fix:

  [  FAILED  ] ParquetPageReaderTest.loadsPendingRepDefsBeforePageRowInfoCheck

  Reason: (1 vs. 1) Seeking past known repdefs for non top level column page 1
  Expression: pageIndex_ < numLeavesInPage_.size()
  Function: setPageRowInfo
  File: bolt/dwio/parquet/reader/PageReader.cpp

### Bolt Version / Commit ID

main

### System Configuration

```markdown
```

### Logs / Stack Trace

```shell
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (5 vs. 5) Seeking past known repdefs for non top level column page 5
Retriable: False
Expression: pageIndex_ < numLeavesInPage_.size()
Additional Context: Operator: TableScan[0] 0
Function: setPageRowInfo
File: bolt/dwio/parquet/reader/PageReader.cpp
Line: 265
```

### Expected Behavior

  When `PageReader::setPageRowInfo()` advances to a non-top-level repeated-column page whose metadata is not yet present in `numLeavesInPage_`, it should first check whether additional rep/def batches are pending
  in `preloadedRepDefs_`.

  If pending batches exist, the reader should materialize them by calling `loadMoreRepDefs()` before enforcing the bounds check.

  The existing safety check should still remain in place. If no pending rep/def batches exist and `pageIndex_` is still beyond `numLeavesInPage_`, the reader should continue to fail as before because that indicates
  a true invalid seek or corrupted/inconsistent state.

### Additional context

  Root cause hypothesis:

  `numLeavesInPage_` tracks the number of leaf values decoded from rep/def metadata for each data page. With batched rep/def decoding, this vector may lag behind the physical data page reached by the value reader.

  Sparse/selective scans can advance the value path across page boundaries while only a subset of rep/def page metadata has been decoded into `numLeavesInPage_`.

  The important detail is that the missing metadata may already be available in `preloadedRepDefs_`. In that case, failing immediately is too strict. The reader should lazily decode pending rep/def batches until
  either:

  1. `pageIndex_ < numLeavesInPage_.size()`, or
  2. there are no pending rep/def batches left.

  The fix is small and preserves the original invariant:

  ```cpp
  while (pageIndex_ >= static_cast<int32_t>(numLeavesInPage_.size()) &&
         !preloadedRepDefs_.empty()) {
    loadMoreRepDefs();
  }

  BOLT_CHECK_LT(
      pageIndex_,
      numLeavesInPage_.size(),
      "Seeking past known repdefs for non top level column page {}",
      pageIndex_);
```

  Validation:

  • The focused regression test fails without the fix with the expected Seeking past known repdefs error.
  • The same test passes with the fix.
  • The fix only materializes already-preloaded rep/def batches and does not suppress the existing bounds check.

### PR DESCRIPTION
**Pull request by oarap**

Issue Number: close #624 

Repeated Parquet columns keep two streams of state in sync:

  1. the value pages that contain the actual leaf values, and
  2. the repetition/definition levels that describe how those leaf values map back to top-level rows.

  For chunk-level repetition/definition levels, `PageReader` materializes per-page leaf counts in `numLeavesInPage_` and uses `pageIndex_` to look up the count for the current non-top-level data page.

  Sparse selective reads can make the value reader reach a later physical page while `numLeavesInPage_` only contains metadata for pages decoded from earlier rep/def batches. This does not necessarily mean the file
  or seek is invalid: `preloadRepDefs()` may already have staged more rep/def batches in `preloadedRepDefs_`, but those batches have not yet been decoded into `numLeavesInPage_`.

  The old code checked `pageIndex_` against `numLeavesInPage_` immediately and could fail with:

  Seeking past known `repdefs` for non top level column page N

  This PR fixes that by lazily decoding pending rep/def batches before enforcing the existing bounds check.

### Type of Change

- [ ] 📝 Documentation only

### Description

  This PR updates `PageReader::setPageRowInfo()` for non-top-level repeated columns.

  Before this change, PageReader advanced `pageIndex_` and immediately checked whether the current page was already covered by `numLeavesInPage_`.

  That was too eager when `preloadedRepDefs_` was non-empty. In that case, the metadata may already be available but still encoded in a pending rep/def batch.

  The new logic decodes pending rep/def batches until either:

  1. the current `pageIndex_` is covered by `numLeavesInPage_`, or
  2. there are no pending rep/def batches left.

  The original bounds check remains in place, so truly inconsistent state still fails. The reader only recovers when the needed metadata was already preloaded.

  The fix is:
```
  while (pageIndex_ >= static_cast<int32_t>(numLeavesInPage_.size()) &&
         !preloadedRepDefs_.empty()) {
    loadMoreRepDefs();
  }
```

  This PR also adds regression coverage:

  1. A focused PageReader regression test that constructs the exact problematic state:
    • one known page in `numLeavesInPage_`,
    • the next page requested by `pageIndex_`,
    • one pending encoded rep/def batch in `preloadedRepDefs_`.
  Without the lazy load, this test throws the same Seeking past known repdefs error. With the fix, the pending batch is materialized and the page row info is populated.
  2. An end-to-end reader scenario with `ARRAY<VARCHAR>`, small Parquet pages, batched rep/def settings, and a sparse filter to exercise the same reader path through public APIs.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
- Fixed a native Parquet reader failure where sparse reads of repeated columns could seek past currently materialized rep/def page metadata even though more rep/def batches were already preloaded.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/reader/PageReader.cpp
+++ b/bolt/dwio/parquet/reader/PageReader.cpp
@@ -262,6 +262,18 @@ void PageReader::setPageRowInfo(bool forRepDef) {
     numRowsInPage_ = numRepDefsInPage_;
   } else if (hasChunkRepDefs_) {
     ++pageIndex_;
+
+    // The repeated reader normally decodes rep/def metadata far enough before
+    // the value reader seeks into a non-top-level data page. However, sparse
+    // selective reads can make the value path reach the next physical page
+    // exactly when only the already-consumed rep/def page counts are present in
+    // numLeavesInPage_. If more preloaded rep/def batches are available, decode
+    // them here instead of failing the scan immediately.
+    while (pageIndex_ >= static_cast<int32_t>(numLeavesInPage_.size()) &&
+           !preloadedRepDefs_.empty()) {
+      loadMoreRepDefs();
+    }
+
     BOLT_CHECK_LT(
         pageIndex_,
         numLeavesInPage_.size(),
--- a/bolt/dwio/parquet/tests/reader/ParquetReaderTest.cpp
+++ b/bolt/dwio/parquet/tests/reader/ParquetReaderTest.cpp
@@ -1504,6 +1504,66 @@ TEST_F(ParquetReaderTest, skip) {
   EXPECT_EQ(0, rowReader->next(1000, result));
 }
 
+TEST_F(ParquetReaderTest, sparseSkipAcrossRepeatedPagesWithBatchedRepDefs) {
+  constexpr int32_t kNumRows = 6'000;
+  constexpr int32_t kTargetRow = 5'000;
+  const std::string payload(256, 'x');
+
+  auto ids = makeFlatVector<int64_t>(
+      kNumRows, [](auto row) { return static_cast<int64_t>(row); });
+  std::vector<std::string> values;
+  values.reserve(kNumRows);
+  std::vector<std::vector<StringView>> arrays;
+  arrays.reserve(kNumRows);
+  for (auto i = 0; i < kNumRows; ++i) {
+    values.push_back(fmt::format("{}-{}", i, payload));
+    arrays.push_back({StringView(values.back())});
+  }
+
+  auto rowType = ROW({"id", "items"}, {BIGINT(), ARRAY(VARCHAR())});
+  auto data = makeRowVector({ids, makeArrayVector<StringView>(arrays)});
+
+  auto tempFile = exec::test::TempFilePath::create();
+  auto writeFile =
+      std::make_unique<LocalWriteFile>(tempFile->path, true, false);
+  auto sink = std::make_unique<dwio::common::WriteFileSink>(
+      std::move(writeFile), tempFile->path);
+
+  bytedance::bolt::parquet::WriterOptions writerOptions;
+  writerOptions.memoryPool = rootPool_.get();
+  writerOptions.enableDictionary = false;
+  writerOptions.dataPageSize = 1024;
+  writerOptions.columnDataPageSizeMap["items.list.element"] = 1024;
+  writerOptions.writeBatchBytes = 1ULL << 30;
+  writerOptions.minBatchSize = 1;
+  auto writer = std::make_unique<bytedance::bolt::parquet::Writer>(
+      std::move(sink), writerOptions, rowType);
+  writer->write(data);
+  writer->close();
+
+  bytedance::bolt::dwio::common::ReaderOptions readerOptions{leafPool_.get()};
+  readerOptions.setFileSchema(rowType);
+  readerOptions.setFileFormat(dwio::common::FileFormat::PARQUET);
+  auto reader = createReader(tempFile->path, readerOptions);
+
+  auto rowReaderOpts = getReaderOpts(rowType);
+  rowReaderOpts.setDecodeRepDefPageCount(1);
+  rowReaderOpts.setParquetRepDefMemoryLimit(1);
+  auto scanSpec = makeScanSpec(rowType);
+  scanSpec->getOrCreateChild(Subfield("id"))
+      ->setFilter(exec::equal(kTargetRow));
+  rowReaderOpts.setScanSpec(scanSpec);
+  auto rowReader = reader->createRowReader(rowReaderOpts);
+
+  VectorPtr result = BaseVector::create(rowType, 0, leafPool_.get());
+  ASSERT_NO_THROW({
+    auto rowsRead = rowReader->next(1, result);
+    ASSERT_EQ(rowsRead, 1);
+  });
+
+  assertEqualVectorPart(data, result, kTargetRow);
+}
+
 TEST_F(ParquetReaderTest, readVarbinaryFromFLBA) {
   const std::string filename("varbinary_flba.parquet");
   const std::string sample(getExampleFilePath(filename));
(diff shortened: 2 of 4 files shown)


bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -262,6 +262,18 @@ void PageReader::setPageRowInfo(bool forRepDef) {
...
   } else if (hasChunkRepDefs_) {
     ++pageIndex_;
+
+    // The repeated reader normally decodes rep/def metadata far enough before
+    // the value reader seeks into a non-top-level data page. However, sparse
+    // selective reads can make the value path reach the next physical page
+    // exactly when only the already-consumed rep/def page counts are present in
+    // numLeavesInPage_. If more preloaded rep/def batches are available, decode
+    // them here instead of failing the scan immediately.
+    while (pageIndex_ >= static_cast<int32_t>(numLeavesInPage_.size()) &&
```
JinyuanZhang617, line 272: Nice fix — I hit the exact same crash in production and arrived at a similar conclusion.

I considered an alternative placement: doing the loadMoreRepDefs() loop in seekToPage(), 
right before readPageHeader() (i.e., the caller side), rather than inside setPageRowInfo() 
(the callee/crash site). I ended up going with that approach in https://github.com/bytedance/bolt/pull/635

Trade-off I'd like to discuss:

Your approach (callee-side): More defensively robust — protects the CHECK_LT regardless 
  of how setPageRowInfo is reached. If a future code path forgets to preload, this still 
  saves the day.

Caller-side approach: Keeps setPageRowInfo() as a pure metadata lookup with no I/O side-effects. The invariant becomes: "seekToPage is the page lifecycle owner and must guarantee repdef coverage before entering the next page." This preserves the fail-fast semantics of BOLT_CHECK_LT for genuinely broken invariants.

My concern with the callee-side fix is that if a future path forgets to preload and there happen to be 
  leftover batches, the callee-side fix would silently mask the bug.

Not a blocker — both approaches fix the production crash correctly. Just wanted to flag 
this trade-off for the discussion. Happy to hear your thoughts, and feel free to check 
#635 for the alternative implementation.

bolt/dwio/parquet/tests/reader/ParquetReaderTest.cpp
```diff
@@ -1504,6 +1504,66 @@ TEST_F(ParquetReaderTest, skip) {
   EXPECT_EQ(0, rowReader->next(1000, result));
 }
 
+TEST_F(ParquetReaderTest, sparseSkipAcrossRepeatedPagesWithBatchedRepDefs) {
```
JinyuanZhang617, line 1507: FYI — I ran sparseSkipAcrossRepeatedPagesWithBatchedRepDefs on main without this patch 
and it passes. The test doesn't seem to trigger the crash path on an unpatched build.

PR #633, opened 2026-06-05
fix(parquet): drain lazy preloadedRepDefs_ from seekToPage

JinyuanZhang617:
## Background

For nested (non top level) columns Bolt only fully decodes the first decodeRepDefPageCount_ (default 10) pages of each column chunk in preloadRepDefs(). The remaining pages are kept as raw rep/def bytes inside preloadedRepDefs_ and decoded on demand by loadMoreRepDefs(). guhaiyan@ commit 27e789e8 added an ahead-by-one-page invariant at the exit of decodeRepDefs() so that the next consumer access can always read numLeavesInPage_[pageIndex_] safely.

That invariant only fires on the read path driven by RepeatedColumnReader::readRepeatedFor. When a sibling top-level filter pushdown causes the nested column leaf reader to take the skip path instead, control flows SelectiveColumnReader::skip -> PageReader::skip -> seekToPage -> prepareDataPageV1 -> setPageRowInfo and never touches decodeRepDefs. If the skip distance is large enough to cross the sampled boundary, setPageRowInfo does ++pageIndex_ followed by numLeavesInPage_[pageIndex_] and raises (10 vs. 10) Seeking past known repdefs for non top level column page 10.

## Fix

Hoist the lazy-load loop into seekToPage page-walk:

- fired only when hasChunkRepDefs_ AND not isTopLevel_ AND maxRepeat_ greater than 0, so the top-level / no-repdef paths stay zero-cost
- uses pageIndex_ + 1 >= numLeavesInPage_.size() so we always have the per-page leaf count for the page we are about to enter, mirroring guhaiyan@ ahead-by-one-page invariant on the read path
- setPageRowInfo stays a pure lookup, keeping the original layering between rep/def producer (preloadRepDefs / loadMoreRepDefs) and consumer (setPageRowInfo / getLengthsAndNulls)
- loadMoreRepDefs is called at most once per page header, vs. an earlier attempt where it could be called per setPageRowInfo call

## Regression test

lazyRepDefSkipPastSampledBoundary writes a single row group containing a sparse MAP STRING STRING column with dataPageSize=1024 so the chunk has many small pages, applies a sparse IN filter on a sibling BIGINT column and lowers decodeRepDefPageCount to 2 via RowReaderOptions to make the sampling boundary deterministic. Without this change the test reproduces the assertion failure; with the change the read completes successfully.

Change-Id: Iaec5fb28f07533762c1421cba0601096eb203fe4

Issue Number: close #xxx

### Type of Change

- [ ] 📝 Documentation only

### Description

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/reader/PageReader.cpp
+++ b/bolt/dwio/parquet/reader/PageReader.cpp
@@ -62,6 +62,26 @@ void PageReader::seekToPage(int64_t row, bool keepRepDefRawData) {
       numRowsInPage_ = 0;
       break;
     }
+    // For nested (non top level) columns, 'preloadRepDefs' only fully
+    // decodes the first 'decodeRepDefPageCount_' pages and keeps the
+    // remaining pages as raw rep/def bytes inside 'preloadedRepDefs_'.
+    // When a leaf-level filter pushdown triggers a skip()/seekToPage()
+    // that walks past the sampled boundary, the consumer side
+    // ('setPageRowInfo' inside 'prepareDataPageV1') indexes
+    // 'numLeavesInPage_[pageIndex_]' and would go out of bounds.
+    //
+    // Materialise just enough preloaded rep/def bytes BEFORE entering
+    // the next page so that 'setPageRowInfo' remains a pure lookup.
+    // The loop only fires when we are about to cross the sampled
+    // boundary and there is something left to decode, mirroring (and
+    // complementing) the "ahead by one page" invariant established at
+    // the exit of 'decodeRepDefs'.
+    if (hasChunkRepDefs_ && !isTopLevel_ && maxRepeat_ > 0) {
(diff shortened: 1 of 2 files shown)


bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -62,6 +62,27 @@ void PageReader::seekToPage(int64_t row, bool keepRepDefRawData) {
       numRowsInPage_ = 0;
       break;
     }
+    // For nested (non top level) columns, 'preloadRepDefs' only fully
```
guhaiyan0221, line 65: where is the fix? comments only ?

PR #675, opened 2026-06-24
fix(flinksql): return false for invalid regex in rlike

### LINKED ISSUE #657: [Bug] REGEXP / rlike throws BoltUserError on invalid regex pattern instead of returning false (Flink fallback semantics)
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

When the `REGEXP` (a.k.a. `rlike`) scalar function is evaluated with an **invalid regular expression pattern** (for example `**`), Bolt / Bolt raises a `BoltUserError INVALID_ARGUMENT` and the whole pipeline fails:

```
BoltUserError
Error Source: USER
Error Code:   INVALID_ARGUMENT
Reason:       invalid regular expression: bad repetition operator: **
Context:      Top-level Expression: rlike(lark_app_ids, rtc_app_id)
Function:     operator()
File:         bolt/functions/lib/Re2Functions.cpp
Line:         250
```

This breaks compatibility with the **Flink Java SQL runtime**, which:

- Catches `PatternSyntaxException` from `java.util.regex.Pattern.compile(...)`,
- Logs the error,
- And **returns `false`** for that row, allowing the streaming job to keep running.

In production we are running a high-throughput stateless Flink streaming job where the regex pattern comes from **data** (not a constant). Because the patterns are user/data-driven they cannot be statically validated, and a single bad row currently takes the entire job down once Native Engine V2 (Bolt/Bolt) is enabled.

This is a **functional-coverage / behavior-compatibility gap** between the Native (Bolt) engine and the Java Flink engine.

### Reproduction Steps

Minimal repro on the Bolt / Bolt expression engine:

1. Build a row of input where one row’s pattern column contains an invalid regex (e.g. `**`, or any string `java.util.regex.Pattern.compile` would reject):
   ```
   input_str          pattern
   "hello world"      "**"
   ```
2. Evaluate the expression:
   ```sql
   SELECT rlike(input_str, pattern) FROM t;
   ```
   (equivalently, in a Flink SQL job with Native Engine V2 enabled:
   `REGEXP(t3.some_string_col, t1.pattern_col) AS is_match`)
3. Observed: the job/task fails with `BoltUserError INVALID_ARGUMENT: invalid regular expression: bad repetition operator: **`.
4. Expected: the row evaluates to `false` (matching the Java Flink runtime), and the job continues.

Notes:
- The bad pattern only needs to appear on **one** row to fail the entire stream.
- The pattern is **non-constant** (read from another column), so the existing constant-pattern compile-time validation path is not the one being hit at runtime.

### Bolt Version / Commit ID

main @ 59f5090a503f328512ea92fc253f789a04107451

### System Configuration

```markdown
- OS:         Linux (production container)
- Compiler:   (build toolchain used by the Bolt build)
- Build Type: Release
- CPU Arch:   x86_64
- Framework:  Apache Flink (streaming job, Native Engine V2 / Blint enabled, Bolt-based execution)
```

### Logs / Stack Trace

```shell
java.lang.RuntimeException: Exception: BoltRuntimeError
 Error Source: RUNTIME
 Error Code:   INVALID_STATE
 Reason:       Error during calling Java code from native code,
               msg: JNI call processBatch() failed,
               description: java.lang.RuntimeException: Exception: BoltRuntimeError
                 Error Source: RUNTIME
                 Error Code:   INVALID_STATE
                 Reason:       Error during calling Java code from native code,
                               msg: JNI call processBatch() failed,
                               description: java.lang.RuntimeException: Exception: BoltUserError
                                 Error Source: USER
                                 Error Code:   INVALID_ARGUMENT
                                 Reason:       invalid regular expression: bad repetition operator: **
                                 Retriable:    False
                                 Context:      Top-level Expression: rlike(lark_app_ids, rtc_app_id)
                                 Function:     operator()
                                 File:         bolt/functions/lib/Re2Functions.cpp
                                 Line:         250

Native stack (top frames):
 # 0  facebook::bolt::process::StackTrace::StackTrace(int)
 # 1  facebook::bolt::BoltException::BoltException(...)
 # 2  facebook::bolt::detail::boltCheckFail<BoltRuntimeError, std::string const&>(...)
 # 3  blint::jni::OneInputFallBackOperatorCaller::callProcessBatch(ArrowSchema*, ArrowArray*)
 # 4  facebook::bolt::exec::LookupStreamOperator<true>::process(...)
 # 5  facebook::bolt::exec::CalcOperator::process(...)
 # 6  facebook::bolt::exec::LookupStreamOperator<true>::pushOutput(...)
 # 7  Java_com_bytedance_blint_BlintJniWrapper_pushOutputForFallbackOperator__JJJ
 ...

Java stack (top frames):
  at com.bytedance.blint.BlintJniWrapper.pushOutputForFallbackOperator(Native Method)
  at com.bytedance.blint.BlintOperatorBridgeForChain.pushOutputForFallbackOperator(...)
  at org.apache.flink.table.runtime.operators.blint.util.NativeOutput.flushAll(...)
  at org.apache.flink.table.runtime.operators.blint.BlintFallbackOneInputOperator.processBatch(...)
  at com.bytedance.blint.BlintJniWrapper.processShuffleInput(Native Method)
  at org.apache.flink.table.runtime.shuffle.NativeShuffleAccumulator.flush(...)
  at org.apache.flink.table.runtime.operators.blint.AbstractBlintOperator.processElement(...)
  at org.apache.flink.streaming.runtime.tasks.StreamTask.runMailboxLoop(...)
  at org.apache.flink.runtime.taskmanager.Task.run(Task.java:831)
  at java.base/java.lang.Thread.run(Thread.java:840)
```

### Expected Behavior

For non-constant `REGEXP(<input>, <pattern>)`, when `<pattern>` fails to compile as a valid regular expression, the function should **return `false`** for that row (and ideally log/warn once) instead of raising a `BoltUserError` that fails the whole query / streaming task.

Concretely, in `bolt/functions/lib/Re2Functions.cpp` around the non-constant pattern path (~line 250), the `re2::RE2::Options` compile failure should be translated to a `false` result rather than a `BOLT_USER_FAIL`.

If we don’t want to change the default (since other engines may rely on the “strict” behavior), then please consider exposing this as a function-level / session option, e.g. `re2.invalid_pattern_returns_false=true`, so Flink / Spark integrations can opt in to the Java-runtime-compatible semantics.

### Additional context

- Java `java.util.regex.Pattern.compile("**")` also throws `PatternSyntaxException`; Flink Java SQL runtime explicitly **catches that exception, logs it, and returns `false`**. The Native Engine V2 (Bolt/Bolt) currently does not, which is the root cause of the divergence.
- Impact: this is blocking the rollout of Native Engine V2 for a large stateless production streaming job where the regex pattern is data-driven and cannot be sanitized upstream.
- Related code location: `bolt/functions/lib/Re2Functions.cpp` (≈ line 250, non-constant pattern compile path).

xiexiaojun88:
Use InvalidRegexPolicy to make Flink rlike return false instead of throwing on invalid RE2 patterns.

Flink `rlike` currently reuses the shared RE2 search implementation, which throws an exception when the supplied RE2 pattern is invalid. This behavior can fail the whole task, but Flink expects `rlike` to return `false` for invalid regex patterns instead of throwing. This PR introduces an `InvalidRegexPolicy` in the shared RE2 path so different callers can choose how invalid patterns are handled. Flink `rlike` is then registered with a `return false` policy, aligning its behavior with Flink semantics while keeping the default behavior for other RE2-based functions unchanged.

Issue Number: close #657 

### Type of Change

- [ ] 📝 Documentation only

### Description

This change adds policy-based handling for invalid RE2 patterns in the shared regex implementation and applies it to Flink `rlike`.

#### What changed
- Added `InvalidRegexPolicy` to the shared RE2 function layer so callers can choose whether invalid patterns should throw or return `false`.
- Added `makeRe2SearchWithPolicy(...)` and kept the default `re2Search` behavior unchanged.
- Introduced a dedicated Flink `rlike` wrapper that uses `InvalidRegexPolicy::kReturnFalse`.
- Registered Flink `rlike` as a stateful vector function using the new wrapper.
- Added unit tests covering:
  - invalid regex returns `false`
  - valid regex behavior remains unchanged
  - null input still preserves null semantics

#### Why
Flink expects `rlike` to be tolerant of invalid RE2 patterns and return `false` rather than fail the query. The previous shared implementation threw on invalid patterns, which made Flink behavior inconsistent with user expectations.

#### How
The shared RE2 matching path now accepts an invalid-regex policy. For Flink `rlike`, invalid constant and non-constant patterns are handled by returning `false` for non-null rows, while null inputs still return null. Other RE2-based functions continue using the existing throw-on-invalid behavior.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
- Fixed Flink `rlike` so invalid RE2 patterns return false instead of throwing an exception.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/functions/flinksql/RegexFunctions.cpp
+++ b/bolt/functions/flinksql/RegexFunctions.cpp
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) ByteDance Ltd. and/or its affiliates
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "bolt/functions/flinksql/RegexFunctions.h"
+
+#include "bolt/functions/lib/Re2Functions.h"
+
+namespace bytedance::bolt::functions::flinksql {
+
+std::shared_ptr<exec::VectorFunction> makeRLike(
+    const std::string& name,
+    const std::vector<exec::VectorFunctionArg>& inputArgs,
+    const core::QueryConfig& config) {
+  return makeRe2SearchWithPolicy(
+      name, inputArgs, config, InvalidRegexPolicy::kReturnFalse);
+}
+
+std::vector<std::shared_ptr<exec::FunctionSignature>> rlikeSignatures() {
+  return re2SearchSignatures();
+}
+
+} // namespace bytedance::bolt::functions::flinksql
--- a/bolt/functions/flinksql/CMakeLists.txt
+++ b/bolt/functions/flinksql/CMakeLists.txt
@@ -17,6 +17,7 @@ bolt_add_library(
   bolt_functions_flink_impl
   HashCodeFunction.cpp
   JsonFunctions.cpp
+  RegexFunctions.cpp
   ElementAt.cpp
   specialforms/FlinkCastExpr.cpp
 )
--- a/bolt/functions/flinksql/RegexFunctions.h
+++ b/bolt/functions/flinksql/RegexFunctions.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) ByteDance Ltd. and/or its affiliates
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <memory>
+#include <vector>
+
+#include "bolt/expression/Expr.h"
+#include "bolt/vector/BaseVector.h"
+
+namespace bytedance::bolt::functions::flinksql {
+
+// Flink rlike reuses the shared RE2 vector-function implementation in
+// lib/Re2Functions.h, but chooses an invalid-regex policy that returns false
+// instead of failing the task.
+std::shared_ptr<exec::VectorFunction> makeRLike(
+    const std::string& name,
+    const std::vector<exec::VectorFunctionArg>& inputArgs,
+    const core::QueryConfig& config);
+
+std::vector<std::shared_ptr<exec::FunctionSignature>> rlikeSignatures();
+
+} // namespace bytedance::bolt::functions::flinksql
--- a/bolt/functions/flinksql/registration/Register.cpp
+++ b/bolt/functions/flinksql/registration/Register.cpp
@@ -22,6 +22,7 @@
 #include "bolt/functions/flinksql/DateTimeFunctions.h"
 #include "bolt/functions/flinksql/HashCodeFunction.h"
 #include "bolt/functions/flinksql/Rand.h"
+#include "bolt/functions/flinksql/RegexFunctions.h"
 #include "bolt/functions/flinksql/String.h"
 #include "bolt/functions/flinksql/ToTimestampFunction.h"
 #include "bolt/functions/flinksql/specialforms/FlinkCastExpr.h"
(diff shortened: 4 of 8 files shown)


bolt/functions/flinksql/RegexFunctions.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+std::shared_ptr<exec::VectorFunction> makeRLike(
+    const std::string& name,
+    const std::vector<exec::VectorFunctionArg>& inputArgs,
+    const core::QueryConfig& config) {
+  return makeRe2SearchWithPolicy(
+      name, inputArgs, config, InvalidRegexPolicy::kReturnFalse);
+}
+
+std::vector<std::shared_ptr<exec::FunctionSignature>> rLikeSignatures() {
```
yangzhg:
```suggestion
std::vector<std::shared_ptr<exec::FunctionSignature>> rlikeSignatures() {
```

yangzhg (approved):
LGTM

PR #740, opened 2026-07-15
Add parquet page read and filtered-out metrics

lizhen-0710:
What problem does this PR solve?
  

  Issue Number: close #xxx

  Type of Change
  

  - [ ] 📝 Documentation only

  Description

  This PR adds Parquet page-level runtime metrics for TableScan so we can observe how many data pages were actually
  considered by the reader, and how many of those pages produced no output after subfield filter evaluation.

  Specifically:

  - Adds totalPages and filteredOutPages to Parquet runtime statistics.
  - Counts pages at the PageReader level for pages that are actually read/considered by kept row groups.
  - Tracks pages that are read/considered but end up producing zero output rows under subfield filters.
  - Does not count pages from row groups eliminated earlier by row-group filtering.
  - Updates ParquetTableScanTest to validate the metric semantics using column_index.parquet.
  - Fixes the test helper so filters on unprojected columns are parsed against the full table schema instead of only the
    output schema.

  Why:

  - Previous visibility into page skipping was limited.
  - We need concrete runtime counters to understand how much work page-level pruning is avoiding.
  - The final metric semantics should reflect actual reader behavior, not an estimated or theoretical page count.

  How:

  - totalPages is incremented when a Parquet data page is actually considered by the page reader for a participating
    column.

  - filteredOutPages is incremented when such a page contributes no output rows.
  - Tests were corrected to match the real page layout in column_index.parquet and current execution semantics:
      - single-column no-filter reads all data pages in kept row groups,
      - row-group-pruned pages are not counted,
      - non-filter projection columns only contribute surviving pages rather than mechanically multiplying total page
        count by projected column count.

  Performance Impact
  

    N/A
      </details>

  Release Note
  

  Release Note:

  Release Note:
  - Added Parquet TableScan runtime metrics for data pages read and pages filtered out by subfield filters.
  - Added/updated tests to validate page metric semantics with column index based fixtures.

  Checklist (For Author)
  

  Breaking Changes
  

  - [x] No
  - [ ] Yes (Description: ...)

    Breaking Changes:
    - None.
      </details>


bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -112,9 +118,20 @@ void PageReader::seekToPage(int64_t row, bool keepRepDefRawData) {
...
     }
+    if (row < rowOfPage_ + numRowsInPage_) {
+      if (FOLLY_LIKELY(statis_ != nullptr) && numRowsInPage_ > 0) {
+        ++statis_->totalPages;
+      }
+      currentPageNumValues_ = 0;
+      break;
+    }
+    if (FOLLY_LIKELY(statis_ != nullptr) && numRowsInPage_ > 0) {
+      ++statis_->totalPages;
```
guhaiyan0221, line 132: I think counting this branch as `filteredOutPages` overestimates the potential benefit of future page-index pruning. Consider a query like `SELECT c2 FROM t WHERE c1 < 10`. After filtering on `c1`, the
  surviving row set can be very sparse. When reading projected column `c2`, `seekToPage()` may skip many `c2` pages simply because none of the surviving `c1` rows fall in those page row ranges.

  Those skipped `c2` pages are unrelated to `c2`'s own data distribution, and they do not imply that `c2`'s page index could prune them. The skip is caused by high selectivity on `c1`, not by a filter on `c2` or
  by `c2` page min/max. If this metric is meant to estimate future page-index pruning potential, this branch should not increment `filteredOutPages`; it should probably only increment `totalPages`, or use a
  separate `seekSkippedPages` metric.

This is not a blocker from my side, since the metric is still useful as an approximate signal for pages that do not contribute output

guhaiyan0221 (approved):

PR #747, opened 2026-07-16
fix(parquet): prevent batch-induced oversized BYTE_ARRAY pages

### LINKED ISSUE #612: [BUG] Parquet writer can produce oversized BYTE_ARRAY data pages for large VARCHAR/VARBINARY batches

olemon111:

The Parquet writer could produce oversized data pages for `BYTE_ARRAY` columns, including `VARCHAR` and `VARBINARY`.

Previously, a complete write batch was appended to the encoder before checking the configured data page size. A batch containing large variable-width values could therefore make the buffered page significantly exceed the page-size target. In extreme cases, the compressed or uncompressed page size could exceed the `int32` fields in the Parquet `PageHeader`, producing invalid Parquet files and causing excessive reader-side allocations.

The same problem could occur after dictionary encoding falls back to PLAIN encoding.

Issue Number: close #612

### Type of Change

- [ ] 📝 Documentation only

### Description

This change prevents batch-induced oversized Parquet `BYTE_ARRAY` pages before values are appended to the page encoder.

The writer now:

- Estimates the PLAIN-encoded size of a `BYTE_ARRAY` chunk using Arrow binary offsets before calling the encoder.
- Writes a normal chunk directly when it fits in the current page.
- Flushes the current page and writes the whole chunk when it fits in an empty page.
- Performs per-level splitting only when the chunk itself cannot fit in an empty page.
- Preserves record boundaries for nested columns when required by DataPageV2 or page-index writing.
- Allows a single oversized value or record to make progress, while relying on the final `PageHeader` size validation for the hard Parquet format limit.

The no-split fast path applies to flat, nullable, and nested `BYTE_ARRAY` columns. It reuses the validity metadata required by the existing write path and avoids an additional per-level value-length scan for normal chunks.

Dictionary encoding is also covered:

- Dictionary encoded size accounting now uses `int64_t`.
- Dictionary growth is checked using a byte budget before appending a potentially large batch.
- Repeated dictionary values do not unnecessarily trigger fallback.
- After dictionary fallback, PLAIN-encoded values use the same data-page splitting protection.

As a final safety boundary, compressed, uncompressed, and encrypted data/dictionary page sizes are validated before narrowing them to the Parquet `PageHeader` `int32` fields.

Tests cover:

- Flat PLAIN-encoded `VARCHAR`.
- Nullable PLAIN-encoded `VARBINARY`.
- Nested `ARRAY<VARCHAR>` with parent and element nulls.
- DataPageV2 record-boundary splitting.
- A final oversized nested record.
- Dictionary page byte budgeting.
- Dictionary fallback followed by PLAIN page splitting.
- Repeated dictionary values.
- A single oversized `BYTE_ARRAY` value.
- Data and dictionary `PageHeader` size overflow guards.

The Parquet writer benchmark was also updated to:

- Honor Folly benchmark iterations.
- Exclude writer construction, data generation, metadata inspection, and destruction from the measured interval.
- Use an in-memory sink for targeted no-split cases.
- Verify that baseline and fixed implementations produce one data page.
- Report output size and data page count.
- Cover short flat, nullable, VARBINARY, nested `ARRAY<VARCHAR>`, and normal
  dictionary-encoded inputs.

### Performance Impact

- [x] **Negative Impact**: Targeted benchmarks show up to 3.22% overhead for
  short VARCHAR batches with no actual null values; other tested no-split
  cases show no clear regression.

  <details>

  ```text
  Benchmark target:
  bolt_dwio_parquet_writer_benchmark

  Note:
  Both versions used the same benchmark harness. Writer construction, input
  generation, metadata inspection, and filesystem I/O are excluded. Results
  are 5-run medians, and both versions produced the same output size and one
  data page for every case.

  Targeted BYTE_ARRAY no-split path:
  --bm_regex='NoSplit'
  --bm_max_secs=1
  --bm_min_usec=100000

  VarcharDictionary 4k null20       607.84us -> 603.63us  -0.69%
  VarcharDictionary 32k null20        4.14ms -> 4.12ms    -0.53%

  VarbinaryPlain 4k null20             395us -> 394.69us  -0.08%
  VarbinaryPlain 32k null20           2.45ms -> 2.45ms    +0.11%
  VarbinaryPlain 256k null20         20.41ms -> 20.07ms   -1.62%

  VarcharListPlain 4k null20          1.96ms -> 1.96ms    +0.23%
  VarcharListPlain 32k null20        15.64ms -> 15.30ms   -2.16%

  VarcharPlain 4k null0             332.12us -> 335.63us  +1.06%
  VarcharPlain 32k null0               2.00ms -> 2.07ms   +3.22%
  VarcharPlain 256k null0             15.38ms -> 15.57ms  +1.26%

  VarcharPlain 4k null20            379.97us -> 380.63us  +0.17%
  VarcharPlain 32k null20              2.46ms -> 2.46ms   +0.10%
  VarcharPlain 256k null20            20.63ms -> 20.49ms  -0.68%

  Result:
  No broad performance regression was observed; the largest measured overhead
  was 3.22%.
  ```

  </details>

### Release Note

Release Note:

```text
Release Note:
- Fixed oversized Parquet BYTE_ARRAY data pages for VARCHAR and VARBINARY columns, including dictionary fallback to PLAIN encoding.
- Added Parquet PageHeader size validation to prevent invalid files caused by int32 page-size overflow.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/arrow/ColumnWriter.cpp
+++ b/bolt/dwio/parquet/arrow/ColumnWriter.cpp
@@ -36,8 +36,11 @@
 #include <algorithm>
 #include <cstdint>
 #include <cstring>
+#include <limits>
 #include <map>
 #include <memory>
+#include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -95,6 +98,19 @@ namespace bytedance::bolt::parquet::arrow {
 using util::CodecOptions;
 
 namespace {
+
+constexpr int64_t kMaxPageHeaderSize = std::numeric_limits<int32_t>::max();
+
+int32_t checkPageHeaderSize(std::string_view size_name, int64_t size) {
+  if (size < 0 || size > kMaxPageHeaderSize) {
+    throw ParquetException(
+        std::string(size_name),
+        " page size cannot be represented in a Parquet PageHeader int32 "
+        "field: ",
+        size);
+  }
+  return static_cast<int32_t>(size);
+}
 
 int64_t bufferAllocatedBytes(const std::shared_ptr<Buffer>& buffer) {
   return buffer != nullptr ? buffer->capacity() : 0;
@@ -410,7 +426,9 @@ class SerializedPageWriter : public PageWriter {
   int64_t WriteDictionaryPage(
       std::unique_ptr<DictionaryPage> pagePtr) override {
     const auto& page = *pagePtr;
-    int64_t uncompressed_size = page.size();
+    const int64_t uncompressed_size = page.buffer()->size();
+    const int32_t uncompressed_page_size =
+        checkPageHeaderSize("Uncompressed dictionary", uncompressed_size);
     std::shared_ptr<Buffer> compressed_data;
     if (has_compressor()) {
       auto buffer = std::static_pointer_cast<ResizableBuffer>(
@@ -427,25 +445,28 @@ class SerializedPageWriter : public PageWriter {
     dict_page_header.__set_is_sorted(page.is_sorted());
 
     const uint8_t* output_data_buffer = compressed_data->data();
-    int32_t output_data_len = static_cast<int32_t>(compressed_data->size());
+    int64_t output_data_len = compressed_data->size();
+    int32_t compressed_page_size =
+        checkPageHeaderSize("Compressed dictionary", output_data_len);
 
     if (data_encryptor_.get()) {
       UpdateEncryption(encryption::kDictionaryPage);
       PARQUET_THROW_NOT_OK(encryption_buffer_->Resize(
-          data_encryptor_->CiphertextSizeDelta() + output_data_len, false));
+          data_encryptor_->CiphertextSizeDelta() + compressed_page_size,
+          false));
       output_data_len = data_encryptor_->Encrypt(
           compressed_data->data(),
-          output_data_len,
+          compressed_page_size,
           encryption_buffer_->mutable_data());
       output_data_buffer = encryption_buffer_->data();
+      compressed_page_size =
+          checkPageHeaderSize("Encrypted dictionary", output_data_len);
     }
 
     format::PageHeader page_header;
     page_header.__set_type(format::PageType::DICTIONARY_PAGE);
-    page_header.__set_uncompressed_page_size(
-        static_cast<int32_t>(uncompressed_size));
-    page_header.__set_compressed_page_size(
-        static_cast<int32_t>(output_data_len));
+    page_header.__set_uncompressed_page_size(uncompressed_page_size);
+    page_header.__set_compressed_page_size(compressed_page_size);
     page_header.__set_dictionary_page_header(dict_page_header);
     if (page_checksum_verification_) {
       uint32_t crc32 =
@@ -531,26 +552,31 @@ class SerializedPageWriter : public PageWriter {
   int64_t WriteDataPage(std::unique_ptr<DataPage> pagePtr) override {
     const DataPage& page = *pagePtr;
     const int64_t uncompressed_size = page.uncompressed_size();
+    const int32_t uncompressed_page_size =
+        checkPageHeaderSize("Uncompressed data", uncompressed_size);
     std::shared_ptr<Buffer> compressed_data = page.buffer();
     const uint8_t* output_data_buffer = compressed_data->data();
-    int32_t output_data_len = static_cast<int32_t>(compressed_data->size());
+    int64_t output_data_len = compressed_data->size();
+    int32_t compressed_page_size =
+        checkPageHeaderSize("Compressed data", output_data_len);
 
     if (data_encryptor_.get()) {
       PARQUET_THROW_NOT_OK(encryption_buffer_->Resize(
-          data_encryptor_->CiphertextSizeDelta() + output_data_len, false));
+          data_encryptor_->CiphertextSizeDelta() + compressed_page_size,
+          false));
       UpdateEncryption(encryption::kDataPage);
       output_data_len = data_encryptor_->Encrypt(
           compressed_data->data(),
-          output_data_len,
+          compressed_page_size,
           encryption_buffer_->mutable_data());
       output_data_buffer = encryption_buffer_->data();
+      compressed_page_size =
+          checkPageHeaderSize("Encrypted data", output_data_len);
     }
 
     format::PageHeader page_header;
-    page_header.__set_uncompressed_page_size(
-        static_cast<int32_t>(uncompressed_size));
-    page_header.__set_compressed_page_size(
-        static_cast<int32_t>(output_data_len));
+    page_header.__set_uncompressed_page_size(uncompressed_page_size);
+    page_header.__set_compressed_page_size(compressed_page_size);
 
     if (page_checksum_verification_) {
       uint32_t crc32 =
@@ -2216,8 +2242,10 @@ class TypedColumnWriterImpl : public ColumnWriterImpl,
       return;
     }
 
+    const int64_t dictionary_page_byte_limit =
+        std::clamp<int64_t>(dictionary_pagesize_limit_, 0, kMaxPageHeaderSize);
     if (current_dict_encoder_->dict_encoded_size() >=
-        dictionary_pagesize_limit_) {
+        dictionary_page_byte_limit) {
       FallbackToPlainEncoding();
     }
   }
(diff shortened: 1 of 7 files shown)


bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -531,26 +552,31 @@ class SerializedPageWriter : public PageWriter {
...
     const uint8_t* output_data_buffer = compressed_data->data();
-    int32_t output_data_len = static_cast<int32_t>(compressed_data->size());
+    int64_t output_data_len = compressed_data->size();
+    int32_t compressed_page_size =
+        checkPageHeaderSize("Compressed data", output_data_len);
 
     if (data_encryptor_.get()) {
       PARQUET_THROW_NOT_OK(encryption_buffer_->Resize(
-          data_encryptor_->CiphertextSizeDelta() + output_data_len, false));
+          data_encryptor_->CiphertextSizeDelta() + compressed_page_size,
```
Weixin-Xu: When the compressed page is close to `INT32_MAX`, this addition can overflow before the post-encryption `checkPageHeaderSize()` is reached. The same issue exists in the dictionary-page path.

bolt/dwio/parquet/arrow/tests/ColumnWriterTest.cpp
```diff
@@ -993,6 +995,83 @@ TEST(TestColumnWriter, RepeatedListsUpdateSpacedBug) {
...
+
+  std::shared_ptr<GroupNode> node_;
+  SchemaDescriptor schemaDescriptor_;
+  std::shared_ptr<WriterProperties> properties_;
+  std::unique_ptr<ColumnChunkMetaDataBuilder> metadata_;
+  std::shared_ptr<::arrow::io::BufferOutputStream> sink_;
+  std::unique_ptr<PageWriter> pageWriter_;
+};
+
+TEST_F(PageWriterSizeGuardTest, rejectsUncompressedDataPageSizeOverflow) {
```
Weixin-Xu, line 1063: These tests verify the final PageWriter guard, but they do not exercise the BYTE_ARRAY splitter against the PageHeader hard limit.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -2976,18 +3004,109 @@ Status TypedColumnWriterImpl<ByteArrayType>::WriteArrowDense(
...
-  int64_t value_offset = 0;
-  auto WriteChunk = [&](int64_t offset, int64_t batch_size, bool check_page) {
-    int64_t batch_num_values = 0;
-    int64_t batch_num_spaced_values = 0;
-    int64_t null_count = 0;
+  // The configured page sizes are soft flush targets. PageHeader fields are
+  // the hard format boundary, so cap byte-budget arithmetic at INT32_MAX and
+  // leave the final encoded/compressed size checks to PageWriter.
+  auto PageByteLimit = [](int64_t configured_limit) {
+    return std::clamp<int64_t>(configured_limit, 0, kMaxPageHeaderSize);
```
luozenglin: Is it more reasonable to throw an exception when `configured_limit` is not within the range?

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -3012,6 +3131,167 @@ Status TypedColumnWriterImpl<ByteArrayType>::WriteArrowDense(
...
+    int64_t null_count = 0;
+    MaybeCalculateValidityBits(
+        AddIfNotNull(def_levels, offset),
+        batch_size,
+        &batch_num_values,
+        &batch_num_spaced_values,
+        &null_count);
+
+    // For PLAIN encoding this is the data-page value size. During dictionary
+    // encoding it is a conservative upper bound on dictionary growth: duplicate
```
luozenglin: When cardinality is low, using the plain-encoded size to estimate the deduplicated dictionary size can introduce an order-of-magnitude error. Is this reasonable?

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -3012,6 +3131,167 @@ Status TypedColumnWriterImpl<ByteArrayType>::WriteArrowDense(
...
+    // values consume no additional dictionary bytes, but treating them as new
+    // entries keeps a single write batch from causing a large peak allocation.
+    // The offset range is exact for canonical Arrow binary arrays and a
+    // conservative upper bound if a null slot retains bytes.
+    const int64_t page_byte_limit = CurrentPageByteLimit();
+    const int64_t batch_encoded_bytes = PlainEncodedRangeSize(
```
luozenglin: When using `DELTA_SYTE-ARRAY` encoding, `batch_encoded_bytes` uses PLAIN size while `CurrentPageBytes()` uses actual DELTA size. This can greatly overestimate repeated or prefix-similar strings and create excessive pages。

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -3012,6 +3131,167 @@ Status TypedColumnWriterImpl<ByteArrayType>::WriteArrowDense(
...
+        value_offset,
+        batch_num_spaced_values,
+        batch_num_values,
+        page_byte_limit);
+    if (CappedPageBytes(
```
luozenglin: I guess there are many cases online where `batch_decoded_bytes` are greater than 1M but less than INT32 MAX, which cannot hit this fast path. Please sample the online situation for performance testing and consider whether it is necessary to increase the default 1M page size

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: data_pagesize is an existing best-effort flush target rather than a hard user-visible limit. I kept the configured value accepted and only cap the internal BYTE_ARRAY budget at the PageHeader int32 boundary, adding a throw here would introduce behavior change.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: Thanks, fixed both data and dictionary paths by validating the encrypted buffer size in int64 before allocation and encryption. Regression tests were added for both paths.

bolt/dwio/parquet/arrow/tests/ColumnWriterTest.cpp
olemon111: Added a ColumnWriter test with a reduced PageHeader limit. It writes real Arrow VARCHAR batches with the configured page size above the hard limit and verifies both PLAIN and dictionary/fallback paths keep emitted pages within that limit. Production still uses INT32_MAX.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: The estimate is intentionally conservative and follows the same policy as [arrow-rs #9972](https://github.com/apache/arrow-rs/pull/9972). Its merged implementation uses the remaining dictionary-page budget and estimates ordinary Utf8/Binary input from its PLAIN-encoded size. Arrow DictionaryArray input is handled separately because its values are already deduplicated ([code and comment](https://github.com/apache/arrow-rs/blob/1e8ea5eb6857bad5247908c1c27aa51e62aa9006/parquet/src/arrow/arrow_writer/byte_array.rs#L547-L552)).
For low-cardinality input this may cause additional subchunk work, but fallback still checks the actual dict_encoded_size(), so the conservative estimate itself does not cause false fallback. `repeatedByteArrayValuesDoNotForceDictionaryFallback` covers this case.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: Thanks. I separated the PLAIN/raw byte budget from the actual encoded page size. DELTA now uses the raw estimate only to bound each encoder Put(), while page flushing is based on EstimatedDataEncodedSize(). PLAIN and dictionary keep their remaining-budget behavior. I also fixed DELTA_LENGTH_BYTE_ARRAY Arrow payload accounting and added compressible/incompressible DELTA page-layout and round-trip tests.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: Thanks. I validated the fix with the affected online cases, but stable reproducible samples are limited because some failures are intermittent. I’ll continue monitoring and collecting samples.
I prefer to keep the default 1 MiB page target unchanged in this PR. Increasing it would raise per-page memory usage and only move the threshold rather than solve oversized batches. Any default-size tuning should be evaluated separately with representative workload data.

luozenglin (approved):
LGTM

PR #784, opened 2026-07-24
Optimize Parquet lazy repdef window decode

Gin-Fung:
## Summary

This PR optimizes Parquet repetition/definition level decoding by adding a lazy windowed decode path.

The change is split into three logical commits:

1. Add option plumbing for Parquet repdef preload window mode.
2. Add the preload window implementation.
3. Optimize the hot path by avoiding extra leaf-count decode work and repeated top-level row boundary scans.

## Validation

Validated locally with:

- bolt_dwio_parquet_reader_test
- bolt_dwio_parquet_table_scan_test
- bolt_dwio_parquet_cli_test

Both default mode and window preload mode passed.

### DIFF
--- a/bolt/dwio/parquet/reader/PageReader.cpp
+++ b/bolt/dwio/parquet/reader/PageReader.cpp
@@ -103,10 +103,11 @@ void PageReader::seekToPage(int64_t row, bool keepRepDefRawData) {
       numRowsInPage_ = 0;
       break;
     }
-    if (row != kRepDefOnly && windowRepDefMode_ && hasChunkRepDefs_ &&
+    if (row != kRepDefOnly && windowRepDef_.enabled && hasChunkRepDefs_ &&
         !isTopLevel_ && maxRepeat_ > 0) {
-      while (pageIndex_ + 1 >= static_cast<int32_t>(numLeavesInPage_.size()) &&
-             (!preloadedRepDefs_.empty() || windowRepDefsRemaining_ > 0)) {
+      while (
+          pageIndex_ + 1 >= static_cast<int32_t>(numLeavesInPage_.size()) &&
+          (!preloadedRepDefs_.empty() || windowRepDef_.repDefsRemaining > 0)) {
         ensureNumLeavesInPage(pageIndex_ + 1);
       }
     } else if (
@@ -852,13 +853,7 @@ void PageReader::preloadPageRepDefs(const bool keepRepDefRawData) {
 void PageReader::preloadRepDefs() {
   const auto startNs = getCurrentTimeNano();
   hasChunkRepDefs_ = true;
-  windowRepDefMode_ = false;
-  windowTopLevelOffsets_.clear();
-  windowRepDefPageData_.clear();
-  windowRepDefsRemaining_ = 0;
-  windowCurrentPageLeafCount_ = 0;
-  windowDecodedPageCount_ = 0;
-  windowTopLevelOffsetBase_ = 0;
+  windowRepDef_.reset();
   bool startWithDictinoaryPage = cryptoCtx_.startDecryptWithDictionaryPage;
   if (maxRepeat_ > 0 && maxDefine_ > 0) {
     const int32_t samplePages = decodeRepDefPageCount_;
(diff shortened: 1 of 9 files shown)


bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -871,52 +918,523 @@ void PageReader::preloadRepDefs() {
...
+  addPageReaderStat("pageReaderLoadWindowRepDefsCount", 1);
+  addPageReaderStat("pageReaderLoadWindowRepDefsBatches", decodeBatchCount);
+  addPageReaderNanos(
+      "pageReaderLoadWindowRepDefsTimeNs", getCurrentTimeNano() - startNs);
 }
 
 void PageReader::decodeRepDefs(int32_t numTopLevelRows) {
-  if (definitionLevels_.empty() && maxDefine_ > 0) {
-    preloadRepDefs();
+  const auto startNs = getCurrentTimeNano();
```
liuneng1994, line 1317: This changes the core rep/def preload/decode path, but I don't see any tests added in this PR. Can we add coverage for window mode vs the existing path on nested repeated/optional data? The cases I would expect at minimum are: window mode disabled, window mode enabled across multiple pages, and skip/seek after only part of the window has been decoded.

bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -871,52 +918,523 @@ void PageReader::preloadRepDefs() {
...
+  BOLT_CHECK_LE(rawData + repeatLength, pageEnd);
+  rawData += repeatLength;
+
+  const auto defineLength = readField<uint32_t>(rawData);
+  BOLT_CHECK_LE(rawData + defineLength, pageEnd);
+  if (leafInfo_.rep_level == 0) {
+    return numRepDefsInPage;
+  }
+
+  auto bitReader = ::arrow::bit_util::BitReader(
```
liuneng1994, line 983: This looks like a second implementation of Parquet def-level RLE decoding. Can we reuse the existing level decoder path, or extract a shared helper, instead of hand-parsing the RLE runs here?
  This is a correctness-sensitive format detail and easy to drift from the normal path.

bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -871,52 +918,523 @@ void PageReader::preloadRepDefs() {
...
+        numLeaves += repeatCount;
+      }
+      remaining -= repeatCount;
+    }
+  }
+  return numLeaves;
+}
+
+void PageReader::compactConsumedLeafNulls() {
+  constexpr int32_t WordBits = 64;
```
liuneng1994, line 1030: This duplicates the leaf-null compaction logic, including the glibc/FSRM memcpy workaround from the existing path. Can we extract that existing block into a shared helper and call it from both
  paths?

bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -807,11 +839,26 @@ void PageReader::preloadPageRepDefs(const bool keepRepDefRawData) {
...
+  }
+  addPageReaderNanos(
+      "pageReaderPreloadPageRepDefsTimeNs", getCurrentTimeNano() - startNs);
   return;
 }
 
 void PageReader::preloadRepDefs() {
+  const auto startNs = getCurrentTimeNano();
   hasChunkRepDefs_ = true;
+  windowRepDefMode_ = false;
```
liuneng1994: These window-mode fields are tightly coupled. Can we group them into a small private state struct, maybe with a `reset()` method? That would make the invariants easier to follow.

liuneng1994:
The optimization direction looks reasonable, but I think this PR needs more work before it is easy to review and maintain.

  A few concerns:

  1. Please add checked-in tests for the new windowed rep/def path. Local validation is not enough for this kind of reader state-machine change. We should cover at least: window mode disabled,
  window mode enabled across multiple pages, repeated calls to `decodeRepDefs()`, and skip/seek after only part of the window has been decoded.

  2. Please add a focused microbenchmark for this optimization. Since this is a performance change, we need benchmark numbers that isolate the new windowed decode path and compare it with the
  existing preload path on representative nested/repeated data.

  3. Some of the new code looks like it is re-implementing existing low-level logic. For example, `countLeavesInRepDefPage()` hand-parses def-level RLE, and `compactConsumedLeafNulls()`
  duplicates the existing null compaction logic including the memcpy workaround. Can we reuse the existing level decoding / null compaction path, or extract shared helpers instead of duplicating
  this logic?

  4. The readability could be improved by grouping the new window-mode state. The new `window*` fields in `PageReader` are tightly coupled and have reset/compact invariants. A small private
  `WindowRepDefState` struct with a `reset()` method would make the state transitions easier to follow.

  5. I would avoid reading `BOLT_PARQUET_REPDEF_PRELOAD_WINDOW_COUNT` directly inside `PageReader`. Since the PR already adds `RowReaderOptions::setParquetRepDefPreloadWindowCount()`, it is
  cleaner to keep this as explicit option/config plumbing rather than a process-wide env var side channel.

  Overall, I think the PR should aim for: enough reuse with the existing rep/def decode machinery, tests for the new behavior, and a microbenchmark showing the performance win.

PR #787, opened 2026-07-27
perf: enable parquet lazy vectors

lizhen-0710:

Enable Parquet readers to produce top-level LazyVectors for projected columns so table scan can defer payload decode/materialization and aggregation can consume lazy columns through ValueHook.

### Type of Change

- [ ] 📝 Documentation only

### Description

Set the Parquet root reader as top-level and carry the upstream lazy-reader correctness fixes needed for safe lazy loading: page-level hook row bias, offset-aware iota rows, string decoder value counts, and decoded string dictionary hook values. Keep scan behavior unchanged for filter columns: pushed-down filter inputs remain eager while projected payload columns can stay lazy and remaining-filter outputs can preserve wrapped lazy vectors.

Add focused coverage for reader lazy output, complex projected columns, stale lazy lifecycle failures, small-page lazy loads, string dictionary ValueHook row/value alignment, DWRF string dictionary hooks, scan filter/lazy boundaries, remaining-filter wrappers, aggregation pushdown metrics, and iota offset fallback.

Verified with targeted Parquet reader/table-scan tests, DWRF string dictionary hook test, RawVector iota test, staged clang-tidy, and Parquet E2E filter/value-hook coverage across direct and dictionary encodings.

Add the follow-up correctness fixes needed after enabling Parquet lazy vectors.

Vector lifecycle and wrapper handling:
- clear stale nulls when reusing LazyVector loaders
- make the lazy-wrapper consistency flag atomic and release it from constant/dictionary wrappers
- unwrap lazy vectors in flattenVector and load wrapper-over-lazy vectors before compare()

Execution and scan materialization:
- load reused lazy inputs in FilterProject and remaining-filter scan paths before they can be partially loaded
- keep HashProbe filter inputs temporary so lazy wrappers do not outlive filter evaluation
- load lazy input before OrderBy sorting and stop lazy hook pushdown across pre-grouped aggregation split remainders
- avoid double-counting lazy load time in noMoreInput timing

Parquet and connector boundaries:
- keep DCMap and Variant parquet readers eager where children are immediately combined
- materialize Paimon split rows before merge-engine copy/resize/aggregate paths
- unwrap top-level lazy vectors at public Arrow export entry points

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/exec/FilterProject.cpp
+++ b/bolt/exec/FilterProject.cpp
@@ -54,6 +54,21 @@ bool checkAddIdentityProjection(
   return false;
 }
 
+void loadReusedLazyVectors(
+    const RowVectorPtr& input,
+    const std::vector<column_index_t>& reusedInputChannels) {
+  if (!input || reusedInputChannels.empty()) {
+    return;
+  }
+
+  auto& children = input->children();
+  for (auto inputChannel : reusedInputChannels) {
+    if (isLazyNotLoaded(*children[inputChannel])) {
+      children[inputChannel]->loadedVector();
+    }
+  }
+}
+
 // Split stats to attrbitute cardinality reduction to the Filter node.
 std::vector<OperatorStats> splitStats(
     const OperatorStats& combinedStats,
@@ -142,6 +157,17 @@ void FilterProject::initialize() {
     }
     isIdentityProjection_ = true;
   }
+  {
+    std::unordered_map<column_index_t, size_t> inputChannelCounts;
+    for (const auto& projection : identityProjections_) {
+      ++inputChannelCounts[projection.inputChannel];
+    }
+    for (const auto& [inputChannel, count] : inputChannelCounts) {
+      if (count > 1) {
+        reusedInputChannels_.push_back(inputChannel);
+      }
+    }
+  }
   numExprs_ = allExprs.size();
   exprs_ = makeExprSetFromFlag(std::move(allExprs), operatorCtx_->execCtx());
 
@@ -169,7 +195,7 @@ void FilterProject::addInput(RowVectorPtr input) {
   if (!skipForCompositeInput_ || !RowVector::isComposite(input_)) {
     for (auto& childVec : input_->children()) {
       if ((acceptCompositeInput_ && childVec == nullptr) ||
-          childVec->isLazy() ||
+          childVec->isLazy() || isLazyNotLoaded(*childVec) ||
           childVec->encoding() != VectorEncoding::Simple::DICTIONARY) {
         continue;
       }
@@ -240,6 +266,7 @@ RowVectorPtr FilterProject::getOutput() {
     if (isCompositeInput) {
       return fillCompositeOutput(size, results);
     } else {
+      loadReusedLazyVectors(input_, reusedInputChannels_);
       return fillOutput(size, nullptr, results);
     }
   }
@@ -264,6 +291,7 @@ RowVectorPtr FilterProject::getOutput() {
     results = project(*rows, evalCtx);
   }
 
+  loadReusedLazyVectors(input_, reusedInputChannels_);
   return fillOutput(
       numOut,
       allRowsSelected ? nullptr : filterEvalCtx_.selectedIndices,
--- a/bolt/exec/FilterProject.cpp
+++ b/bolt/exec/FilterProject.cpp
@@ -32,6 +32,7 @@
 #include "bolt/core/Expressions.h"
 #include "bolt/expression/Expr.h"
 #include "bolt/expression/FieldReference.h"
+#include "bolt/vector/LazyVector.h"
 #include "bolt/vector/VectorEncoding.h"
 namespace bytedance::bolt::exec {
 namespace {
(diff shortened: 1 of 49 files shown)


bolt/exec/OperatorUtils.cpp
```diff
@@ -297,12 +295,13 @@ std::vector<VectorPtr> wrapChildren(
     }
     if (nulls == nullptr &&
         children[i]->encoding() == VectorEncoding::Simple::DICTIONARY &&
-        children[i]->rawNulls() == nullptr &&
-        !children[i]->valueVector()->containingLazyAndWrapped()) {
+        children[i]->rawNulls() == nullptr) {
+      auto baseValues = children[i]->valueVector();
+      baseValues->clearContainingLazyAndWrapped();
```
guhaiyan0221: > Clearing the wrapped-lazy flag here is unsafe because the original dictionary is still alive in children[i]. The new combined dictionary and the old dictionary can then wrap the same unloaded LazyVector
> and load different row sets; the first partial load makes the second wrapper observe an incomplete backing vector, which reproduces as a wrong value or SIGSEGV. Please transfer ownership only after the
> old wrapper is unreachable, or replace this boolean with ownership/ref-counted semantics, and add a partial-row loader test that keeps both wrappers alive.

bolt/exec/FilterProject.cpp
```diff
@@ -54,6 +54,21 @@ bool checkAddIdentityProjection(
...
+    const RowVectorPtr& input,
+    const std::vector<column_index_t>& reusedInputChannels) {
+  if (!input || reusedInputChannels.empty()) {
+    return;
+  }
+
+  auto& children = input->children();
+  for (auto inputChannel : reusedInputChannels) {
+    if (isLazyNotLoaded(*children[inputChannel])) {
+      children[inputChannel]->loadedVector();
```
guhaiyan0221: > This eagerly materializes the entire input channel for duplicate identity projections, even when the filter keeps only a small subset of rows. That can negate the Parquet lazy-loading benefit and increase
> peak memory. Could we share one output wrapper or load only the selected rows via ensureLoadedRows(), and assert the requested row set in the test?

bolt/exec/OperatorUtils.cpp
lizhen-0710: Agreed. I re-checked this in the context of wrapChildren(): children[i] is still owned by the input RowVector, so this path cannot safely treat the new combined dictionary as having taken ownership of the lazy base.
I removed the clearContainingLazyAndWrapped() call and kept the guard that avoids combining dictionary-over-lazy when the lazy base is already wrapped. This preserves the existing single-wrapper ownership invariant for unloaded LazyVector.
I also added a regression test, OperatorUtilsTest.projectDictionaryOverLazyKeepsExistingLazyWrapper, which keeps the old wrapper alive, verifies the new projection still wraps the existing dictionary instead of the lazy base directly, checks that the ownership flag is not cleared, and validates a partial-row load through the wrapper translation.

bolt/exec/FilterProject.cpp
lizhen-0710: Agreed. The eager loadedVector() call was too broad for the duplicate identity projection case.
I changed this path to use LazyVector::ensureLoadedRows() with the actual output row set instead of materializing the whole input channel. For filtered identity-only projections, the selected row bits are now applied before loading reused lazy input channels, so only rows that survive the filter are requested.
I added regression coverage for selective rows, all rows after filter, no-filter, and no-passing-rows cases. I also added a direct TaskCursor test that avoids the test helper's eager result loading and simulates two duplicate output fields being consumed with different row sets, to verify the output is safe while still preserving lazy loading benefits.

bolt/exec/tests/FilterProjectTest.cpp
```diff
@@ -339,6 +339,204 @@ TEST_F(FilterProjectTest, projectAndIdentityOverLazy) {
   assertQuery(plan, "SELECT c0 < 10 AND c1 < 10, c1 FROM tmp");
 }
 
+TEST_F(FilterProjectTest, duplicateIdentityProjectionOverLazy) {
+  vector_size_t size = 100;
+  auto valueAt = [](auto row) -> int32_t { return row; };
+  auto lazyVector = makeLazyFlatVector<int32_t>(
```
guhaiyan0221: Non-blocking style nit: these duplicate-identity lazy projection tests repeat the same input/table/plan setup several times. The coverage is useful, but it would be easier to maintain if the common lazy input construction and `values -> filter -> project({"c1", "c1"})` assertion were moved into a small helper, leaving each test to specify only the filter and lazy row pattern. The special consumer-safety test can stay explicit.

guhaiyan0221 (approved):

PR #805, opened 2026-07-29
misc: release parquet row group readers before prefetch

lizhen-0710:

Release Parquet column reader state when a row group slides out of the prefetch window. Before loading the next current+prefetch window, the old row group's BufferedInput was erased, but PageReader state such as decompressed page buffers and dictionary string storage could remain attached to column readers until seekToRowGroup() replaced it later.

### Type of Change

- [ ] 📝 Documentation only

### Description

That ordering can transiently overlap old row-group decode state with new DirectBufferedInput prefetch allocations. This is most visible for wide dictionary string columns where the dictionary page is held as both decompressed page data and dictionary string backing storage.

Add an explicit recursive release that clears ScanState and resets ParquetData's PageReader before the new row-group window is loaded. This keeps the steady row-group window bounded without changing read semantics; seekToRowGroup() still constructs the reader for the next active row group.

Add a regression test that writes multiple row groups with a large repeated string dictionary and reads them through a constrained memory pool. The test exercises row-group advancement and verifies the scan stays under the memory limit while reading all row groups.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/reader/StructColumnReader.cpp
+++ b/bolt/dwio/parquet/reader/StructColumnReader.cpp
@@ -172,6 +172,23 @@ StructColumnReader::findBestLeaf() {
   return best;
 }
 
+namespace {
+void releaseRowGroupReaderRecursive(
+    dwio::common::SelectiveColumnReader* reader) {
+  reader->scanState().clear();
+  reader->formatData().as<ParquetData>().releaseRowGroupReader();
+  for (auto* child : reader->children()) {
+    if (child != nullptr) {
+      releaseRowGroupReaderRecursive(child);
+    }
+  }
+}
+} // namespace
+
+void StructColumnReader::releaseRowGroupReader() {
+  releaseRowGroupReaderRecursive(this);
+}
+
 void StructColumnReader::read(
     int64_t offset,
     const RowSet& rows,
--- a/bolt/dwio/parquet/reader/ParquetData.cpp
+++ b/bolt/dwio/parquet/reader/ParquetData.cpp
@@ -437,6 +437,10 @@ dwio::common::PositionProvider ParquetData::seekToRowGroup(int64_t index) {
   return dwio::common::PositionProvider(empty);
 }
 
+void ParquetData::releaseRowGroupReader() {
+  reader_.reset();
+}
+
 std::pair<int64_t, int64_t> ParquetData::getRowGroupRegion(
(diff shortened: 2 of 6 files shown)


bolt/dwio/parquet/reader/StructColumnReader.cpp
```diff
@@ -172,6 +172,23 @@ StructColumnReader::findBestLeaf() {
   return best;
 }
 
+namespace {
+void releaseRowGroupReaderRecursive(
+    dwio::common::SelectiveColumnReader* reader) {
+  reader->scanState().clear();
```
guhaiyan0221, line 178: will this affect lazy column loading?

lizhen-0710: The early release should not affect lazy column loading because of the scan/lazy-vector lifecycle.

A LazyVector produced by the Parquet reader is tied to the current reader pass. The downstream operator must consume it in that pass. If the column data is needed, it will be materialized through the lazy loader before the scan reader is advanced to produce another batch. In other words, the Parquet reader is not expected to keep old row-group PageReader state alive for lazy loads after the scan has moved forward.

This change releases the PageReader only when the scan is advancing from row group i - 1 to row group i:

```
  if (currentGroup >= 1) {
    inputs_.erase(rowGroupIds[currentGroup - 1]);
    reader.releaseRowGroupReader();
  }
```

At that point, row group i - 1 has already been consumed by the scan pipeline. Then seekToRowGroup(i) creates a new PageReader for the current row group before any data from row group i is read:

`  reader_ = std::make_unique<PageReader>(std::move(streams_[index]), ...);`

So the release only drops page-reader state for the previous row group after its lifecycle is complete. It does not release the PageReader that will be used by lazy vectors produced for the current row group. The intent is only to avoid overlapping old row-group decode state with the next row-group prefetch window.

guhaiyan0221: Thank you for the explanation.

guhaiyan0221 (approved):

PR #810, opened 2026-07-31
fix: avoid overlapping scan output buffers

lizhen-0710:

The scan path has two reuse caches: selective readers keep values_ buffers, while HiveDataSource keeps output_ as the backing reader output. When output_ still references the previous batch's leaf value buffers, the reader-side values_ cache is no longer uniquely owned. That causes two problems: the previous output buffers overlap with buffers for the current read and increase peak memory, and values_ cannot be reused so the reader repeatedly allocates replacement value buffers.

### Type of Change

- [ ] 📝 Documentation only

### Description

Prepare the reader output before each splitReader->next() with BaseVector::prepareForReuse(output_, 0). This resets the previous output_'s references to shared leaf values/null/string buffers before the current read, while reusing the existing top-level reader output instead of allocating an extra empty output tree.

The focused Hive tests cover shared and unique backing reuse, flat value/null buffer detach, string buffer detach, lazy leaves, and empty output isolation. The Parquet test adds a VARIANT read-after-reset path, and the exec scan test covers deeply nested ROW/ARRAY/MAP outputs.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/connectors/hive/HiveDataSource.cpp
+++ b/bolt/connectors/hive/HiveDataSource.cpp
@@ -620,6 +620,15 @@ vector_size_t getLength(std::shared_ptr<ConnectorSplit>& split) {
   BOLT_FAIL("Unsupported split type for getting length");
 }
 
+void HiveDataSource::prepareReaderOutputForNextRead() {
+  if (!output_ || *output_->type() != *readerOutputType_) {
+    output_ = BaseVector::create(readerOutputType_, 0, pool_);
+    return;
+  }
+
+  BaseVector::prepareForReuse(output_, 0);
+}
+
 std::optional<RowVectorPtr> HiveDataSource::next(
     uint64_t size,
     bolt::ContinueFuture& /*future*/) {
@@ -639,9 +648,7 @@ std::optional<RowVectorPtr> HiveDataSource::next(
     return nullptr;
   }
 
-  if (!output_) {
-    output_ = BaseVector::create(readerOutputType_, 0, pool_);
-  }
+  prepareReaderOutputForNextRead();
 
   // TODO Check if remaining filter has a conjunct that doesn't depend on
   // any column, e.g. rand() < 0.1. Evaluate that conjunct first, then scan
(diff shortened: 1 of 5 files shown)


bolt/connectors/hive/HiveDataSource.cpp
```diff
@@ -639,9 +639,7 @@ std::optional<RowVectorPtr> HiveDataSource::next(
     return nullptr;
   }
 
-  if (!output_) {
-    output_ = BaseVector::create(readerOutputType_, 0, pool_);
-  }
+  output_ = BaseVector::create(readerOutputType_, 0, pool_);
```
guhaiyan0221: performance concern: this now allocates a new top-level RowVector for every batch. The overhead is likely small compared with the child value buffers, but it gives up top-level vector reuse. Have you benchmarked small-batch scans?

lizhen-0710: Good point. The main cost we are trying to avoid here is the large leaf buffer allocation, especially the reader-side values_ buffer allocation.

The problematic case is that HiveDataSource::output_ and the selective reader values_ are both used as reuse caches. When the previous output_ still references the leaf values buffer, the reader-side values_ is no longer uniquely owned, so the next read cannot reuse it and has to allocate a new values_ buffer. This causes the previous output buffer and the current read buffer to coexist, increasing peak memory.

The earlier version fixed the overlap by recreating the reader output from an empty output structure before each read. That worked for memory peak, but it could introduce unnecessary top-level/vector-tree construction and small 0-row buffer allocation overhead.

In the latest version, I changed this to reuse the top-level reader output and call BaseVector::prepareForReuse(output_, 0) before splitReader->next(). This keeps the top-level RowVector reuse path, recursively prepares the children, and detaches shared leaf values/null/string buffers before the current read. So we still avoid the large values_ buffer overlap/reallocation issue, but without the extra empty-output/vector-tree allocation overhead from the earlier version.

guhaiyan0221 (approved):

PR #811, opened 2026-07-31
perf: Optimize fused nested level conversion

liuneng1994:

Parquet nested reads currently convert the same DataPage def/rep levels separately when a nullable struct directly wraps a list or map child: once for repeated list/map lengths and validity, and once for the parent struct null bitmap. This repeats the same level traversal on a hot nested-read path.

This PR adds a fused conversion path for the direct nullable struct -> list/map case. The reader now attempts to populate the repeated lengths/list validity and parent struct null bitmap in one pass, then falls back to the existing separate conversions when the level relationship is not supported.

Issue Number: N/A

### Type of Change
- [ ] 📝 Documentation only

### Description

This PR adds:

- `DefRepLevelsToListLengthsAndStructBitmap`, which fuses direct list/map length conversion with direct parent struct bitmap conversion.
- A `PageReader::getListLengthsAndStructNulls` wrapper used by nested Parquet readers.
- Reader-side wiring for direct struct -> list/map children, with fallback to the original separate conversion path when the fused level relationship is not supported.
- Oracle-based level conversion tests comparing fused output against the existing separate conversions.
- End-to-end Parquet reader coverage in `ParquetReaderTest` for continuous reads, non-contiguous `skip + next` reads, and top-level filtering.
- Level conversion microbenchmarks for fused fast path and unsupported fallback path.

The fused reader path avoids setting repeated lengths twice by delaying the old repeated setup until the fused path either succeeds or explicitly falls back.

### Performance Impact
    <details>

    ```text
    Benchmark: bolt_dwio_parquet_level_conversion_benchmark
    Comparison: separateStructAndList_* vs fusedStructAndList_*

    Fast-path fused conversion speedups:
    - single lists: ~1.50x to 1.74x
    - mixed lists:  ~1.85x to 2.17x
    - long lists:   ~1.98x to 1.99x

    Unsupported fallback path:
    - fallbackStructAndList_* vs separateStructAndList_* stayed within ~99.0% to 101.1%
    - no systematic regression observed when the fused fast path is not applicable
    ```
    </details>

### Release Note

Release Note:

```text
- Optimized Parquet nested level conversion for direct nullable struct -> list/map reads by fusing repeated length and struct null bitmap generation.
```

### Checklist (For Author)

Validation run locally:

```text
cmake --build --preset conan-release --target bolt_dwio_parquet_level_conversion_benchmark bolt_dwio_parquet_arrow_test bolt_dwio_parquet_reader_test --parallel 16

bolt_dwio_parquet_arrow_test --gtest_filter='NestedListTest.*'
bolt_dwio_parquet_reader_test --gtest_filter='ParquetReaderTest.fusedLevelConversion*'
bolt_dwio_parquet_level_conversion_benchmark --bm_regex='(separateStructAndList|fallbackStructAndList)_'
clang-format --dry-run --Werror on touched C++ files
git diff --check
```

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    - N/A
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/reader/RepeatedColumnReader.cpp
+++ b/bolt/dwio/parquet/reader/RepeatedColumnReader.cpp
@@ -37,10 +37,82 @@ namespace bytedance::bolt::parquet {
 class ParquetTypeWithId;
 
 namespace {
+arrow::ValidityBitmapInputOutput prepareRepeatedNulls(
+    BufferPtr& nulls,
+    int32_t maxItems,
+    memory::MemoryPool& memoryPool) {
+  dwio::common::ensureCapacity<uint64_t>(
+      nulls, bits::nwords(maxItems), &memoryPool);
+  arrow::ValidityBitmapInputOutput output;
+  output.values_read_upper_bound = maxItems;
+  output.values_read = 0;
+  output.null_count = 0;
+  output.valid_bits = reinterpret_cast<uint8_t*>(nulls->asMutable<uint64_t>());
+  output.valid_bits_offset = 0;
+  return output;
+}
+
+bool isArrayOrMap(const dwio::common::SelectiveColumnReader& reader) {
+  const auto kind = reader.fileType().type()->kind();
+  return kind == TypeKind::ARRAY || kind == TypeKind::MAP;
+}
+
+void setRepeatedRepDefsFromLeaf(
+    dwio::common::SelectiveColumnReader& reader,
+    PageReader& pageReader) {
+  if (auto* list = dynamic_cast<ListColumnReader*>(&reader)) {
+    list->setLengthsFromRepDefs(pageReader);
+  } else {
+    auto* map = dynamic_cast<MapColumnReader*>(&reader);
+    BOLT_CHECK_NOT_NULL(map);
+    map->setLengthsFromRepDefs(pageReader);
+  }
+}
+
+template <typename RepeatedReader>
+bool trySetFusedStructAndRepeatedRepDefs(
+    PageReader& pageReader,
+    StructColumnReader& structReader,
+    RepeatedReader& repeatedReader) {
+  const auto repDefRange = pageReader.repDefRange();
+  const int32_t numRepDefs = repDefRange.second - repDefRange.first;
+  auto lengths = repeatedReader.prepareRepDefLengths(numRepDefs);
+  auto repeatedBits = repeatedReader.prepareRepDefNulls(numRepDefs);
+  auto structBits = structReader.prepareRepDefNulls(numRepDefs);
+  if (!pageReader.getListLengthsAndStructNulls(
+          repeatedReader.levelInfo(),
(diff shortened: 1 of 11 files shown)


bolt/dwio/parquet/reader/RepeatedColumnReader.cpp
```diff
@@ -36,10 +36,82 @@ namespace bytedance::bolt::parquet {
...
+}
+
+template <typename RepeatedReader>
+bool trySetFusedStructAndRepeatedRepDefs(
+    PageReader& pageReader,
+    StructColumnReader& structReader,
+    RepeatedReader& repeatedReader) {
+  const auto repDefRange = pageReader.repDefRange();
+  const int32_t numRepDefs = repDefRange.second - repDefRange.first;
+  auto lengths = repeatedReader.prepareRepDefLengths(numRepDefs);
```
guhaiyan0221, line 96: Please avoid destructively preparing the output buffers before checking whether the level shape is supported. For a valid but unsupported shape, such as a required list directly under the struct, prepareRepDefLengths() moves the reusable lengths_.lengths() buffer into the local variable, then the fused call returns false and the local buffer is destroyed. The fallback at line 165 can no longer reuse it and must allocate a new buffer on every batch. Please perform the cheap LevelInfo shape check before moving/allocating the outputs, or restore the lengths buffer before returning false.

bolt/dwio/parquet/reader/RepeatedColumnReader.cpp
```diff
@@ -56,26 +128,46 @@ PageReader* FOLLY_NULLABLE readLeafRepDefs(
...
+    }
     return pageReader;
   }
   if (auto structReader = dynamic_cast<StructColumnReader*>(reader)) {
-    pageReader = readLeafRepDefs(structReader->childForRepDefs(), numTop, true);
+    auto repDefChild = structReader->childForRepDefs();
+    if (repDefChild == nullptr) {
+      return nullptr;
+    }
+    const bool fuseStructAndRepeated = isArrayOrMap(*repDefChild);
```
guhaiyan0221, line 167: Could we restrict this optimization to nullable structs, e.g. structReader->levelInfo().def_level > 0? For a required struct, the previous setNullsFromRepDefs() path is a no-op, so there are not two conversions to fuse. The new path still allocates and writes an all-valid struct bitmap and installs it as preset nulls, adding work to a path that previously only converted the list.

liuneng1994: done

bolt/dwio/parquet/reader/RepeatedColumnReader.cpp
liuneng1994: done

guhaiyan0221 (approved):

PR #812, opened 2026-07-31
fix(parquet): limit rows per data page to prevent overflow

### LINKED ISSUE #800: [Bug] Parquet writer can overflow DataPage row/value counts for highly compressible columns
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

The Bolt Parquet writer can keep a highly compressible column in one DataPage
for an extremely large number of logical rows.

Before this fix, a page was closed mainly when its estimated encoded value size
reached `data_pagesize`. This is not a sufficient bound for dictionary-encoded,
low-cardinality, or all-null columns: their encoded value buffers may remain
small while the logical value/row counters continue to grow.

Parquet stores `DataPageHeader.num_values` and the corresponding DataPageV2
counts as signed 32-bit integers. Once the writer's 64-bit in-memory counter
exceeds `INT32_MAX`, the unchecked narrowing conversion produces an invalid
page header. The write can still finish successfully, but a conforming reader
later rejects the file.

One production file contained:

```text
ColumnMetaData.num_values = 2,785,305,817
DataPageHeader.num_values = -1,509,661,479
```

The second value is the signed 32-bit result of the first value modulo
`2^32`:

```text
2,785,305,817 - 4,294,967,296 = -1,509,661,479
```

### Reproduction Steps

1. Write a Parquet column using dictionary encoding and compression.
2. Use a low-cardinality `BYTE_ARRAY` value, or an optional column containing
   mostly/all nulls, so that encoded bytes grow much more slowly than logical
   row count.
3. Allow a very large row group and repeatedly append batches to the same
   column writer.
4. Continue until the current DataPage contains more than `INT32_MAX` logical
   values.
5. Close the writer and read the generated file.

The writer can report success, while the reader fails with an error similar to:

```text
Expected -1509661479 values in column chunk ... but got 0 values instead
```

### Bolt Version / Commit ID

main branch @bd7d19c

### System Configuration

```markdown
- **Build Type**: Release
- **CPU Arch**: AMD64
- **Framework**: Spark with Gluten/Bolt
```

### Logs / Stack Trace

```shell
Expected -1509661479 values in column chunk at offset 32065246
but got 0 values instead over 0 pages ending at file offset 32065223
```

### Expected Behavior

The writer should:

1. Close a DataPage when either its byte-size limit or logical row-count limit
   is reached.
2. Use a safe default maximum row count per page.
3. Validate every count before converting it to a Parquet page-header `int32`.
4. Fail during writing, instead of silently producing a file that only fails
   during a later read, if an unrepresentable page is ever encountered.

### Additional context

Apache Arrow added the same independent row limit because a byte-only limit can
produce excessively large pages for highly compressible data:

- [Apache Arrow issue #47030](https://github.com/apache/arrow/issues/47030)
- [Arrow C++ row-per-page implementation](https://apache.googlesource.com/arrow/+/b2190db54c7ee31a6ce3c3990a66020345471c99%5E%21/)
- [Parquet Thrift definitions](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift)

The proposed default is 20,000 rows per DataPage, matching current Arrow C++,
Arrow Rust, and parquet-java behavior.

olemon111:

Bolt's Parquet writer could generate a file that completed successfully during
write but failed later during read with an error like:

```text
Expected -1509661479 values in column chunk at offset 32065246
but got 0 values instead over 0 pages ending at file offset 32065223
```

The invalid negative value came from an overflowing DataPage count. The writer
previously used the estimated encoded byte size as the main DataPage flush
condition. For highly compressible data, especially dictionary-encoded
low-cardinality `BYTE_ARRAY` values and all-null columns, billions of logical
values could accumulate in one page while the encoded buffer remained below the
configured byte limit.

Issue Number: close #800 

### Type of Change

- [ ] 📝 Documentation only

### Description

This PR:

- adds `WriterProperties::max_rows_per_page` with an initial conservative
  default of `INT32_MAX / 2` (`1,073,741,823`) rows;
- exposes the limit as Bolt `WriterOptions::maxRowsPerDataPage` and passes it
  into the Arrow-derived Parquet writer properties;
- makes block-size-based buffered writes honor the row count carried by an
  existing custom `flushPolicyFactory`, without adding another
  `WriterOptions` field; this lets existing integrations enforce both row and
  compressed-byte RowGroup limits;
- changes all relevant column-writing paths to split work according to both
  the internal write-batch size and the remaining row capacity of the current
  DataPage;
- preserves record boundaries for repeated/nested columns where page-index or
  DataPageV2 semantics require record-aligned pages;
- flushes a DataPage when the encoded byte threshold, logical row threshold,
  or internal value-count safety threshold is reached;
- validates the configured row limit against the safe range and uses unsigned
  arithmetic when serializing RLE run lengths;
- validates DataPage V1/V2 counts before narrowing them to the Parquet
  page-header `int32` fields, providing a fail-fast safety net;
- adds property validation and regression coverage for:
  - default and custom row limits;
  - limits outside the supported range;
  - all-null optional columns with DataPage V1 and V2;
  - ZSTD-compressed, dictionary-encoded, low-cardinality `BYTE_ARRAY` columns;
  - repeated columns and record-boundary page splitting;
  - the high-level Bolt writer round trip.

The incoming Spark/Gluten RecordBatch split is intentionally not used as the
correctness boundary. Multiple input batches can be appended to the same open
DataPage, so the limit must be enforced inside `ColumnWriter`.

The initial default is deliberately set to a high value. The purpose of this PR
is to prevent an `int32` overflow without aggressively changing the DataPage
layout of existing workloads. 

This value is a correctness-oriented initial threshold, not a claim that it is
the optimal DataPage row count. Apache Arrow C++, Arrow Rust, and parquet-java
use a much smaller 20K row default to balance page-index pruning, cache behavior,
metadata size, and compression overhead. Selecting a lower Bolt default and
quantifying its write/read performance impact are intentionally left to a
follow-up change with representative benchmarks.

- https://github.com/apache/arrow/issues/47030
- https://github.com/apache/arrow/pull/47090
- https://github.com/apache/parquet-java/commit/910bcc4edc2d

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

No performance improvement is claimed by this correctness fix. The initial
`INT32_MAX / 2` threshold is intentionally high, so ordinary columns and even
very large but valid pages continue to be governed by the existing encoded-byte
threshold. Only pages approaching the Parquet `int32` representation boundary
are split by the new count limits.

The exact row threshold can affect page count, metadata size, compression
finalization overhead, peak memory, cache behavior, and page-index pruning.
Choosing a lower default therefore requires dedicated write/read benchmarks on
representative normal, all-null, low-cardinality dictionary, and nested data.
That tuning and performance evaluation will be handled separately rather than
mixed into this correctness fix.

### Release Note

Release Note:

```text
Release Note:
Fix the Parquet writer so highly compressible columns cannot overflow DataPage
row/value count fields. DataPages now use a high `INT32_MAX / 2` default row
limit plus an internal value-count safety threshold, and page-header counts are
validated before serialization.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/dwio/parquet/arrow/ColumnWriter.cpp
+++ b/bolt/dwio/parquet/arrow/ColumnWriter.cpp
@@ -1623,6 +1623,15 @@ inline void DoInBatchesNonRepeated(
     int64_t max_batch_size = std::min(batch_size, num_levels - offset);
     max_batch_size =
         std::min(max_batch_size, max_rows_per_page - page_buffered_rows);
+    if (max_batch_size <= 0) {
+      throw ParquetException(
+          "Non-repeated column batching cannot make progress: batch size ",
+          batch_size,
+          ", buffered rows ",
+          page_buffered_rows,
+          ", maximum rows per page ",
+          max_rows_per_page);
+    }
     const int64_t end_offset = offset + max_batch_size;
 
     DCHECK_LE(offset, end_offset);
@@ -1645,6 +1654,11 @@ inline void DoInBatchesRepeated(
   int64_t offset = 0;
   while (offset < num_levels) {
     const int64_t max_batch_size = std::min(batch_size, num_levels - offset);
+    if (max_batch_size <= 0) {
+      throw ParquetException(
+          "Repeated column batching cannot make progress: batch size ",
+          batch_size);
+    }
     int64_t end_offset = num_levels;
     int64_t check_page_limit_end_offset = -1;
     int64_t page_buffered_rows = current_page_buffered_rows();
@@ -1696,7 +1710,7 @@ inline void DoInBatches(
     bool pages_change_on_record_boundaries,
     Action&& action,
     GetBufferedRows&& current_page_buffered_rows) {
-  if (rep_levels == nullptr || !pages_change_on_record_boundaries) {
+  if (rep_levels == nullptr) {
     DoInBatchesNonRepeated(
         num_levels,
         batch_size,
@@ -2280,7 +2294,7 @@ class TypedColumnWriterImpl : public ColumnWriterImpl,
     num_buffered_encoded_values_ += num_values;
     num_buffered_nulls_ += num_nulls;
 
-    if (check_page_limit &&
+    if (num_buffered_values_ > 0 && check_page_limit &&
         (current_encoder_->EstimatedDataEncodedSize() >= data_pagesize_ ||
          num_buffered_rows_ >= properties_->max_rows_per_page())) {
       AddDataPage();
--- a/bolt/dwio/parquet/arrow/ColumnWriter.cpp
+++ b/bolt/dwio/parquet/arrow/ColumnWriter.cpp
@@ -100,6 +100,8 @@ using util::CodecOptions;
 namespace {
 
 constexpr int64_t kMaxPageHeaderSize = std::numeric_limits<int32_t>::max();
+constexpr int64_t kDataPageValueCountFlushThreshold =
+    std::numeric_limits<int32_t>::max() / 2;
 
 int32_t checkPageHeaderSize(std::string_view size_name, int64_t size) {
   if (size < 0 || size > kMaxPageHeaderSize) {
@@ -2154,6 +2156,7 @@ class TypedColumnWriterImpl : public ColumnWriterImpl,
       int64_t num_values,
       const int16_t* def_levels,
       const int16_t* rep_levels) {
+    CheckPageValueCountCanBeBuffered(num_values);
     int64_t values_to_write = 0;
     // If the field is required and non-repeated, there are no definition levels
     if (descr_->max_definition_level() > 0) {
@@ -2263,6 +2266,7 @@ class TypedColumnWriterImpl : public ColumnWriterImpl,
       int64_t num_levels,
       const int16_t* def_levels,
       const int16_t* rep_levels) {
+    CheckPageValueCountCanBeBuffered(num_levels);
     // If the field is required and non-repeated, there are no definition levels
     if (descr_->max_definition_level() > 0) {
       WriteDefinitionLevels(num_levels, def_levels);
@@ -2296,8 +2300,21 @@ class TypedColumnWriterImpl : public ColumnWriterImpl,
 
     if (num_buffered_values_ > 0 && check_page_limit &&
         (current_encoder_->EstimatedDataEncodedSize() >= data_pagesize_ ||
-         num_buffered_rows_ >= properties_->max_rows_per_page())) {
+         num_buffered_rows_ >= properties_->max_rows_per_page() ||
+         num_buffered_values_ >= kDataPageValueCountFlushThreshold)) {
       AddDataPage();
+    }
+  }
+
+  void CheckPageValueCountCanBeBuffered(int64_t num_levels) const {
+    if (num_levels < 0 ||
+        num_levels > kMaxPageHeaderSize - num_buffered_values_) {
+      throw ParquetException(
+          "DataPage num_values cannot be represented in a Parquet page "
+          "header int32 field: buffered ",
+          num_buffered_values_,
+          ", incoming ",
+          num_levels);
     }
   }
 
--- a/bolt/dwio/parquet/arrow/ColumnWriter.cpp
+++ b/bolt/dwio/parquet/arrow/ColumnWriter.cpp
@@ -112,6 +112,16 @@ int32_t checkPageHeaderSize(std::string_view size_name, int64_t size) {
   return static_cast<int32_t>(size);
 }
 
+int32_t checkPageHeaderCount(std::string_view count_name, int64_t count) {
+  if (count < 0 || count > std::numeric_limits<int32_t>::max()) {
+    throw ParquetException(
+        std::string(count_name),
+        " cannot be represented in a Parquet page header int32 field: ",
+        count);
+  }
+  return static_cast<int32_t>(count);
+}
+
 int64_t bufferAllocatedBytes(const std::shared_ptr<Buffer>& buffer) {
   return buffer != nullptr ? buffer->capacity() : 0;
 }
@@ -1442,7 +1452,8 @@ void ColumnWriterImpl::BuildDataPageV1(
     compressed_data = std::move(combined);
   }
 
-  int32_t num_values = static_cast<int32_t>(num_buffered_values_);
+  const int32_t num_values =
+      checkPageHeaderCount("DataPageV1 num_values", num_buffered_values_);
   int64_t first_row_index = rows_written_ - num_buffered_rows_;
 
   // Write the page to OutputStream eagerly if there is no dictionary or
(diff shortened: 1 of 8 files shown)


bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -1594,74 +1608,114 @@ void ColumnWriterImpl::FlushBufferedDataPages() {
...
-            last_record_begin_offset - offset,
-            /*check_page_size=*/true);
-        offset = last_record_begin_offset;
-      }
+    DCHECK_LE(offset, end_offset);
+    DCHECK_LE(check_page_limit_end_offset, end_offset);
 
-      // There is no record boundary in this chunk and cannot check page size.
-      action(offset, end_offset - offset, /*check_page_size=*/false);
+    if (check_page_limit_end_offset >= 0) {
```
luozenglin, line 1687: `check_page_limit_end_offset` can equal `offset` whenever the batch begins at a record boundary, including the single-record case. This invokes `action(offset, 0, true)`.

With dictionary encoding, an empty encoder still reports the one-byte bit-width prefix. Therefore, with a small `data_pagesize` (for example, 1), `CommitWriteAndCheckPageLimit()` can call `AddDataPage()` while `num_buffered_values_ == 0`. When the page index is enabled, this empty page and the following real page can receive the same `first_row_index`, violating the OffsetIndex requirement that `first_row_index` be strictly increasing.

Could we skip the zero-length action, or guard page flushing with `num_buffered_values_ > 0`? Please also add a regression test using a repeated INT32 column with `rep_levels = {0}`, dictionary encoding, page index enabled, and `data_pagesize(1)`, asserting that no zero-value page is written and that `first_row_index` is strictly increasing.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -1594,74 +1608,114 @@ void ColumnWriterImpl::FlushBufferedDataPages() {
...
+inline void DoInBatches(
+    const int16_t* def_levels,
+    const int16_t* rep_levels,
+    int64_t num_levels,
+    int64_t batch_size,
+    int64_t max_rows_per_page,
+    bool pages_change_on_record_boundaries,
+    Action&& action,
+    GetBufferedRows&& current_page_buffered_rows) {
+  if (rep_levels == nullptr || !pages_change_on_record_boundaries) {
```
luozenglin: When `rep_levels != nullptr` but `pages_change_on_record_boundaries` is false, as in the common DataPageV1/no-page-index path, this dispatches to `DoInBatchesNonRepeated()`.

That helper assumes every level entry represents one row, whereas the writer increments `num_buffered_rows_` only when `rep_levels[i] == 0`. Consequently, the batching capacity is calculated in level entries rather than logical rows. For example, with a large byte limit, `max_rows_per_page = 1`, and `rep_levels = {0, 1, 1, 0, 1, 1}`, this can produce pages containing 1, 3, and 2 levels instead of two complete three-level records.

It can also degenerate into one encoder/statistics call per repetition level for a long record when the current page has little remaining row capacity, even though `rep_level == 1` does not consume that capacity.

The Apache Arrow implementation being ported dispatches solely on whether `rep_levels` is present:
https://github.com/apache/arrow/commit/b2190db54c7ee31a6ce3c3990a66020345471c99

Could we use the repeated-aware helper whenever `rep_levels != nullptr` and add a DataPageV1/no-page-index regression test with a long repeated record?

bolt/dwio/parquet/arrow/ColumnWriter.cpp
```diff
@@ -1594,74 +1608,114 @@ void ColumnWriterImpl::FlushBufferedDataPages() {
...
+    int64_t batch_size,
+    int64_t max_rows_per_page,
+    Action&& action,
+    GetBufferedRows&& current_page_buffered_rows) {
+  int64_t offset = 0;
+  while (offset < num_levels) {
+    const int64_t page_buffered_rows = current_page_buffered_rows();
+    DCHECK_LE(page_buffered_rows, max_rows_per_page);
+
+    int64_t max_batch_size = std::min(batch_size, num_levels - offset);
```
luozenglin, line 1625: These new loops require `batch_size > 0` in order to make progress, but `WriterProperties::Builder::write_batch_size()` currently accepts zero and negative values.

With zero, `max_batch_size` is zero, `end_offset == offset`, and the loop never terminates. With a negative value, `offset` can move backwards in release builds after the DCHECKs are compiled out.

Could we reject `write_batch_size <= 0` when building the properties and retain a defensive progress check here? Tests for both zero and negative values would prevent this from becoming a configuration-triggered hang.

luozenglin (commented):
The Performance Impact section selects “Positive Impact: I have run benchmarks”, but the reported numbers are still template placeholders.

Did you really execute the benchmark or did you check it incorrectly? Can you provide real benchmark results and explain why there is a performance improvement?

olemon111:
> The Performance Impact section selects “Positive Impact: I have run benchmarks”, but the reported numbers are still template placeholders.
> 
> Did you really execute the benchmark or did you check it incorrectly? Can you provide real benchmark results and explain why there is a performance improvement?

You're right. I didn't run that benchmark. The numbers were template placeholders. I'll remove them and mark this as a possible negative impact rather than claiming a performance improvement.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: Thanks, fixed. Page flushing now requires num_buffered_values_ > 0, preventing empty pages while preserving checks for previously buffered data. The test also verifies strictly increasing OffsetIndex row IDs.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: Fixed. I changed the dispatch to use `DoInBatchesRepeated()` whenever `rep_levels` is
present, while keeping `pages_change_on_record_boundaries` as an input to that
helper for its final page-check decision. I also added a DataPageV1/no-page-index
test covering multiple three-level records and a long repeated record.

bolt/dwio/parquet/arrow/ColumnWriter.cpp
olemon111: Agreed. The new loops require a strictly positive write batch size, and DCHECKs
are not sufficient because they are compiled out in release builds. I added
runtime validation in `WriterProperties::Builder::write_batch_size()` for zero
and negative values, plus a defensive progress check in the batching path. The
properties tests now cover 0, -1, and a valid positive value.

luozenglin (approved):
LGTM

PR #813, opened 2026-08-01
fix: handle float-to-integer overflow boundary

### LINKED ISSUE #809: [Bug] Spark float-to-integer casts return minimum at positive overflow boundary
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

In Spark-compatible mode, Bolt returns incorrect results when a
runtime-computed floating-point value is exactly the first value outside the
positive range of the destination integer type:

- `FLOAT 2^31` cast to `INT`
- `DOUBLE 2^63` cast to `BIGINT`

Java Spark saturates these values to the destination type's maximum value.
Bolt native instead returns the destination type's minimum value.

The expressions intentionally depend on `id` so Catalyst cannot constant-fold
the casts before native execution.

### Reproduction Steps

Run with ANSI mode disabled:

```sql
SET spark.sql.ansi.enabled=false;

WITH src AS (
  SELECT id FROM range(0, 2)
)
SELECT
  id,
  CAST(
    CAST(
      CAST(id AS FLOAT) * CAST('2147483648' AS FLOAT)
      AS FLOAT
    )
    AS INT
  ) AS float_to_int,
  CAST(
    CAST(id AS DOUBLE) * CAST('9223372036854775808' AS DOUBLE)
    AS BIGINT
  ) AS double_to_bigint
FROM src
ORDER BY id;
```

Bolt native result:

```text
id  float_to_int  double_to_bigint
0   0             0
1   -2147483648   -9223372036854775808
```

Java Spark result:

```text
id  float_to_int  double_to_bigint
0   0             0
1   2147483647    9223372036854775807
```

### Bolt Version / Commit ID

`8040d32a87d951349544089134a66cf91e86e214`

### System Configuration

- **OS**: Linux
- **Compiler**: Clang 22 with libc++
- **Build Type**: Release
- **CPU Arch**: x86_64
- **Framework**: Spark, comparing Bolt native execution with Java Spark execution

### Expected Behavior

Bolt's Spark-compatible cast should match Java Spark:

```text
id  float_to_int  double_to_bigint
0   0             0
1   2147483647    9223372036854775807
```

### Additional context

The relevant path appears to be the floating-point-to-integral conversion in
`bolt/expression/CastExpr-tpl.cpp`.

For these destination types, converting
`std::numeric_limits<ToType>::max()` to the source floating-point type rounds
it up:

- `INT_MAX` becomes `FLOAT 2^31`
- `INT64_MAX` becomes `DOUBLE 2^63`

As a result, a strict comparison such as
`from > std::numeric_limits<ToType>::max()` does not identify the exact `2^N`
boundary as overflow after the usual conversion. The subsequent out-of-range
C++ floating-point-to-integer conversion is undefined behavior; on this x86_64
build it produces the integer minimum value.

Pybsama:

Spark-compatible floating-point-to-integer casts currently compare the input
against the target's maximum after converting that maximum to the floating-point
source type. `INT_MAX` rounds to `2^31` as a `float`, and `INT64_MAX` rounds to
`2^63` as a `double`. Equality therefore misses the first positive out-of-range
value and lets it enter an undefined C++ conversion, which can produce the
target's minimum value.

Issue Number: close #809

### Type of Change

- [ ] 📝 Documentation only

### Description

- Derive an exactly representable exclusive upper bound from the negative
  integer minimum and use `>=` against that boundary.
- Keep `int32_t` as the first-stage type for tinyint and smallint so Spark's
  Java-style two-stage narrowing behavior remains unchanged.
- Share the boundary helper between the general conversion utility and the
  Spark Cast expression path.
- Cover the last in-range floating-point value, the positive overflow boundary,
  and the negative minimum for float/double conversions to all integral widths.

Using `>= max()` would reject a valid `double(INT_MAX)`. In contrast, the
negated minimum is a power of two and is exactly representable in the relevant
floating-point type, so it identifies the first positive out-of-range value
without narrowing the valid range.

Current-main verification (2026-08-29):

- Reapplied the exact four-file `+67/-6` change to upstream `main` at
  `d5d50bb2a430fb52ccaaecb277fa076163eb3ff2`; it applies without conflicts.
- The targeted Release build completed all 675 compile/link steps for
  `bolt_type_test` and `bolt_functions_spark_test`. The macOS-only generated
  build shim used for current main is outside the repository and the source diff
  remains limited to the four PR files.
- `ConversionsTest.*`: 5 passed.
- `SparkCastExprTest.*`: 23 passed, including the new overflow boundaries.
- Full `bolt_type_test`: 213 passed, 1 skipped, 0 failed (214 total).
- Spark SQL test binary excluding two data-dependent performance tests:
  716 passed, 4 skipped, 0 failed (720 total).
- The two excluded tests fail before their test logic because this checkout does
  not contain `data/large_slow_parse.json`; no Cast or conversion test was
  excluded.
- Repository pre-commit checks, including clang-format, C++ lint, license
  headers, secret detection, and codespell, passed; `git diff --check` passed.
- Independent final review reported no Critical, Important, or Minor findings
  and marked the change ready.
- A standalone UBSan reproduction confirms that the previous `2^31f`
  conversion enters undefined behavior.

### Performance Impact

- [x] **No Impact**: The hot path still performs one comparison against a
  compile-time-derived boundary; no allocation, loop, or additional branch is
  introduced.

### Release Note

Release Note:

```text
Release Note:
- Fixed Spark-compatible float and double casts at positive integer overflow boundaries.
```

### Checklist (For Author)

- [x] I have verified the code with a local Release build.
- [x] (Optional) I have run UBSan locally on the boundary reproduction.

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/expression/CastExpr-tpl.cpp
+++ b/bolt/expression/CastExpr-tpl.cpp
@@ -368,7 +368,7 @@ class Converter {
         constexpr TypeKind originKind = ToKind::storage;
         using LimitType = typename util::
             Converter<originKind, void, util::TruncateCastPolicy>::LimitType;
-        if (from > LimitType::maxLimit()) {
+        if (LimitType::isPositiveOverflow(from)) {
           to = LimitType::max();
           return ConvertStatus::INTEGER_OVERFLOW;
         }
--- a/bolt/functions/sparksql/tests/SparkCastExprTest.cpp
+++ b/bolt/functions/sparksql/tests/SparkCastExprTest.cpp
@@ -655,6 +655,18 @@ TEST_F(SparkCastExprTest, overflow) {
   testCast<double, int8_t>("tinyint", {127.8}, {127});
   testCast<double, int8_t>("tinyint", {129.9}, {-127});
   testCast<double, int16_t>("smallint", {1234567.89}, {-10617});
+  testCast<float, int32_t>(
+      "integer",
+      {0x1.fffffep30f, 0x1p31f, -0x1p31f},
+      {2'147'483'520,
+       std::numeric_limits<int32_t>::max(),
+       std::numeric_limits<int32_t>::min()});
+  testCast<double, int64_t>(
+      "bigint",
+      {0x1.fffffffffffffp62, 0x1p63, -0x1p63},
+      {9'223'372'036'854'774'784LL,
+       std::numeric_limits<int64_t>::max(),
+       std::numeric_limits<int64_t>::min()});
   testCast<double, int64_t>(
       "bigint", {std::numeric_limits<double>::max()}, {9223372036854775807});
   testCast<double, int64_t>(
--- a/bolt/type/Conversions.h
+++ b/bolt/type/Conversions.h
@@ -323,6 +323,17 @@ struct Converter<
   struct LimitType {
     static constexpr bool kByteOrSmallInt =
         std::is_same_v<T, int8_t> || std::is_same_v<T, int16_t>;
+    using FirstStageType = std::conditional_t<kByteOrSmallInt, int32_t, T>;
+
+    template <typename FP>
+    static bool isPositiveOverflow(const FP& value) {
+      // Integer max may round up to the first out-of-range value when it is
+      // converted to FP. The negative minimum is a power of two, so its
+      // negation gives an exact exclusive upper bound.
+      return value >=
+          -static_cast<FP>(std::numeric_limits<FirstStageType>::min());
+    }
+
     static int64_t minLimit() {
       if (kByteOrSmallInt) {
         return std::numeric_limits<int32_t>::min();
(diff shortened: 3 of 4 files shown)


Pybsama:
Current-main verification is complete (2026-08-29).

- Reapplied the exact four-file `+67/-6` patch cleanly to upstream `main` at `d5d50bb2a430fb52ccaaecb277fa076163eb3ff2`; the verified local tree is `27efab9adfe67a0ebf4917a9fbb0715c010661f5`.
- Rebuilt and linked the Release `bolt_type_test` and `bolt_functions_spark_test` targets (675/675 compile/link steps after the patch).
- `ConversionsTest.*`: 5/5 passed.
- `SparkCastExprTest.*`: 23/23 passed, including the overflow-boundary regression.
- Full `bolt_type_test`: 213 passed, 1 skipped, 0 failed.
- Spark SQL excluding only the two tests requiring absent `data/large_slow_parse.json`: 716 passed, 4 skipped, 0 failed.
- Repository pre-commit checks and `git diff --check` passed.
- Independent final review found no Critical, Important, or Minor issues and marked the change ready.

The public head remains the existing GitHub Verified single commit; current-main verification did not change the source patch.

PR #816, opened 2026-08-02
fix: correct shuffle compression buffer capacity

### LINKED ISSUE #817: [Bug] Shuffle compression advertises capacity beyond the output allocation
### Component Selection

- [ ] Build
- [x] Other — Spark SQL native shuffle compression

### Describe the Bug

`BlockPayload::fromBuffers` can advertise more codec output capacity than is
actually available when compressing a nonempty Spark SQL shuffle buffer.

The payload allocation reserves a 16-byte record header followed by the
codec's maximum compressed length. `compressBuffer` writes that header and
advances the output pointer past it. Before the fix, the capacity passed to
the codec still included the consumed 16 bytes. For the final nonempty buffer,
the advertised destination interval therefore extended 16 bytes past the
allocation.

A codec can legally write within the interval supplied by its caller and
still overwrite adjacent heap memory. The corruption is detected later and
can appear as unrelated native failures, including a null source reported by
`checkCopyValue:fastCopy`, a fatal signal, or a JVM core dump. Hardware
watchpoint evidence identified a Zstandard `HUF_flushBits` store as the direct
write, but Zstandard remained within the incorrect interval supplied by Bolt;
the capacity calculation in Bolt is the root cause.

### Reproduction Steps

A deterministic unit reproduction does not need the original production SQL
or data:

1. Create one nonempty 64-byte Arrow buffer.
2. Use a recording codec whose `maxCompressedLen` returns 48.
3. Call `BlockPayload::fromBuffers` in compressed-buffer mode while tracking
   the output allocation.
4. Record the destination capacity passed to the codec and compare it with the
   bytes from the codec destination pointer to the allocation end.

Before the fix:

```text
advertised codec capacity = 64
actual bytes available    = 48
```

The 16-byte difference is the already-consumed compressed-buffer header. The
recording codec intentionally does not perform the out-of-bounds write, so the
test reproduces the invalid API contract safely and deterministically.

### Bolt Version / Commit ID

Affected base: `main` at `6ac9d88b19378373d073a92b77123fbd051577cb`.

### System Configuration

- **OS**: Not captured in the sanitized investigation report.
- **Compiler**: Not captured in the sanitized investigation report.
- **Build Type**: Release for the regression test and fixed production package.
- **CPU Arch**: Not captured in the sanitized investigation report.
- **Framework**: Spark-on-Bolt; the exact Spark version was not captured in the
  sanitized investigation report.

### Logs / Stack Trace

```shell
checkCopyValue:fastCopy source is null
SparkShuffleWriter
VeloxShuffleWriterV2::assembleBuffersGeneral
```

Other affected attempts could instead terminate with a native fatal signal or
JVM core dump after the earlier heap overwrite. These are downstream
detection points rather than the owner of the invalid destination capacity.

### Expected Behavior

After the output pointer advances past a compressed-buffer record header, the
capacity passed to the codec must describe only the bytes that remain in the
allocation. Compression must not be able to overwrite adjacent heap objects.

### Additional context

The proposed correction subtracts
`kCompressedBufferHeaderLength` from the remaining destination capacity. The
deterministic regression then observes 48 advertised bytes and 48 actually
available bytes.

Validation of the fixed package covered 29,882 Stage 2 task attempts across
ten concurrent applications. The original crash signatures remained absent:
zero `checkCopyValue:fastCopy` occurrences, native fatal signals, JVM fatal
errors, core dumps, problematic frames, or diagnostic overwrite hits across
9,530 watchpoint arms.

The same run had 29 task-attempt failures in two applications caused by
container memory-cgroup OOM; all were recovered by retry. Those resource
failures are separate from this native overwrite and are not evidence that
the compression-capacity bug remained.

zhangxffff:

Issue Number: close #817 

Spark-on-Bolt executors could crash during shuffle compression even when task
retries allowed the Spark application to finish as `Completed`.

`BlockPayload::fromBuffers` allocates a 16-byte record header for each
nonempty compressed buffer. `compressBuffer` advances the codec output pointer
past that header, but the caller previously still included the consumed 16
bytes in the destination capacity passed to the codec. For the final nonempty
buffer, the advertised destination interval therefore ended 16 bytes beyond
the real output allocation, allowing a legal codec store to corrupt adjacent
heap memory.

### Type of Change

- [ ] 📝 Documentation only

### Description

This PR contains only the production fix and its deterministic unit test:

- `bolt/shuffle/sparksql/Payload.cpp`
  - Moves the existing two-`int64_t` compressed-buffer header length to shared
    local scope.
  - Subtracts that header from the destination capacity after the output
    pointer advances past it.
- `bolt/shuffle/sparksql/tests/BoltShuffleReaderTest.cpp`
  - Adds `CompressedBufferCapacityExcludesHeader`.
  - Uses a tracking Arrow memory pool to measure the bytes actually available
    from the codec destination pointer to the allocation end.
  - Uses a recording fake codec to capture the advertised destination
    capacity without performing an out-of-bounds write.

Before the fix, the test observes:

```text
advertised codec capacity = 64
actual bytes available    = 48
```

After the fix, both values are 48.

#### Validation

- Built `bolt_shuffle_reader_test` with the Conan Release preset.
- `BoltShuffleReaderTest.CompressedBufferCapacityExcludesHeader`: 1/1 passed.
- Complete `bolt_shuffle_reader_test`: 5/5 passed.
- Clang-format 14.0.6 dry-run passed for both changed files.
- `git diff --check` passed.
- A fixed-package run covered 3,075 Stage 2 task attempts with zero failed
  attempts, 3,260 diagnostic watchpoint arms, zero overwrite hits, zero core
  dumps, and zero fatal native markers.
- A ten-application concurrent run covered 29,882 Stage 2 task attempts and
  9,530 diagnostic watchpoint arms. The original crash signatures remained
  absent: zero `checkCopyValue:fastCopy` occurrences, native fatal signals,
  JVM fatal errors, core dumps, problematic frames, or overwrite hits.
- The ten-application run did have 29 task-attempt failures in two
  applications caused by container memory-cgroup OOM; all were recovered by
  retry. Those resource failures are separate from the overwrite fixed here,
  so this validation does not claim that Stage 2 had zero errors.

### Performance Impact

    <details>

    ```text
    No benchmark was run.
    ```
    </details>

No material performance impact is expected. The compression critical path
adds one integer subtraction per nonempty buffer compression call and adds no
allocation, copy, branch, lock, or codec invocation. No benchmark was run.

### Release Note

```text
Release Note:
- Fixed a Spark shuffle compression buffer-capacity overstatement that could
  corrupt adjacent memory and crash executors.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    - None.
    ```
    </details>

### DIFF
--- a/bolt/shuffle/sparksql/Payload.cpp
+++ b/bolt/shuffle/sparksql/Payload.cpp
@@ -54,6 +54,7 @@ static const Payload::Mode kRowVectorMode = BlockPayload::kRowVector;
 static constexpr int64_t kZeroLengthBuffer = 0;
 static constexpr int64_t kNullBuffer = -1;
 static constexpr int64_t kUncompressedBuffer = -2;
+static constexpr int64_t kCompressedBufferHeaderLength = 2 * sizeof(int64_t);
 
 using RowSizeType = int32_t;
 static constexpr int32_t kRowSizeBytes = sizeof(RowSizeType);
@@ -133,7 +134,7 @@ arrow::Result<std::tuple<uint32_t, uint8_t>> readRowsAndMode(
 arrow::Result<int64_t> compressBuffer(
     const std::shared_ptr<arrow::Buffer>& buffer,
     uint8_t* output,
-    int64_t outputLength,
+    int64_t outputRemainingLength,
     Codec* codec) {
   auto outputPtr = &output;
   if (!buffer) {
@@ -144,11 +145,13 @@ arrow::Result<int64_t> compressBuffer(
     write<int64_t>(outputPtr, kZeroLengthBuffer);
     return sizeof(int64_t);
   }
-  static const int64_t kCompressedBufferHeaderLength = 2 * sizeof(int64_t);
   auto* compressedLengthPtr = advance<int64_t>(outputPtr);
   write(outputPtr, static_cast<int64_t>(buffer->size()));
-  auto compressedLength =
-      codec->compress(buffer->data(), buffer->size(), *outputPtr, outputLength);
+  auto compressedLength = codec->compress(
+      buffer->data(),
+      buffer->size(),
+      *outputPtr,
+      outputRemainingLength - kCompressedBufferHeaderLength);
   if (compressedLength >= buffer->size()) {
     // Write uncompressed buffer.
     memcpy(*outputPtr, buffer->data(), buffer->size());
@@ -184,11 +187,15 @@ arrow::Status compressAndFlush(
     ARROW_ASSIGN_OR_RAISE(
         compressed,
         arrow::AllocateResizableBuffer(
-            sizeof(int64_t) * 2 + maxCompressedLength, pool));
+            kCompressedBufferHeaderLength + maxCompressedLength, pool));
     auto output = compressed->mutable_data();
     ARROW_ASSIGN_OR_RAISE(
         compressedSize,
-        compressBuffer(buffer, output, maxCompressedLength, codec));
+        compressBuffer(
+            buffer,
+            output,
+            kCompressedBufferHeaderLength + maxCompressedLength,
+            codec));
   }
 
   {
--- a/bolt/shuffle/sparksql/tests/CMakeLists.txt
+++ b/bolt/shuffle/sparksql/tests/CMakeLists.txt
@@ -180,6 +180,21 @@ add_test(
     COMMAND bolt_shuffle_reader_test
 )
 
+add_executable(bolt_shuffle_payload_test PayloadTest.cpp)
+
+target_link_libraries(
+    bolt_shuffle_payload_test
+    PRIVATE
+        bolt_shuffle_spark_impl
+        bolt_testutils
+        GTest::gtest
+)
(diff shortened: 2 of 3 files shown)


bolt/shuffle/sparksql/tests/BoltShuffleReaderTest.cpp
```diff
@@ -196,6 +303,29 @@ TEST_F(BoltShuffleReaderTest, AllocationCountBounded) {
       << "saw " << allocs << " allocations; perf fix may have regressed";
 }
 
+TEST_F(BoltShuffleReaderTest, CompressedBufferCapacityExcludesHeader) {
+  TrackingMemoryPool trackingPool(pool_);
+  RecordingCodec codec;
```
fzhedu: why not use the real pool and codec of the repo?

fzhedu (approved):
LGTM

PR #823, opened 2026-08-03
fix: prevent crash casting dictionary/constant-wrapped complex type to VARCHAR

### LINKED ISSUE #822: [Bug] Crash casting dictionary/constant-wrapped complex type to VARCHAR
### Component Selection

### Describe the Bug

Casting a complex type (`ARRAY`/`MAP`/`ROW`) to `VARCHAR` crashes with `SIGSEGV`
when the complex child arrives wrapped in a dictionary or constant encoding.

`CastExpr::apply` only peels the **top-level** encoding before dispatching to
`doCast()`. When a complex column is nested inside another complex column (e.g.
`ARRAY<ROW<...>>`), the inner child can still be dictionary/constant-encoded when
`doCast()` recurses into it via the complex→VARCHAR path. In that state
`input.as<ArrayVector>()` (and the `MapVector`/`RowVector` equivalents) return
`nullptr`, and `doCast()` dereferenced that null pointer unconditionally,
crashing the process.

### Reproduction Steps

Minimal reproducible example (gtest), casting an outer `ARRAY` whose element
(`ROW`) child is dictionary-encoded, to `VARCHAR`:

1. Build an `ARRAY<ROW<BIGINT, VARCHAR>>` column whose `ROW` element child is
   wrapped in a dictionary (e.g. a reversing-index dictionary).
2. Evaluate `cast(c0 as VARCHAR)` on that column.
3. The engine crashes with `SIGSEGV` (null pointer dereference) inside
   `doCast()`.

The same crash reproduces for `ARRAY<MAP<...>>`, `ARRAY<ARRAY<...>>`, and for a
constant-encoded nested complex child.

### Bolt Version / Commit ID

main branch @ commit `2617ce877`

### System Configuration

- **OS**: Debian GNU/Linux 12
- **Compiler**: GCC 12
- **Build Type**: Release
- **CPU Arch**: x86_64 AVX2
- **Framework**: Spark 3.2 (spark_compatible build)

### Logs / Stack Trace

```shell
[ RUN      ] CastExprTest.complexTypeToStringWrappedNestedInput
*** Aborted at <ts> (Unix time) ***
*** Signal 11 (SIGSEGV) received by PID ... (code: address not mapped to object),
    stack trace: ***
    ... doCast() dereferencing input.as<ArrayVector>() == nullptr ...
```

### Expected Behavior

Casting a complex type to `VARCHAR` should correctly handle dictionary/constant
encoded (nested) complex inputs and produce the same result as the equivalent
flat input, instead of crashing.

fzhedu:

Casting a complex type (`ARRAY`/`MAP`/`ROW`) to `VARCHAR` crashed with `SIGSEGV`
when the complex child arrived wrapped in a dictionary/constant encoding.
`CastExpr::apply` only peels the top-level encoding, so a complex child nested
inside another complex column can still be dictionary/constant-encoded when
`doCast()` recurses into it. In that case `input.as<ArrayVector>()` (and the
`MapVector`/`RowVector` equivalents) return `nullptr`, and `doCast()`
dereferenced the null pointer unconditionally.

Issue Number: close #822

### Type of Change

- [ ] 📝 Documentation only

### Description

Add `doCastWrappedComplexToVarchar()` which handles the case where the complex
input to a `... -> VARCHAR` cast is dictionary/constant-encoded:

- **Why**: `input.as<ArrayVector>()` / `MapVector` / `RowVector` return `nullptr`
  for a wrapped (non-flat) complex vector, so the previous unconditional
  dereference crashed. This only surfaces for nested complex children, because
  top-level encodings are already peeled by `CastExpr::apply`.
- **How**: The new helper decodes the wrapped input down to its flat complex
  base vector, casts the base to `VARCHAR`, then copies the results back through
  the decoded indices (`decoded->indices()`) while re-applying the nulls
  introduced by the wrapper layers (`decoded->nulls()`). It validates the
  decoded base encoding with `BOLT_CHECK`/`BOLT_CHECK_NOT_NULL` and short-circuits
  when every selected row is null.

The dispatch in `doCast()` for `array/map/row -> VARCHAR` now checks for the
`nullptr` case and routes through the new helper, keeping the fast flat path
unchanged.

### Performance Impact

- [x] **No Impact**: This change does not affect the critical path. The new
  branch is only taken when the complex input is non-flat (dictionary/constant
  wrapped); the existing flat path is unchanged.

### Release Note

```text
Release Note:
- Fixed a SIGSEGV crash when casting a dictionary/constant-encoded complex type
  (ARRAY/MAP/ROW), typically nested inside another complex column, to VARCHAR.
```

### Checklist (For Author)

- [x] I have verified the code with local build (Release).

Verification detail: the two added tests
(`complexTypeToStringWrappedNestedInput`, `complexTypeToStringConstantNestedInput`)
crash with `SIGSEGV` on the pre-fix dispatch and pass with the fix; the full
`CastExprTest` suite (51 tests) passes.

### Breaking Changes

- [x] No

### DIFF
--- a/bolt/expression/CastExpr-tpl.cpp
+++ b/bolt/expression/CastExpr-tpl.cpp
@@ -1079,6 +1079,61 @@ void doCast(
     VectorPtr& result,
     CastErrorPolicy errorPolicy);
 
+// Casts a complex (ARRAY/MAP/ROW) vector to VARCHAR when the input arrives
+// wrapped in a dictionary/constant encoding, so `input.as<ArrayVector>()` (and
+// the map/row equivalents) return nullptr. Decodes the input down to its flat
+// complex base vector, casts the base, then copies the results back through the
+// decoded indices while re-applying nulls introduced by the wrapper layers.
+void doCastWrappedComplexToVarchar(
+    const SelectivityVector& rows,
+    const BaseVector& input,
+    exec::EvalCtx& context,
+    const TypePtr& fromType,
+    VectorPtr& result,
+    CastErrorPolicy errorPolicy) {
+  LocalDecodedVector decodedHolder(context, input, rows);
+  auto* decoded = decodedHolder.get();
+  auto* base = decoded->base();
+  BOLT_CHECK_NOT_NULL(
+      base, "Decoded vector for {} has no base vector.", fromType->toString());
+  BOLT_CHECK(
+      (fromType->isArray() && base->as<ArrayVector>() != nullptr) ||
+          (fromType->isMap() && base->as<MapVector>() != nullptr) ||
+          (fromType->isRow() && base->as<RowVector>() != nullptr),
+      "Decoded vector for {} has unexpected base encoding {}.",
+      fromType->toString(),
+      VectorEncoding::mapSimpleToName(base->encoding()));
+
+  LocalSelectivityVector nonNullRows(context, rows);
+  auto* rawNulls = decoded->nulls(nonNullRows.get());
+  if (rawNulls) {
+    nonNullRows->deselectNulls(
+        rawNulls, nonNullRows->begin(), nonNullRows->end());
+  }
+
+  context.ensureWritable(rows, VARCHAR(), result);
+  result->clearNulls(rows);
+
+  if (!nonNullRows->hasSelections()) {
+    result->addNulls(rows);
+    return;
+  }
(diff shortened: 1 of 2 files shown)


wangxinshuo-bolt (approved):
LGTM

PR #836, opened 2026-08-05
fix(row): validate CompactRow deserialization bounds and overflow

olemon111:
CompactRow deserialization trusted serialized lengths, array cardinalities, and
nested-element offsets before reading from the input row. Truncated, corrupted,
or schema-misaligned data could therefore read beyond a row boundary or treat
payload bytes as implausibly large sizes.

CompactRow serialization also accumulated some row sizes and nested offsets in
`int32_t`, or narrowed larger intermediate values without validation. Values
near the format's 2 GiB representation limit could overflow instead of failing
cleanly.

### Type of Change

- [ ] 📝 Documentation only

### Description

This PR:

- accumulates serialized sizes and offsets in `size_t` and validates every
  value that must fit in CompactRow's signed 32-bit wire fields;
- computes null-bitmap sizes without signed round-up overflow;
- validates row bitmaps, fixed-width fields, string lengths and payloads, array
  cardinalities, and minimum array layouts before reading them;
- rejects negative lengths, negative array sizes, cumulative array element-count
  overflow, and truncated payloads with contextual `BoltUserError` messages;
- validates complex-array payload sizes and element offsets, requires non-null
  offsets to be non-decreasing, and bounds each nested element by the next
  element offset so it cannot consume another element's bytes;
- rejects maps whose key and value arrays have different cardinalities;
- preserves the existing behavior that allows trailing bytes after a decoded
  top-level row;
- adds regression tests for the overflow boundary, malformed layouts, nested
  element isolation, map cardinality, error messages, and trailing-byte
  compatibility.

Valid CompactRow wire bytes and the public API are unchanged. Rows or nested
offsets larger than `INT32_MAX` now fail explicitly because they cannot be
represented by the existing format.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>
- [x] **Negative Impact**: Explained below (e.g., trade-off for correctness).

Valid-input deserialization now performs bounds checks before reading row
bitmaps, fixed-width values, string payloads, arrays, and nested offsets. Map
deserialization also compares key/value cardinalities. These checks add a small
CPU cost, most visibly for very narrow fixed-width rows.

The implementation avoids redundant checks: array layouts are validated once
at their common entry point, serialization uses `size_t` for unchecked local
accumulation and checks only at signed 32-bit boundaries, fixed-width array
checks are reused, and vectorized row serialization does not add another
per-row completion scan. Bounding a nested element uses `string_view::substr`
and does not copy payload data.

### Release Note

Release Note:

```text
Fix CompactRow serialization overflow and fail safely when deserializing
truncated or inconsistent row, string, array, nested-array, and map data.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/row/CompactRow.cpp
+++ b/bolt/row/CompactRow.cpp
@@ -29,11 +29,55 @@
  */
 
 #include "bolt/row/CompactRow.h"
+
+#include <limits>
+
 #include "bolt/vector/FlatVector.h"
 namespace bytedance::bolt::row {
 namespace {
 constexpr size_t kSizeBytes = sizeof(int32_t);
+constexpr size_t kMaxSerializedSize = std::numeric_limits<int32_t>::max();
 using TRowSize = uint32_t;
+
+size_t nullBytes(size_t size) {
+  return size / 8 + (size % 8 != 0);
+}
+
+int32_t checkedSerializedSize(size_t size) {
+  BOLT_USER_CHECK_LE(
+      size,
+      kMaxSerializedSize,
+      "CompactRow serialized value exceeds the maximum supported size of {} "
+      "bytes",
+      kMaxSerializedSize);
+  return static_cast<int32_t>(size);
+}
+
+void checkAvailable(
+    std::string_view data,
+    size_t row,
+    size_t offset,
+    size_t bytes,
+    std::string_view description) {
+  BOLT_CHECK(
+      offset <= data.size() && bytes <= data.size() - offset,
+      "Invalid CompactRow: {} needs {} bytes at offset {} in row {} with size "
+      "{}",
+      description,
+      bytes,
+      offset,
+      row,
+      data.size());
+}
+
+void checkAvailable(
+    const std::vector<std::string_view>& data,
+    size_t row,
+    size_t offset,
+    size_t bytes,
+    std::string_view description) {
+  checkAvailable(data[row], row, offset, bytes, description);
+}
 
 void writeInt32(char* buffer, int32_t n) {
   ::memcpy(buffer, &n, kSizeBytes);
@@ -246,7 +290,7 @@ void CompactRow::initialize(const TypePtr& type) {
         childIsFixedWidth_.push_back(child->type()->isFixedWidth());
       }
 
-      rowNullBytes_ = bits::nbytes(type->size());
+      rowNullBytes_ = nullBytes(type->size());
       break;
     }
     case TypeKind::BOOLEAN:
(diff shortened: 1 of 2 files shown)


bolt/row/CompactRow.cpp
```diff
@@ -305,7 +349,7 @@ std::optional<int32_t> fixedValueSize(const TypePtr& type) {
 // static
 std::optional<int32_t> CompactRow::fixedRowSize(const RowTypePtr& rowType) {
   const size_t numFields = rowType->size();
-  const size_t nullLength = bits::nbytes(numFields);
+  const size_t nullLength = nullBytes(numFields);
```
fzhedu, line 352: why change it ?

bolt/row/CompactRow.cpp
```diff
@@ -29,12 +29,56 @@
...
+constexpr size_t kMaxSerializedSize = std::numeric_limits<int32_t>::max();
 using TRowSize = uint32_t;
 
+size_t nullBytes(size_t size) {
+  return size / 8 + (size % 8 != 0);
+}
+
+int32_t checkedSerializedSize(size_t size) {
+  BOLT_USER_CHECK_LE(
+      size,
```
fzhedu, line 48: size > 0 or. size >=0?

bolt/row/CompactRow.cpp
olemon111: Replaced bits::nbytes(int32_t) with a local nullBytes(size_t) helper to avoid the intermediate (bits + 7) signed overflow and keep null bitmap size calculations in the size_t domain. This preserves identical results for valid inputs while making the calculation overflow-safe.

bolt/row/CompactRow.cpp
olemon111: size_t type is guaranteed to be non-negative, so there's no need to check

bolt/row/CompactRow.cpp
```diff
@@ -29,12 +29,56 @@
...
+  return static_cast<int32_t>(size);
+}
+
+void checkAvailable(
+    std::string_view data,
+    size_t row,
+    size_t offset,
+    size_t bytes,
+    std::string_view description) {
+  BOLT_USER_CHECK(
```
zhangxffff: `BOLT_USER_CHECK` does not seem appropriate here. If this check fails, the most likely cause is malformed internal state within Bolt rather than an error attributable to the user.

olemon111: Thanks. I changed the CompactRow deserialization integrity checks to BOLT_CHECK* and updated the tests to expect BoltRuntimeError. The serialization size-limit check remains a user error because it represents an unencodable input value.

zhangxffff (approved):
LGTM!

guhaiyan0221 (approved):

fzhedu (approved):
LGTM

PR #858, opened 2026-08-11
perf: skip clearing Parquet repdef lengths buffers

liuneng1994:
## Summary

- Remove the full-buffer `memset` before List/Map rep/def length conversion.
- Keep the existing `setSize(values_read * sizeof(int32_t))` boundary, since direct and fused conversion write every length in `[0, values_read)`.
- Add a poisoned-tail regression test covering direct and fused conversion. It verifies the unwritten tail is not modified and cannot be observed after the buffer is resized.
- Add clear-vs-no-clear microbenchmarks using the same stable-state visible buffer size as the reader.

## Performance

Release build, 64K lists, multiple runs:

| Shape | Direct | Fused |
| --- | ---: | ---: |
| Single element | 3.8%-4.5% faster | 2.5%-3.0% faster |
| Mixed | 0.3%-0.7% faster | 0.3%-0.5% faster |
| Long | 0.2%-0.5% faster | 0.3%-0.4% faster |

## Validation

- `ParquetReaderTest.repDefLengthsExposeOnlyConvertedValues`
- Full `bolt_dwio_parquet_reader_test`
- Direct/fused list-length conversion tests, including randomized fixed-seed and upper-bound cases
- `clang-format-14 --dry-run --Werror`
- `git diff --check`

### DIFF
--- a/bolt/dwio/parquet/arrow/tests/LevelConversionBenchmark.cpp
+++ b/bolt/dwio/parquet/arrow/tests/LevelConversionBenchmark.cpp
@@ -34,6 +34,7 @@
 #include <folly/init/Init.h>
 
 #include <algorithm>
+#include <cstring>
 #include <map>
 #include <memory>
 #include <random>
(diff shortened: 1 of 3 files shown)


guhaiyan0221 (approved):
LGTM!

PR #877, opened 2026-08-14
feat: add adaptive radix sort buffer

lizhen-0710:

Add a radix-first SortBuffer implementation for OrderBy behind order_by_radix_sort_enabled while preserving legacy SortBuffer selection. The new path normalizes sort keys into byte-comparable key records, stores only the payload needed to reconstruct output rows, writes the same compact key/payload format to radix spill files, and merges spill streams with the final in-memory run.

### Type of Change

- [ ] 📝 Documentation only

### Description

This change focuses on three optimization points:

1. Radix-first adaptive sorting. Detect natural runs, radix-sort unstable ranges over encoded key records, and fall back to key-record comparison when a prefix cannot fully decide order. This is faster for most tested data distributions, including random, duplicate, low-cardinality, null-heavy, wide-key, and wide-payload workloads.

2. Compact sort-state layout. Sort fixed-width key records instead of full rows, keep payload data behind key-record pointers, omit payload columns that can be reconstructed from keys, and avoid the extra sortedRows pointer array and general RowContainer padding. The compact layout is also the on-disk radix spill row format, so spill can serialize the sort state directly.

3. Smarter spill and faster merge. Spill current runs only when memory pressure requires it, keep the final in-memory run as a merge stream instead of forcing it to disk in noMoreInput, and re-spill only the not-yet-output suffix during output-stage reclaim. Reuse radix spill streams, raw key/payload write paths, and specialized merge paths to reduce spill merge overhead.

Add benchmarks and coverage for key encoding, payload row layout and IO, run sorting/storage, spill row IO, radix buffer output, output-stage reclaim, OrderBy integration, and ContainerRowSerde fixed-key spill reads.

### Performance Impact

Benchmark results:

Benchmarks were run with the Release benchmark binaries. Each value is the median time/iter across three independent process runs; lower is better.

- Sort algorithm only: 14 scenarios, 3-run median time/iter.
- SortBuffer E2E in-memory: 19 scenarios, 3-run median time/iter.
- SortBuffer with spill, no compression: 15 scenarios, 3-run median time/iter.
- SortBuffer with spill, LZ4: 15 scenarios, 3-run median time/iter.

Sort algorithm only:

| scenario | legacy ms | radix ms | speedup |
| --- | ---: | ---: | ---: |
| random_w1_10k | 0.78 | 0.37 | 2.12x |
| random_w1_100k | 10.18 | 3.11 | 3.27x |
| random_w1_1m | 131.15 | 40.91 | 3.21x |
| random_w1_10m | 1978.80 | 416.59 | 4.75x |
| low_cardinality_w1_1m | 44.62 | 8.05 | 5.54x |
| random_w2_1m | 140.89 | 49.43 | 2.85x |
| random_w4_1m | 149.50 | 58.17 | 2.57x |
| random_w8_1m | 172.01 | 72.30 | 2.38x |
| low_cardinality_w4_1m | 190.19 | 162.48 | 1.17x |
| random_w2_10m | 2246.62 | 646.58 | 3.47x |
| random_w4_10m | 2390.73 | 776.35 | 3.08x |
| low_cardinality_w4_10m | 3093.19 | 2150.47 | 1.44x |
| random_w8_10m | 2518.29 | 741.33 | 3.40x |
| low_cardinality_w8_10m | 3248.38 | 2632.11 | 1.23x |

SortBuffer E2E in-memory:

| scenario | legacy ms | radix ms | speedup |
| --- | ---: | ---: | ---: |
| random_i64_narrow_256k | 43.28 | 18.65 | 2.32x |
| duplicate_i64_narrow_256k | 19.88 | 12.01 | 1.65x |
| low_cardinality_i64_256k | 9.89 | 8.93 | 1.11x |
| null_heavy_i64_128k | 14.91 | 7.92 | 1.88x |
| eight_key_i64_128k | 27.24 | 22.10 | 1.23x |
| inline_varchar_128k | 25.41 | 17.80 | 1.43x |
| long_varchar_64k | 40.74 | 26.72 | 1.52x |
| varchar_common_prefix_128k | 90.03 | 65.37 | 1.38x |
| nullable_fixed_payload_128k | 20.62 | 9.80 | 2.10x |
| wide_fixed_payload_128k | 33.37 | 18.79 | 1.78x |
| very_wide_fixed_payload_64k | 49.24 | 35.58 | 1.38x |
| wide_string_payload_64k | 83.59 | 35.48 | 2.36x |
| random_i64_narrow_1m | 201.31 | 80.69 | 2.49x |
| low_cardinality_i64_1m | 48.69 | 40.69 | 1.20x |
| eight_key_i64_1m | 341.66 | 278.35 | 1.23x | 
| sixteen_key_i64_1m | 468.67 | 401.01 | 1.17x | 
| wide_fixed_payload_1m | 488.35 | 217.36 | 2.25x | 
| very_wide_fixed_payload_1m | 2020.26 | 666.05 | 3.03x |
| wide_string_payload_1m | 2066.06 | 737.32 | 2.80x |

SortBuffer with spill, no compression:

| scenario | legacy ms | radix ms | speedup |
| --- | ---: | ---: | ---: |
| random_i64_256k_spill | 68.20 | 25.19 | 2.71x |
| duplicate_i64_256k_spill | 48.03 | 17.96 | 2.67x |
| null_heavy_i64_128k_spill | 26.95 | 11.27 | 2.39x |
| eight_key_i64_128k_spill | 48.31 | 38.50 | 1.25x |
| inline_varchar_128k_spill | 40.00 | 25.34 | 1.58x |
| long_varchar_64k_spill | 59.51 | 37.34 | 1.59x |
| varchar_common_prefix_128k_spill | 111.17 | 80.83 | 1.38x |
| wide_fixed_payload_128k_spill | 39.65 | 16.27 | 2.44x |
| very_wide_fixed_payload_64k_spill | 89.71 | 56.03 | 1.60x |
| wide_string_payload_64k_spill | 58.62 | 37.03 | 1.58x |
| random_i64_1m_spill | 305.28 | 110.27 | 2.77x |
| eight_key_i64_1m_spill | 536.80 | 410.12 | 1.31x |
| sixteen_key_i64_1m_spill | 709.52 | 551.27 | 1.29x |
| wide_fixed_payload_1m_spill | 395.23 | 159.86 | 2.47x |
| wide_string_payload_1m_spill | 1366.80 | 671.70 | 2.03x |

SortBuffer with spill, LZ4:

| scenario | legacy ms | radix ms | speedup |
| --- | ---: | ---: | ---: |
| random_i64_256k_spill | 77.55 | 31.18 | 2.49x |
| duplicate_i64_256k_spill | 51.48 | 19.11 | 2.69x |
| null_heavy_i64_128k_spill | 30.74 | 13.55 | 2.27x |
| eight_key_i64_128k_spill | 70.04 | 52.94 | 1.32x |
| inline_varchar_128k_spill | 44.33 | 27.34 | 1.62x |
| long_varchar_64k_spill | 60.44 | 37.16 | 1.63x |
| varchar_common_prefix_128k_spill | 119.90 | 82.27 | 1.46x |
| wide_fixed_payload_128k_spill | 54.84 | 24.69 | 2.22x |
| very_wide_fixed_payload_64k_spill | 122.10 | 71.62 | 1.70x |
| wide_string_payload_64k_spill | 48.22 | 29.70 | 1.62x |
| random_i64_1m_spill | 341.40 | 129.60 | 2.63x |
| eight_key_i64_1m_spill | 703.13 | 540.53 | 1.30x |
| sixteen_key_i64_1m_spill | 1019.06 | 778.30 | 1.31x |
| wide_fixed_payload_1m_spill | 520.75 | 219.62 | 2.37x |
| wide_string_payload_1m_spill | 1056.57 | 612.69 | 1.72x |

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>


bolt/core/QueryConfig.h
```diff
@@ -1135,6 +1138,10 @@ class QueryConfig {
     return get<bool>(kOrderBySpillEnabled, true);
   }
 
+  bool orderByRadixSortEnabled() const {
+    return get<bool>(kOrderByRadixSortEnabled, true);
```
lizhen-0710, line 1142: Should be disabled by default, when check in.

bolt/exec/radixsort/RadixSortRun.cpp
```diff
@@ -0,0 +1,883 @@
...
+  std::vector<RadixSortOutputColumn> columns(outputType->size());
+  std::vector<column_index_t> payloadChannels;
+  std::vector<std::string> payloadNames;
+  std::vector<TypePtr> payloadTypes;
+  payloadChannels.reserve(outputType->size());
+  payloadNames.reserve(outputType->size());
+  payloadTypes.reserve(outputType->size());
+  for (uint32_t output = 0; output < outputType->size(); ++output) {
+    const auto decodedKey = decodedKeyByOutput[output];
+    if (decodedKeyByOutput[output] >= 0 && directOccurrences[output] == 1) {
```
guhaiyan0221, line 136: > This projects every uniquely referenced sort key from the decoded radix key, but the floating-point encoding is not lossless: encodeFloat/encodeDouble collapse +0.0 and -0.0 to the same key and canonicalize
  > all NaN payloads. Decoding therefore changes the actual value returned by OrderBy (for example, -0.0 becomes +0.0, so a downstream 1.0 / c0 changes from -inf to +inf). The same issue applies to floating-
  > point values nested in complex sort keys. Please retain any key type that is not losslessly round-trippable in the payload, or otherwise preserve its original representation. A regression test should compare
  > the output IEEE bits for signed zero and distinct NaN payloads in both in-memory and spill paths; the current comparator-based assertions consider these values equivalent and therefore miss the mutation.

bolt/exec/SpillFile.cpp
```diff
@@ -518,6 +691,73 @@ uint64_t SpillWriter::write(
...
+  BOLT_CHECK_LE(bodySize, maxUncompressedSpillBlockSize(compressionKind_));
+
+  auto* file = ensureFile();
+  BOLT_CHECK_NOT_NULL(file);
+
+  const char* writeBuffer = frame;
+  uint64_t writeSize = headerSize + static_cast<uint64_t>(bodySize);
+  uint64_t compressTimeUs = 0;
+  int32_t storedSize = bodySize;
+  if (isSpillCompressionEnabled(compressionKind_)) {
```
guhaiyan0221: > Do not silently treat supported spill codecs as uncompressed
  >
  > spill_compression_codec accepts codecs such as ZLIB, SNAPPY, and GZIP, but the radix spill path only recognizes LZ4 and ZSTD and silently writes raw blocks for every other value. Although the matching reader
  > can still read these files, enabling radix sort changes the configured spill behavior and can substantially increase disk usage or cause the query to hit max_spill_bytes unexpectedly. Please either support
  > the configured codec through the generic codec interface or reject unsupported codecs explicitly instead of silently treating them as NONE. The current "zlib" OrderBy test only checks correctness and non-
  > zero spill stats, so it does not detect that no compression occurred.

bolt/exec/radixsort/RadixSortRun.cpp
lizhen-0710: Thanks for raising this concern. We reviewed the floating-point normalization behavior against DuckDB’s sort implementation and Spark’s window/join key handling, where canonicalizing floating-point values for key processing is acceptable.
To avoid changing the exact IEEE representation returned by OrderBy, we added kOrderByRadixSortFallbackForFloatingPointKeysEnabled. It is enabled by default: when a sort key is REAL or DOUBLE, or recursively contains either type inside ARRAY, MAP, or ROW, OrderBy falls back to the legacy sort buffer. Users can explicitly disable this guard to opt into radix sort for these keys.
We sampled our workloads and found that floating-point sort keys account for only a very small fraction of both query count and execution time, so this correctness-first fallback has limited performance impact.
We initially tried retaining a separate copy of the original key values and restoring them after radix sorting to preserve NaN payloads and signed-zero representations. That approach added extra storage and data movement and did not perform well, so we chose the explicit fallback instead.
The tests now cover scalar and nested floating-point keys, NaN payloads, +0.0/-0.0, in-memory and spill execution, both config states, and confirm that floating-point payload columns do not trigger fallback and retain their original IEEE representations on the radix path.

bolt/exec/SpillFile.cpp
lizhen-0710: Thanks for pointing this out. The original radix spill implementation followed the row-format spill behavior and only handled LZ4 and ZSTD in its block compression path. As a result, configuring ZLIB, SNAPPY, or GZIP could leave radix spill blocks uncompressed.
We have updated the radix encoded-block path to support all applicable Bolt spill codecs: ZLIB, SNAPPY, LZ4, ZSTD, and GZIP. NONE remains an explicit uncompressed path, and unsupported LZO is rejected instead of silently falling back to raw data.
The implementation uses caller-provided reusable buffers for all five codecs and preserves the existing native LZ4 and ZSTD fast paths. The legacy row-format spill path and its codec-selection behavior are unchanged.
We also added tests that verify, for every supported codec:
- the on-disk body is actually compressed;
- the bytes can be independently decoded using Bolt’s common codec implementation;
- the radix reader decompresses them correctly;
- keys and payloads round-trip through the radix spill path;
- the complete OrderBy spill path produces the expected result.

bolt/exec/radixsort/RadixSortKeyCodec.cpp
```diff
@@ -0,0 +1,5348 @@
...
+  while (cursor < remaining) {
+    auto byte = static_cast<uint8_t>(body[cursor++]);
+    if (descending) {
+      byte = static_cast<uint8_t>(~byte);
+    }
+    if (byte == kStringDelimiter) {
+      encodedSize = cursor;
+      return;
+    }
+    if (byte == kBlobEscape) {
```
guhaiyan0221, line 4367: After encountering kBlobEscape, scanStringBody unconditionally reads the next byte through body[cursor++]. If the encoded input is truncated and ends with a standalone escape byte, cursor is already equal to remaining, causing an out-of-bounds read past the EncodedKeyView. Please check the remaining length before reading the escaped byte and report an explicit error for an incomplete escape sequence.

guhaiyan0221 (approved):
LGTM! Thanks for radix sort support and performance improvement.

PR #886, opened 2026-08-18
feat(parquet): allow disabling implicit casts of parquet reader in Spark

kexianda:

Issue Number: close #xxx

### Type of Change

- [ ] 📝 Documentation only

### Description

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
```

### Checklist (For Author)

### Breaking Changes

- [ ] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    ```
    </details>

### DIFF
--- a/bolt/core/QueryConfig.h
+++ b/bolt/core/QueryConfig.h
@@ -623,6 +623,9 @@ class QueryConfig {
   static constexpr const char* kParquetRepDefMemoryLimit =
       "parquet_repdef_memory_limit";
 
+  static constexpr const char* kParquetReaderImplicitCastMask =
+      "parquet_reader_implicit_cast_mask";
+
   static constexpr const char* kHybridJoinEnabled = "hybrid_join_enabled";
 
   /// If true, reorder rows by containerId during hybrid join extraction for
@@ -1720,6 +1723,10 @@ class QueryConfig {
 
   int32_t parquetRepDefMemoryLimit() const {
     return get<int32_t>(kParquetRepDefMemoryLimit, 128UL << 20);
+  }
+
+  int64_t parquetReaderImplicitCastMask() const {
+    return get<int64_t>(kParquetReaderImplicitCastMask, 0);
   }
 
   bool enableDynamicConcurrencyAdjustment() const {
--- a/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
+++ b/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
@@ -110,6 +110,46 @@ bool matchType(TypeKind schemaType, TypeKind requestType) {
   return true;
 }
 
+bool isIntegerType(const Type& type) {
+  return type.equivalent(*TINYINT()) || type.equivalent(*SMALLINT()) ||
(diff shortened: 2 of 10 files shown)


bolt/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -93,7 +119,17 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
       fileType->childByName("value")->type()->isVarbinary() &&
       fileType->childByName("metadata")->type()->isVarbinary();
 
-#ifndef SPARK_COMPATIBLE
+#ifdef SPARK_COMPATIBLE
+  BOLT_CHECK(
+      canReadVariantStructAsVariant ||
+          !maskSparkImplicitCast(
```
guhaiyan0221: kAll is checked against the root ROW type, so it treats column projection as an implicit cast. Hive builds an empty ROW selector for count(*); comparing it with the file’s non-empty root schema is not equivalent and therefore throws ROW can not convert to ROW. No cast is involved here. Could we apply this mask only to leaf types that actually use reader-side casts, and add coverage for count(*), projected schemas, and missing-column null filling?

kexianda: fixed

bolt/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -77,6 +77,36 @@ bool matchType(TypeKind schemaType, TypeKind requestType) {
...
+    return false;
+  }
+  if (castMask == static_cast<int64_t>(CastMask::kAll)) {
+    return true;
+  }
+  const auto varcharIntegerMask =
+      static_cast<int64_t>(CastMask::kVarcharInteger);
+  const auto fileKind = fileType.kind();
+  const auto requestedKind = requestedType.kind();
+  return (castMask & varcharIntegerMask) != 0 &&
```
guhaiyan0221: DateType also uses TypeKind::INTEGER, so this condition makes kVarcharInteger block the supported DATE -> VARCHAR reader cast as well. ParquetTableScanTest explicitly treats this conversion as valid. Could we classify the logical type here instead of relying only on TypeKind, and add a regression test for DATE -> VARCHAR with this mask?

bolt/core/QueryConfig.h
```diff
@@ -1715,6 +1718,10 @@ class QueryConfig {
     return get<int32_t>(kParquetRepDefMemoryLimit, 128UL << 20);
   }
 
+  int64_t parquetReaderImplicitCastMask() const {
+    return get<int64_t>(kParquetReaderImplicitCastMask, 0);
```
guhaiyan0221, line 1744: it is better to disable VARCHAR ↔ INTEGER implicit cast by default

kexianda: default behavior is compatible with Spark

guhaiyan0221 (approved):

PR #887, opened 2026-08-18
perf(arrow): add reusable Arrow batch pool

liuneng1994:

Repeated Arrow C Data Interface exports rebuild the ArrowArray holder tree,
offset buffers, and ArrowSchema for every batch. This creates avoidable
allocation and release overhead for callers that repeatedly export compatible
Bolt vectors.

Issue Number: N/A

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Performance improvement (optimization)
- [ ] Documentation only

### Description

Add `ReusableArrowBatchPool`, a bounded pool for exporting `ArrowArray` and
`ArrowSchema` together.

- Reuse Arrow array holder trees through a flat release registry, avoiding
  recursive release work between batches.
- Reuse offsets scratch buffers while capping retained scratch memory at 32
  KiB, plus the trailing Arrow offset.
- Cache exported schemas and invalidate the cache when the vector type,
  encoding shape, field names, or relevant Arrow options change.
- Lease each pool slot until the consumer invokes the Arrow release callback.
- Fall back to the regular export path when all slots are leased or the pool
  has zero slots.
- Keep the low-level reusable-array lifecycle internal to `Bridge.cpp`; the
  public API continues to use standard `ArrowArray` and `ArrowSchema` structs.

Tests cover offsets reuse and eviction, nested and constant vectors, export
failure recovery, vector shape changes, concurrent slot leases, fallback
exports, schema cache hits and invalidation, and zero-slot behavior.

The Arrow export microbenchmark now compares full schema and array export plus
consumer release for regular and reusable paths. It also fixes the existing
array-only cases to honor Folly iteration counts and release exported arrays.

### Performance Impact

    <details>

    ```text
    Release build, pinned to CPU 0.
    Seven paired rounds per case, alternating regular/reusable execution order.
    Each process measured the selected case for at least 500 ms.

    Case                         Regular median   Reusable median   Speedup
    8 columns x 128 rows              6.753 us          3.586 us      1.88x
    8 columns x 8192 rows            299.01 us         293.76 us      1.02x
    Nested 4 columns x 8192 rows     274.96 us         259.82 us      1.06x
    1700 columns x 8192 rows           1.283 ms         425.58 us      3.01x

    Exact paired sign-flip test on per-round log ratios:
    - 128-row follow-up: -46.94%, paired p=0.015625
    - 8192-row 8-column:  -1.90%, Holm p=0.0469, below 3% effect threshold
    - 8192-row nested:     -5.63%, Holm p=0.0469
    - 8192-row 1700-col:  -66.64%, Holm p=0.0469
    ```

    The 128-row case was measured as a separate seven-round follow-up. The
    three original 8192-row cases were corrected together using Holm's method.
    The 1700-column case shares child data vectors to isolate the per-column
    Arrow metadata allocation, traversal, and release costs targeted by this
    change.
    </details>

The reusable path is opt-in. Existing callers continue to use the unchanged
export path, and pool exhaustion falls back to that path.

### Release Note

```text
Release Note:
- Added a bounded Arrow batch export pool that reuses array metadata, offsets
  scratch buffers, and schemas across compatible exports.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/vector/arrow/Bridge.cpp
+++ b/bolt/vector/arrow/Bridge.cpp
@@ -73,18 +73,10 @@ void appendArrowMetadataInt32(std::string& metadata, int32_t value) {
 class BoltToArrowBridgeHolder {
  public:
   struct ReusableReleaseState {
-    struct Entry {
-      ArrowArray* array;
-      BoltToArrowBridgeHolder* holder;
-    };
-
-    // Child nodes stay at stable addresses, but consumers may move the root
-    // ArrowArray before releasing it. Keep the stable tree here for flat reset.
-    std::vector<Entry> entries;
-
-    void add(ArrowArray& array, BoltToArrowBridgeHolder& holder) {
-      entries.push_back({&array, &holder});
-    }
+    // Child nodes stay at stable addresses, but consumers may move an
+    // ArrowArray before releasing it. Keep the stable tree here for reuse.
+    std::atomic_size_t activeCount{0};
+    std::atomic_bool inUse{false};
   };
 
   explicit BoltToArrowBridgeHolder(bool reusable = false)
@@ -120,17 +112,41 @@ class BoltToArrowBridgeHolder {
     return ownsReleaseState_;
   }
 
-  void markActive(ArrowArray& array) {
+  bool tryAcquireReusableSlot() {
+    BOLT_CHECK(reusable_);
+    BOLT_CHECK(ownsReleaseState_);
+    bool expected = false;
+    return releaseState_->inUse.compare_exchange_strong(
+        expected, true, std::memory_order_acq_rel, std::memory_order_acquire);
+  }
+
+  size_t reusableActiveCount() const {
+    return releaseState_ == nullptr
+        ? 0
+        : releaseState_->activeCount.load(std::memory_order_acquire);
+  }
+
+  void registerArray(ArrowArray& array) {
     if (!releaseState_) {
       return;
     }
     array.private_data = this;
     if (registeredArray_ == nullptr) {
       registeredArray_ = &array;
-      releaseState_->add(array, *this);
       return;
     }
     BOLT_CHECK(registeredArray_ == &array);
+  }
+
+  void markActive(ArrowArray& array) {
+    if (!releaseState_) {
+      return;
+    }
+    registerArray(array);
+    if (!active_) {
+      active_ = true;
+      releaseState_->activeCount.fetch_add(1, std::memory_order_relaxed);
+    }
   }
 
   void reset() {
@@ -189,27 +205,34 @@ class BoltToArrowBridgeHolder {
       return;
     }
     BOLT_CHECK(ownsReleaseState_);
-    bool registeredArray = false;
-    for (auto [entryArray, holder] : releaseState_->entries) {
-      registeredArray |= entryArray == &releasedArray;
-      holder->resetLocalBuffers();
-      resetArrowArray(*entryArray, holder->getArrowBuffers());
-    }
-    if (!registeredArray) {
-      resetArrowArray(releasedArray, nullptr);
-      releasedArray.private_data = nullptr;
-    }
+    resetReusableRecursive(releasedArray);
   }
 
   void resetReusableChild(ArrowArray& releasedArray) {
     BOLT_CHECK(reusable_);
     BOLT_CHECK(!ownsReleaseState_);
-    resetLocalBuffers();
-    resetArrowArray(releasedArray, getArrowBuffers());
+    resetReusableRecursive(releasedArray);
+  }
+
+  void abortReusableArray(ArrowArray& releasedArray) {
+    if (!releaseState_) {
+      return;
+    }
+    resetRecursive();
+    active_ = false;
+    releaseState_->activeCount.store(0, std::memory_order_release);
+    if (registeredArray_ != &releasedArray) {
+      resetArrowArray(releasedArray, nullptr);
+      releasedArray.private_data = nullptr;
+    }
   }
 
   void releaseResources() {
     resetRecursive();
+    if (releaseState_) {
+      releaseState_->activeCount.store(0, std::memory_order_release);
+      releaseState_->inUse.store(false, std::memory_order_release);
+    }
     for (auto& child : childrenPtrs_) {
       if (!child) {
         continue;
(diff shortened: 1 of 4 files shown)


bolt/vector/arrow/Bridge.cpp
```diff
@@ -68,14 +72,178 @@ void appendArrowMetadataInt32(std::string& metadata, int32_t value) {
...
+
+  void resetReusableArray(ArrowArray& releasedArray) {
+    if (!releaseState_) {
+      return;
+    }
+    BOLT_CHECK(ownsReleaseState_);
+    bool registeredArray = false;
+    for (auto [entryArray, holder] : releaseState_->entries) {
+      registeredArray |= entryArray == &releasedArray;
+      holder->resetLocalBuffers();
```
guhaiyan0221: > The Arrow C Data Interface allows moving a child array and then immediately releasing the parent. This loop resets every registered holder, including a moved child whose source release is now null, so the
  > moved child loses its buffers and ownership. The slot also needs to remain leased until all moved children are released.

bolt/vector/arrow/Bridge.cpp
```diff
@@ -110,19 +278,60 @@ class BoltToArrowBridgeHolder {
...
+    if (offsetsScratch_ && offsetsScratch_->pool() == pool &&
+        offsetsScratch_->isMutable()) {
+      if (offsetsScratch_->capacity() < bytes) {
+        AlignedBuffer::reallocate<uint8_t>(&offsetsScratch_, bytes);
+      } else {
+        offsetsScratch_->setSize(bytes);
+      }
+      return offsetsScratch_;
+    }
+    offsetsScratch_ = AlignedBuffer::allocate<uint8_t>(bytes, pool);
```
guhaiyan0221, line 317: > Could we avoid retaining this buffer after the exported array is released? A caller can release both output structs and then destroy the supplied MemoryPool, but the reusable slot still owns a BufferPtr
  > allocated from it. This currently triggers the pool leak check and may later free through a dead pool.

bolt/vector/arrow/Bridge.cpp
```diff
@@ -3125,4 +3490,359 @@ VectorPtr importFromArrowAsOwner(
...
+  std::pair<std::shared_ptr<CachedSchema>, bool> get(
+      const VectorPtr& vector,
+      const ArrowOptions& options,
+      const std::vector<std::string>& fieldNames,
+      memory::MemoryPool* pool) {
+    const auto signature = makeSchemaSignature(vector, options, fieldNames);
+    std::lock_guard<std::mutex> lock(mutex_);
+    // The signature covers encoding and options; equivalent() validates the
+    // logical type and protects against hash collisions.
+    if (schema_ != nullptr && signature_.has_value() &&
```
guhaiyan0221, line 3733: Could the cache key or validation include RowType field names when fieldNames is empty? RowType::equivalent() ignores names, so ROW<first: BIGINT> and ROW<second: BIGINT> reuse the same cached schema and the second export still returns "first".

bolt/vector/arrow/Bridge.cpp
```diff
@@ -110,19 +278,60 @@ class BoltToArrowBridgeHolder {
...
+    return offsetsScratch_;
+  }
+
   // Allocates space for `numChildren` ArrowArray pointers.
   void resizeChildren(size_t numChildren) {
-    childrenPtrs_.resize(numChildren);
+    const auto oldSize = childrenPtrs_.size();
+    // Reusable nodes only grow: registered addresses must remain stable. The
+    // current batch shape is represented by ArrowArray::n_children.
+    if (!reusable_ || oldSize < numChildren) {
```
guhaiyan0221, line 326: The reusable tree only grows, while every root release scans all registered entries. After one wide batch, later narrow batches on the same slot retain and scan the historical maximum tree. Could we prune inactive nodes or benchmark a wide-to-narrow workload on one pool?

bolt/vector/arrow/Bridge.cpp
liuneng1994: Fixed in 717b5474da. The slot stays leased until moved children are released, with a new test covering it.

bolt/vector/arrow/Bridge.cpp
liuneng1994: Reusing the offsets buffer improves the hot path. It is now capped at 32 KiB, and the API documents that the MemoryPool must outlive the reusable pool.

bolt/vector/arrow/Bridge.cpp
liuneng1994: Fixed in 717b5474da. Schema cache validation uses exact Type equality, and a row-name regression test was added.

bolt/vector/arrow/Bridge.cpp
liuneng1994: The intended usage is one reusable pool per Velox pipeline, so shape history is bounded by that pipeline.

bolt/vector/arrow/Bridge.cpp
```diff
@@ -3125,4 +3554,353 @@ VectorPtr importFromArrowAsOwner(
...
+  bool exportToArrow(
+      const VectorPtr& vector,
+      memory::MemoryPool* pool,
+      const ArrowOptions& options,
+      ArrowSchema* schema,
+      ArrowArray* array,
+      const std::vector<std::string>& fieldNames) {
+    BOLT_CHECK_NOT_NULL(schema);
+    BOLT_CHECK_NOT_NULL(array);
+    BOLT_CHECK_NULL(schema->release);
```
guhaiyan0221: This reads `schema->release` before initializing the caller-provided output struct. Existing Bolt export APIs allow callers to pass `ArrowSchema schema;` / `ArrowArray array;` without value-initializing them,
  so `release` may contain an indeterminate value here. Reading it is undefined behavior and can randomly fail this check. Please remove these checks and overwrite the outputs directly.

liuneng1994: This check is necessary; using this approach requires ensuring that `schema->release` is `null`.

liuneng1994: done

guhaiyan0221 (approved):

PR #897, opened 2026-08-19
perf(agg): DictionaryVector fast path for numeric aggregates

### LINKED ISSUE #896: [Performance] Aggregate functions: add DictionaryVector fast-path for numeric inputs
### Performance Bottleneck Type

CPU Bound (High CPU usage, slow calculation)

### Environment Details (Critical)

- **CPU Model**: generic x86_64 (any target)
- **SIMD Support**: AVX2 / AVX-512
- **RAM**: N/A (not memory bound)
- **Storage**: N/A (in-memory Vector path)
- **OS & Kernel**: Linux
- **Bolt Build Type**: Release

### Workload Description

Numeric aggregates (SUM / AVG / MIN / MAX / bitwise, etc.) whose input column
is a `DictionaryVector`, e.g. the output of a Filter, HashJoin build side, or
a low-cardinality column read from Parquet dict-encoded pages.

```
Query pattern:
  SELECT sum(v), avg(v), min(v), max(v)
  FROM t
  WHERE ...             -- produces dictionary-wrapped output
  GROUP BY k;

Data:  numeric column wrapped by DictionaryVector<int/float/double/...>.
Size:  applies to any scale; benefit grows with row count.
```

### Performance Comparison

| System | Time / Latency | Throughput (Rows/s) | CPU Usage |
| :--- | :--- | :--- | :--- |
| **Bolt (current, DecodedVector path)** | baseline | baseline | 100% |
| **Bolt (proposed dictionary fast path)** | lower | higher | lower |

The current base classes always call `DecodedVector::decode()` even when the
input is already a fully usable `DictionaryVector`, incurring an extra
allocation + populate step per batch.

### Reproduction Steps

1. Build Bolt in Release.
2. Any TPC-H / TPC-DS query where a numeric column arrives at the aggregation
   as `DictionaryVector` (very common after a Filter or upstream dict scan).
3. The new unit tests in `AverageAggregationTest` and `SumTest`
   (see PR) also exercise both `addRawInput` and `addSingleGroupRawInput`
   dictionary paths.

### Profiling Data / Metrics

Root cause (from source inspection):

- [`SimpleNumericAggregate::updateGroups`](https://github.com/bytedance/bolt/blob/main/bolt/functions/lib/aggregates/SimpleNumericAggregate.h) and `updateOneGroup`
- [`AverageAggregateBase::addRawInput`](https://github.com/bytedance/bolt/blob/main/bolt/functions/lib/aggregates/AverageAggregateBase.h) and `addSingleGroupRawInput`

both fall through into a `DecodedVector::decode(*arg, rows)` path
regardless of the input encoding.

For `DICTIONARY` encoding we can iterate through `valueAtFast(i)` /
`isNullAt(i)` directly on the `DictionaryVector` and skip the
`DecodedVector` materialization entirely.

### Additional context

Fix will be sent as a separate PR (linked to this issue). Guarded by
`DictionaryVector::loadedVector()` so it is safe when the dictionary values
underneath are still a lazy vector.

fzhedu:

Numeric aggregates (SUM / AVG / MIN / MAX / bitwise / etc.) whose input is
already a `DictionaryVector` currently go through `DecodedVector::decode()`
and materialize an intermediate `DecodedVector`, even though the values can
be read directly through `valueAtFast()`. This PR adds a dedicated fast path.

Issue Number: close #896

### Type of Change

### Description

Two shared base classes gain a dictionary fast path:

- `SimpleNumericAggregate::updateGroups` / `updateOneGroup`
- `AverageAggregateBase::addRawInput` / `addSingleGroupRawInput`

When the input encoding is `DICTIONARY`, we downcast to
`DictionaryVector<T>` and call `loadedVector()` once, then iterate through
`valueAtFast(i)` / `isNullAt(i)` directly — skipping the `DecodedVector`
allocation and population.

**Why the `loadedVector()` guard.**
`DictionaryVector::setInternalState()` returns early **without** setting
`initialized_ = true` when its `dictionaryValues_` is still an unloaded
`LazyVector`. Touching `mayHaveNulls()` / `valueAtFast(i)` in that state
either trips `BOLT_DCHECK(initialized_)` in debug builds or reads
uninitialized cache pointers in release. `loadedVector()` normalizes the
vector before the fast path executes; it is a no-op when the state is
already initialized.

**Tests.** New unit tests in `AverageAggregationTest` and `SumTest` cover
dictionary-wrapped inputs for both the multi-group and single-group paths.

### Performance Impact

- [x] **Positive Impact**: eliminates the `DecodedVector` allocation +
  population per batch when the input is already dictionary-encoded. Cost
  on the non-dictionary path is unchanged (a single encoding check).

### Release Note

```text
Release Note:
- perf(agg): add a DictionaryVector fast path for numeric aggregates
  (SUM/AVG/MIN/MAX/bitwise/etc.), avoiding DecodedVector overhead when the
  input is already dictionary-encoded.
```

### Checklist (For Author)

- [x] I have verified the code with local build.

### Breaking Changes

- [x] No

### DIFF
--- a/bolt/functions/lib/aggregates/AverageAggregateBase.h
+++ b/bolt/functions/lib/aggregates/AverageAggregateBase.h
@@ -36,6 +36,7 @@
 #include "bolt/type/DecimalUtil.h"
 #include "bolt/vector/ComplexVector.h"
 #include "bolt/vector/DecodedVector.h"
+#include "bolt/vector/DictionaryVector.h"
 #include "bolt/vector/FlatVector.h"
 namespace bytedance::bolt::functions::aggregate {
 
@@ -188,6 +189,13 @@ class AverageAggregateBase : public exec::Aggregate {
       const SelectivityVector& rows,
       const std::vector<VectorPtr>& args,
       bool /*mayPushdown*/) override final {
+    if (args[0]->encoding() == VectorEncoding::Simple::DICTIONARY &&
+        args[0]->valueVector()->encoding() !=
+            VectorEncoding::Simple::DICTIONARY &&
+        !isLazyNotLoaded(*args[0])) {
+      addDictionaryRawInput(groups, rows, *args[0]);
+      return;
+    }
     decodedRaw_.decode(*args[0], rows);
     if (decodedRaw_.isConstantMapping()) {
       if (!decodedRaw_.isNullAt(0)) {
@@ -231,6 +239,13 @@ class AverageAggregateBase : public exec::Aggregate {
       const SelectivityVector& rows,
       const std::vector<VectorPtr>& args,
       bool /*mayPushdown*/) override {
+    if (args[0]->encoding() == VectorEncoding::Simple::DICTIONARY &&
+        args[0]->valueVector()->encoding() !=
+            VectorEncoding::Simple::DICTIONARY &&
+        !isLazyNotLoaded(*args[0])) {
+      addSingleGroupDictionaryRawInput(group, rows, *args[0]);
+      return;
+    }
     decodedRaw_.decode(*args[0], rows);
 
     if (decodedRaw_.isConstantMapping()) {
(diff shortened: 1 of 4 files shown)


fzhedu:
it can avoid calling expensive combining dictionaries in decoded vectors, but it has extra cost, so we close it.

PR #937, opened 2026-08-25
Fix OrderBy reclaim race during input allocation

### LINKED ISSUE #938: [Bug] OrderBy input allocation can race with memory reclaim
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

`OrderBy::addInput()` keeps the whole method inside a `ReclaimableSectionGuard`. This includes `SortBuffer::addInput()`, which allocates rows and mutates its `RowContainer`.

If row allocation triggers memory arbitration, another thread can reclaim the OrderBy operator at the same time. Reclaim spills and clears the sort buffer while the driver is still updating allocator-backed row state. This race can corrupt `HashStringAllocator` ownership metadata and terminate the process with a native crash.

### Reproduction Steps

1. Run an OrderBy workload with spilling enabled and enough input to trigger memory arbitration while rows are being inserted into `SortBuffer`.
2. Inject or trigger arbitration from `MemoryPoolImpl::allocateNonContiguous()` for the OrderBy pool.
3. Concurrently reclaim the OrderBy operator while `SortBuffer::addInput()` is allocating a new row.
4. Observe that OrderBy reports itself reclaimable during the RowContainer mutation window. Under production load, the race can crash in `HashStringAllocator::freeToPool()`.

The existing `OrderByTest.reclaimDuringAllocation` test can cover the unsafe window by asserting that `op->canReclaim()` and `op->reclaimableBytes()` are false while the allocation hook is running.

### Bolt Version / Commit ID

main branch at `6a814e3c30134b4e5089e16cdd17dfc9bbf27f00`

### System Configuration

- **OS**: Linux 5.4, x86_64
- **Compiler**: GCC 12.5.0
- **Build Type**: Release for production reproduction; Debug test-value injection for the focused regression test
- **CPU Arch**: x86_64 AVX2
- **Framework**: Apache Spark with Gluten

### Logs / Stack Trace

```shell
HashStringAllocator::freeToPool
RowContainer::newRow
SortBuffer::addInput
OrderBy::addInput

freeToPool for block not allocated from pool of HashStringAllocator
```

### Expected Behavior

OrderBy must remain non-reclaimable while `SortBuffer::addInput()` mutates its `RowContainer`. Lazy input materialization may remain reclaimable before the mutation starts. Memory arbitration should never race with row allocation or corrupt allocator state.

### Additional context

The issue reproduces with a package built from main. Narrowing `ReclaimableSectionGuard` to cover only lazy vector materialization allows the same spill-heavy workload to finish without a JVM core dump or allocator corruption.

fzhedu:

Closes #938

`OrderBy::addInput()` marked both lazy input loading and `SortBuffer::addInput()` as reclaimable. During a memory arbitration triggered by row allocation, another thread could therefore reclaim the operator while `SortBuffer` was mutating its `RowContainer`. The concurrent spill clears allocator-backed row state and can lead to a native crash such as `HashStringAllocator::freeToPool` rejecting a block that no longer belongs to the allocator.

### Type of Change

- [ ] 📝 Documentation only

### Description

Limit `ReclaimableSectionGuard` to lazy vector materialization. `SortBuffer::addInput()` now runs in the driver's non-reclaimable section so row allocation cannot race with spill and clear.

Update `reclaimDuringAllocation` to verify that OrderBy is non-reclaimable while its `RowContainer` allocates rows.

### Performance Impact

- [x] **No Impact**: The change only narrows the interval in which OrderBy can be reclaimed to exclude an unsafe mutation window.

### Release Note

```text
Release Note:
- Fixed a native crash caused by concurrent OrderBy input allocation and memory reclaim.
```

### Validation

- Rebuilt `bolt_exec_test` successfully in Release mode.
- `git diff --check` passes.
- Reproduced the original crash with a main-based Gluten package.
- Ran the same Spark SQL workload with a package containing this fix: the application finished successfully, including all 80 tasks in the affected spill-heavy stage, with no JVM core dump or allocator corruption marker.
- The focused test is a `DEBUG_ONLY_TEST_P`; the available local build is Release, where test-value injection is compiled out, so its injected execution path was compile-checked but not executed locally.

### Checklist (For Author)

The Release test target builds successfully; a Debug build was not available locally. The changed lines follow the repository style and `git diff --check` passes. The available clang-format is version 20, while the repository requires 14.0.6, so I did not reformat unrelated code with a mismatched version. The pre-commit run was blocked while installing the gitleaks hook by the local Go environment (`error obtaining VCS status: exit status 128`).

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/exec/tests/OrderByTest.cpp
+++ b/bolt/exec/tests/OrderByTest.cpp
@@ -1583,11 +1583,14 @@ DEBUG_ONLY_TEST_P(OrderByTest, reclaimDuringAllocation) {
               }
               // Allocating rows mutates SortBuffer's RowContainer and must not
               // race with reclaim(), which spills and clears the same state.
-              EXPECT_FALSE(op->canReclaim());
+              EXPECT_TRUE(op->testingNonReclaimable());
+              EXPECT_EQ(op->canReclaim(), enableSpilling);
               uint64_t reclaimableBytes{0};
               const bool reclaimable = op->reclaimableBytes(reclaimableBytes);
-              EXPECT_FALSE(reclaimable);
-              EXPECT_EQ(reclaimableBytes, 0);
+              EXPECT_EQ(reclaimable, enableSpilling);
+              if (!enableSpilling) {
+                EXPECT_EQ(reclaimableBytes, 0);
+              }
               auto* driver = op->testingOperatorCtx()->driver();
               SuspendedSection suspendedSection(driver);
               testWait.notify();
@@ -1635,11 +1638,14 @@ DEBUG_ONLY_TEST_P(OrderByTest, reclaimDuringAllocation) {
       ASSERT_EQ(reclaimableBytes, 0);
     }
 
-    BOLT_ASSERT_THROW(
-        op->reclaim(
-            folly::Random::oneIn(2) ? 0 : folly::Random::rand32(rng_),
-            reclaimerStats_),
-        "");
+    reclaimerStats_.reset();
+    reclaimAndRestoreCapacity(
+        op,
+        folly::Random::oneIn(2) ? 0 : folly::Random::rand32(rng_),
+        reclaimerStats_);
+    ASSERT_EQ(reclaimerStats_.reclaimedBytes, 0);
+    ASSERT_EQ(
+        reclaimerStats_.numNonReclaimableAttempts, enableSpilling ? 1 : 0);
 
     driverWait.notify();
     Task::resume(task);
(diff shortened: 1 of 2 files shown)


bolt/exec/tests/OrderByTest.cpp
```diff
@@ -1581,15 +1581,13 @@ DEBUG_ONLY_TEST_P(OrderByTest, reclaimDuringAllocation) {
               if (!injectOnce.exchange(false)) {
                 return;
               }
-              ASSERT_EQ(op->canReclaim(), enableSpilling);
+              // Allocating rows mutates SortBuffer's RowContainer and must not
+              // race with reclaim(), which spills and clears the same state.
+              EXPECT_FALSE(op->canReclaim());
```
guhaiyan0221: This assertion does not match the reclaim contract. `OrderBy::canReclaim()` only checks that spilling is enabled and `sortBuffer_` exists; it does not consider `nonReclaimableSection_`. The non-reclaimable
  guard is enforced in `Operator::MemoryReclaimer::reclaim()`, which returns 0 and records a non-reclaimable attempt when `op_->nonReclaimableSection_` is set. In the spill-enabled branch this test can still see
  `op->canReclaim() == true` while `SortBuffer::addInput()` is correctly protected by the driver's non-reclaimable section, so the debug test will fail. The test should exercise the memory reclaimer path instead
  of asserting `canReclaim()` is false here.

fzhedu: fixed

guhaiyan0221 (approved):

PR #939, opened 2026-08-26
fix(dwrf): support reading BIGINT columns as VARCHAR

### LINKED ISSUE #941: [Bug] DWRF selective reader fails nested BIGINT-to-VARCHAR and VARCHAR-to-BOOLEAN schema evolution
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

The DWRF selective reader does not support nested schema evolution from `BIGINT` to `VARCHAR` or from `VARCHAR` to `BOOLEAN`. A query that reads an older DWRF/ORC physical schema through a newer requested schema can therefore fail native validation or crash while decoding the nested field.

Sparse struct projections expose an additional issue: `SelectiveStructColumnReader` uses `ScanSpec::channel()` as a requested-schema child index. The channel represents the output position, so it can differ from the schema position when only a subset of fields is projected. This can select the wrong child reader and lead to an invalid stream access.

### Reproduction Steps

1. Write DWRF data with this physical schema:

   ```text
   ROW<chat_id:BIGINT,meta_details:ROW<chatter_id:BIGINT,is_manual_set_nickname:VARCHAR>>
   ```

2. Read it with this requested schema:

   ```text
   ROW<chat_id:BIGINT,meta_details:ROW<chatter_id:VARCHAR,is_manual_set_nickname:BOOLEAN>>
   ```

3. Apply an `IS NOT NULL` filter to the nested struct and project only `meta_details`, so its output channel is `0` while its requested-schema index is `1`.
4. Read both direct and dictionary-encoded string stripes.

Before the fix, schema compatibility rejects the conversions or the selective reader can select the wrong child and fail during native decoding. A regression test is included in the linked PR.

### Bolt Version / Commit ID

`main` at `c7bd647cec78c9b6a646b4565d17c811433068c7`

### System Configuration

- **OS**: Linux
- **Compiler**: repository CI toolchain
- **Build Type**: Release, Spark-compatible
- **CPU Arch**: x86_64
- **Framework**: Apache Spark 3.2.1 with Gluten

### Logs / Stack Trace

```shell
Native validation failed: schema mismatch while reading nested DWRF fields

After allowing the requested conversions, a sparse projection could fail in native decoding because the output channel was used as the requested-schema child index.
```

### Expected Behavior

The DWRF reader should:

- convert `BIGINT` values to their `VARCHAR` representation;
- parse Spark-compatible boolean strings when reading `VARCHAR` as `BOOLEAN`, returning null for invalid values;
- map sparse projected struct children by their requested-schema field position, independently of their output channel; and
- support direct and dictionary string encodings without changing the physical reader type.

### Additional context

The proposed fix was validated with focused DWRF unit tests and an end-to-end Spark run that exercised the full scan, aggregation, and join path. The Spark application completed all 27 jobs successfully with no schema mismatch, native validation error, decoder crash, or sparse-projection assertion.

fzhedu:

Issue Number: close #941

The DWRF reader rejects schema evolution from `BIGINT` to `VARCHAR`, preventing Spark/Gluten queries from reading compatible data through the requested string schema.

### Type of Change

- [ ] 📝 Documentation only

### Description

- Add DWRF compatibility for reading `BIGINT` columns as `VARCHAR`.
- Support the conversion in both regular and selective DWRF readers.
- Add nested-reader and filter regression coverage.
- Reject value-dependent filters before BIGINT-to-VARCHAR selective reader
  construction while preserving IS NULL / IS NOT NULL, including filters
  installed dynamically after construction.
- Keep short-decimal-to-`VARCHAR` compatibility rejected.
- Use Bolt's legacy/unreachable macros required by this codebase.

This PR intentionally does not include `VARCHAR -> BOOLEAN`, selective string reader, sparse projection mapping, or other unrelated schema-evolution changes.

### Performance Impact

- [x] **No Impact**: Existing same-type read paths are unchanged; conversion is only performed when `BIGINT` is requested as `VARCHAR`.
- [ ] **Negative Impact**: Explained below.

### Release Note

```text
Support reading DWRF BIGINT columns as VARCHAR.
```

### Validation

- The five commits are patch-equivalent to internal Bolt MR !2523.
- Shared, DWRF, and Parquet regression coverage includes pushed-down,
  extracted, runtime, IS NULL, and IS NOT NULL filter paths.
- `pre-commit` passes for all changed files.
- Local compilation was attempted, but the reusable CMake cache points to a
  different checkout and has stale dependency include paths; GitHub Actions is
  the authoritative build validation.
- GitHub Actions were retriggered after syncing onto the latest `main`.

### Checklist (For Author)

- [x] I have run clang-format / linters for the changed code.

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/dwio/common/SelectiveColumnReader.cpp
+++ b/bolt/dwio/common/SelectiveColumnReader.cpp
@@ -209,6 +209,24 @@ void SelectiveColumnReader::getIntValues(
     RowSet rows,
     const TypePtr& requestedType,
     VectorPtr* result) {
+  if (requestedType->isVarchar()) {
+    VectorPtr integerResult;
+    getIntValues(rows, BIGINT(), &integerResult);
+    auto stringResult = BaseVector::create<FlatVector<StringView>>(
+        VARCHAR(), integerResult->size(), &memoryPool_);
+    auto integers = integerResult->as<SimpleVector<int64_t>>();
+    for (vector_size_t i = 0; i < integerResult->size(); ++i) {
+      if (integerResult->isNullAt(i)) {
+        stringResult->setNull(i, true);
+      } else {
+        const auto value = folly::to<std::string>(integers->valueAt(i));
+        stringResult->set(i, StringView(value));
+      }
+    }
+    *result = std::move(stringResult);
+    return;
+  }
+
   switch (requestedType->kind()) {
     case TypeKind::SMALLINT:
       switch (valueSize_) {
--- a/bolt/dwio/common/SelectiveColumnReader.cpp
+++ b/bolt/dwio/common/SelectiveColumnReader.cpp
@@ -78,8 +78,9 @@ SelectiveColumnReader::SelectiveColumnReader(
 void SelectiveColumnReader::validateReaderCastFilter() const {
   const auto* filter = scanSpec_->filter();
   if (fileType_->type()->kind() == TypeKind::BIGINT &&
-      !fileType_->type()->isDecimal() && requestedType_->isVarchar() &&
-      filter && !filter->isValueIndependent()) {
+      !fileType_->type()->isDecimal() &&
+      requestedType_->equivalent(*VARCHAR()) && filter &&
+      !filter->isValueIndependent()) {
     BOLT_USER_FAIL(
         "Cannot apply VARCHAR filter to physical BIGINT column {}",
         scanSpec_->fieldName());
@@ -222,7 +223,7 @@ void SelectiveColumnReader::getIntValues(
     RowSet rows,
     const TypePtr& requestedType,
     VectorPtr* result) {
-  if (requestedType->isVarchar()) {
+  if (requestedType->equivalent(*VARCHAR())) {
     VectorPtr integerResult;
     getIntValues(rows, BIGINT(), &integerResult);
     auto stringResult = BaseVector::create<FlatVector<StringView>>(
--- a/bolt/dwio/common/SelectiveColumnReader.cpp
+++ b/bolt/dwio/common/SelectiveColumnReader.cpp
@@ -71,7 +71,20 @@ SelectiveColumnReader::SelectiveColumnReader(
       requestedType_(requestedType),
       fileType_(fileType),
       formatData_(params.toFormatData(fileType, scanSpec)),
-      scanSpec_(&scanSpec) {}
+      scanSpec_(&scanSpec) {
+  validateReaderCastFilter();
+}
+
+void SelectiveColumnReader::validateReaderCastFilter() const {
+  const auto* filter = scanSpec_->filter();
+  if (fileType_->type()->kind() == TypeKind::BIGINT &&
+      !fileType_->type()->isDecimal() && requestedType_->isVarchar() &&
+      filter && !filter->isValueIndependent()) {
+    BOLT_USER_FAIL(
+        "Cannot apply VARCHAR filter to physical BIGINT column {}",
+        scanSpec_->fieldName());
+  }
+}
 
 void SelectiveColumnReader::filterRowGroups(
     uint64_t rowGroupSize,
@@ -557,6 +570,7 @@ void SelectiveColumnReader::setNulls(BufferPtr resultNulls) {
 }
 
 void SelectiveColumnReader::resetFilterCaches() {
+  validateReaderCastFilter();
   if (scanState_.filterCache.empty() && scanSpec_->hasFilter()) {
     scanState_.filterCache.resize(std::max<int32_t>(
         1, scanState_.dictionary.numValues + scanState_.dictionary2.numValues));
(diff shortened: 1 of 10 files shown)


bolt/dwio/common/TypeUtils.cpp
```diff
@@ -123,6 +123,7 @@ std::unordered_set<uint32_t> makeCompatibilityMap() {
   compat.insert(getKey(TypeKind::INTEGER, TypeKind::BIGINT));
   compat.insert(getKey(TypeKind::BIGINT, TypeKind::HUGEINT));
   compat.insert(getKey(TypeKind::REAL, TypeKind::DOUBLE));
+  compat.insert(getKey(TypeKind::BIGINT, TypeKind::VARCHAR));
```
guhaiyan0221, line 126: > Adding this pair to the global compatibility map makes the cast reachable for both DWRF and Parquet, but the selective readers still apply ScanSpec filters and row-group statistics in the physical BIGINT
  > domain before stringification. A logical VARCHAR predicate such as col = '11' produces a BytesRange; both reader paths then call testInt64Range() on it and fail with BytesRange ... testInt64Range() is not
  > supported. I reproduced this for both DWRF and Parquet.
  >
  > Please either evaluate a cast-aware filter or, consistently with the existing reader-cast policy, reject non-value-independent filters for BIGINT -> VARCHAR before reader construction while preserving IS
  > NULL / IS NOT NULL. Please also add coverage for pushed-down and extracted value filters.

fzhedu: Fixed in f0c4e8c022. BIGINT -> VARCHAR selective readers now reject non-value-independent filters before reader construction, while preserving IS NULL / IS NOT NULL (and other value-independent filters). The same validation runs from resetFilterCaches() so dynamically installed filters cannot bypass it. Added shared regression coverage for pushed-down, extracted, and runtime filters, plus DWRF and Parquet format-level coverage. pre-commit passes; the local targeted build remains unavailable because the reused CMake cache points to a different checkout and has stale dependency include paths, so GitHub CI is the authoritative build validation.

bolt/dwio/common/SelectiveColumnReader.cpp
```diff
@@ -209,6 +222,24 @@ void SelectiveColumnReader::getIntValues(
     RowSet rows,
     const TypePtr& requestedType,
     VectorPtr* result) {
+  if (requestedType->isVarchar()) {
```
guhaiyan0221: >  Apply the reader cast before invoking a ValueHook
  >
  > The BIGINT-to-VARCHAR conversion is performed only in getIntValues(), but lazy loads with a ValueHook never call getValues(). ColumnLoader::loadInternal() installs the hook and calls read(), and
  > SelectiveIntegerColumnReader::readCommon() then passes the decoded int64_t directly to ExtractToGenericHook. Since the exposed LazyVector type is VARCHAR, a string hook interprets that pointer as a
  > StringView, which can produce corrupted values or an out-of-bounds read.
  >
  > Please either disable the ValueHook path for reader-cast columns and materialize the converted vector first, or convert each BIGINT value to StringView before invoking the hook. A regression test that loads
  > a lazy BIGINT-as-VARCHAR column with TestingHook<StringView> would cover this path.

bolt/dwio/common/TypeUtils.cpp
guhaiyan0221, line 126: > Restrict this compatibility rule to the concrete VARCHAR type
  >
  > This check only compares TypeKind, but logical types may share the VARCHAR physical kind. In particular, JSON() derives from VarcharType, so a DWRF BIGINT column requested as JSON now passes this
  > compatibility check. Both new DWRF reader paths subsequently match isVarchar() and construct a vector with VARCHAR(), bypassing JsonCastOperator and losing the requested logical type.
  >
  > Please restrict this conversion to a target equivalent to VARCHAR() (or otherwise explicitly reject VARCHAR-backed logical types), and add a BIGINT-to-JSON compatibility regression test.

bolt/dwio/common/SelectiveColumnReader.cpp
fzhedu: Fixed in cdf5028cf1. The BIGINT-to-VARCHAR lazy ValueHook path now disables direct hook pushdown, materializes the converted VARCHAR vector first, and then invokes the hook with string values. Added `readBigintAsVarcharWithValueHook` coverage for negative, null, and large BIGINT values.

bolt/dwio/common/TypeUtils.cpp
fzhedu: Fixed in cdf5028cf1. BIGINT-to-VARCHAR compatibility and both DWRF reader paths now require a target equivalent to concrete `VARCHAR()`, so VARCHAR-backed logical types are not routed through this conversion. Added a BIGINT-to-JSON compatibility regression test.

bolt/dwio/common/SelectiveColumnReader.cpp
fzhedu: Follow-up: Release CI exposed that the regression test was not actually creating a LazyVector because it used the default eager scan spec. Fixed in 5aa145d42b by supplying an explicit projected ScanSpec, so the test now exercises the intended lazy ValueHook path.

PR #576, opened 2026-05-19
[fix] Throw on schema mismatch in ReaderBase::convertType

zhangxffff:
### What problem does this PR solve?

Issue Number: close #575 

### Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🚀 Performance improvement (optimization)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🔨 Refactoring (no logic changes)
- [ ] 🔧 Build/CI or Infrastructure changes
- [ ] 📝 Documentation only

### Description

When the table-declared column type doesn't match the Parquet file's physical/converted leaf type, `ReaderBase::convertType` used to silently return the file's leaf type. Downstream operators then reinterpreted the resulting vector under the wrong logical type and crashed. This PR adds a strict `BOLT_USER_CHECK` in every branch of `convertType`, throwing a clean `BoltUserError` at file-open time, before any column reader is built to avoid crash.

Also, since integer and varchar type data is implicitly cast during `IntegerColumnReader` and `StringColumnReader`, the newly added check also accepts mismatches between these types.

 #### Upstream references

  Adapted from these Velox PRs:

  - [facebookincubator/velox#12350](https://github.com/facebookincubator/velox/pull/12350) — main fix introducing `isCompatible` + per-branch `VELOX_CHECK` in `convertType`
  - [facebookincubator/velox#13864](https://github.com/facebookincubator/velox/pull/13864) — unannotated array exemption (`isRepeated && requestedType->isArray()`)

### Performance Impact

- [x] **No Impact**: This change does not affect the critical path (e.g., build system, doc, error handling).
- [ ] **Positive Impact**: I have run benchmarks.
    <details>
    <summary>Click to view Benchmark Results</summary>

    ```text
    Paste your google-benchmark or TPC-H results here.
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>
- [ ] **Negative Impact**: Explained below (e.g., trade-off for correctness).

### Release Note

Please describe the changes in this PR

Release Note:

```text
Release Note:
- fix coredump when schema not match
```

### Checklist (For Author)

- [x] I have added/updated unit tests (ctest).
- [x] I have verified the code with local build (Release/Debug).
- [x] I have run clang-format / linters.
- [ ] (Optional) I have run Sanitizers (ASAN/TSAN) locally for complex C++ changes.
[...]

PR #942, opened 2026-08-27
fix(parquet): support same-scale decimal precision widening

### LINKED ISSUE #940: [Bug] Parquet short-to-long decimal widening fails with Buffer::setSize capacity assertion
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

Bolt fails to read a Parquet DECIMAL column when the file stores the value as a short decimal (physical INT32 / INT64 / FIXED_LEN_BYTE_ARRAY with type length <= 8) but the Hive metastore / requested schema declares a same-scale long decimal (precision > 18, physical HUGEINT). The reader aborts with a buffer capacity assertion in `Buffer::setSize`, even though this is a valid precision widening that Spark can read successfully.

The production trigger is a column stored as `DECIMAL(13, 10)` and requested as `DECIMAL(28, 10)`. The column reader always uses `getFlatValues<int128_t, int128_t>` for the HUGEINT path, so it treats the decoded int64 buffer as if it held 16-byte values and requests a size larger than the buffer capacity.

This issue covers same-scale Decimal precision widening only. Scale-changing Decimal casts and rescaling are outside its scope and should be implemented separately together with complete value-filter and metadata-pruning semantics.

### Reproduction Steps

1. Write a Parquet file with a `DECIMAL(13, 10)` column.
2. Define the table schema with the column as `DECIMAL(28, 10)`.
3. Run a query that selects the column:

   ```sql
   SELECT a FROM <table>;
   ```

4. The query fails with a buffer capacity assertion.

A self-contained C++ reproduction that mirrors the production data shape:

```cpp
// file type: DECIMAL(13, 10) as FIXED_LEN_BYTE_ARRAY(6), 1024 rows (268 nulls)
// requested type: DECIMAL(28, 10) (HUGEINT)
auto data = makeRowVector(
    {"c0"},
    {makeNullableFlatVector<int64_t>(shortValues, DECIMAL(13, 10))});
writeToParquetFile(path, {data}, writerOptions);

auto declared = ROW({"c0"}, {DECIMAL(28, 10)});
auto plan = PlanBuilder(pool()).tableScan(declared, {}, "", declared).planNode();
AssertQueryBuilder(plan).split(makeSplit(path)).copyResults(pool());
```

### Bolt Version / Commit ID

main branch

### System Configuration

```markdown
```

### Logs / Stack Trace

```shell
BoltRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (8192 vs. 6048)
Expression: size <= capacity_
Function: setSize
```

### Expected Behavior

- Same-scale short-decimal Parquet values are widened safely to the requested long-decimal result.
- HUGEINT output reads dispatch according to the physical value width (`valueSize_` 4 / 8 / 16).
- Scale-changing Decimal reader casts are rejected with a user-facing error; this issue does not require rescaling.
- Value-dependent filters that cross the BIGINT/HUGEINT representation boundary fail fast; value-independent filters such as `IS NULL` continue to work.
- Empty files and direct RowReader entry points enforce the same compatibility checks.

### Additional context

Decimal rescaling and safe post-materialization evaluation for filters that cannot be pushed down should be tracked as separate follow-up work.

mikasa1110:
Fix Parquet DECIMAL schema evolution when the file uses a narrower physical representation than the requested schema, with a deliberately narrow scope: same-scale precision widening only.

Core changes:

- Dispatch HUGEINT reads according to the physical source width: 4, 8, or 16 bytes.
- Support same-scale Decimal precision widening, including short-to-long Decimal reads.
- Reject scale-changing Decimal reader casts before reading.
- Reject value-dependent filter pushdown when same-scale widening crosses the BIGINT/HUGEINT representation boundary; value-independent filters remain supported.
- Apply the same validation to direct RowReader users, dynamically added filters, and empty files.

Issue Number: close #940

Parquet files can store a Decimal column using a narrower physical representation than the requested table schema. For example, a file-level `DECIMAL(13, 10)` may be stored as `FIXED_LEN_BYTE_ARRAY(6)`, while the requested schema declares `DECIMAL(28, 10)`.

The HUGEINT reader previously always decoded the source as `int128_t`. When the physical source used 4 or 8 bytes, this caused the decoded buffer to be resized using the wrong source width and could trigger a `Buffer::setSize` capacity assertion.

This PR fixes that physical-width mismatch. It does not add Decimal rescaling: file and requested scales must match. Scale-changing Decimal casts are rejected explicitly and can be implemented separately together with complete filter and metadata-pruning semantics.

### Type of Change

- [ ] 📝 Documentation only

### Description

The HUGEINT path in `SelectiveColumnReader::getIntValues` now dispatches according to `valueSize_`:

- 16 bytes: `int128_t` → `int128_t`
- 8 bytes: `int64_t` → `int128_t`
- 4 bytes: `int32_t` → `int128_t`

Decimal schema compatibility is limited to precision widening with an unchanged scale. This covers the production case `DECIMAL(13, 10)` → `DECIMAL(28, 10)` without changing the unscaled value.

Scale-changing requests fail before materialization, including for empty files. When same-scale widening changes the native representation from BIGINT to HUGEINT, value-dependent reader filters fail fast instead of being evaluated against the physical representation. Value-independent filters such as `IS NULL` remain valid.

Regression coverage includes:

- `DECIMAL(13, 10)` stored as `FIXED_LEN_BYTE_ARRAY(6)` and read as `DECIMAL(28, 10)`.
- Rejection of scale-changing Decimal reader casts.
- Rejection of unsafe value filters across the BIGINT/HUGEINT representation boundary.
- Value-independent filters across short-to-long widening.
- Consistent validation for direct RowReader, dynamic filters, and empty files.

### Performance Impact

- [x] **No Material Impact**: Non-Decimal readers are unchanged. Same-scale widening performs no per-value rescaling; it only decodes according to the physical source width and widens the result type when needed.
- [ ] **Negative Impact**: Explained below.

### Validation

Release targets compiled:

```text
cmake --build --preset conan-release --target bolt_dwio_parquet_reader_test bolt_dwio_parquet_table_scan_test --parallel 8
```

Relevant tests:

```text
ParquetReaderTest.decimalReaderCast*: 2/2 passed
ParquetTableScanTest.convertTypePolicyMatrix and convertTypePolicyValueChecks: 2/2 passed
Related existing reader-cast, filter-only, map-filter, and metadata-filter regressions: 9/9 passed
```

### Release Note

```text
Release Note:
- Fixed same-scale Parquet Decimal precision widening across physical source widths.
```

### Checklist (For Author)

- [x] I have added/updated unit tests.
- [x] I have verified the affected targets with a local Release build.

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/dwio/common/SelectiveColumnReader.cpp
+++ b/bolt/dwio/common/SelectiveColumnReader.cpp
@@ -253,7 +253,19 @@ void SelectiveColumnReader::getIntValues(
       }
       break;
     case TypeKind::HUGEINT:
-      getFlatValues<int128_t, int128_t>(rows, result, requestedType);
+      switch (valueSize_) {
+        case 16:
+          getFlatValues<int128_t, int128_t>(rows, result, requestedType);
+          break;
+        case 8:
+          getFlatValues<int64_t, int128_t>(rows, result, requestedType);
+          break;
+        case 4:
+          getFlatValues<int32_t, int128_t>(rows, result, requestedType);
+          break;
+        default:
+          BOLT_FAIL("Unsupported value size: {}", valueSize_);
+      }
       break;
     case TypeKind::BIGINT:
       switch (valueSize_) {
--- a/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
+++ b/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
@@ -48,7 +48,64 @@
 #include "bolt/dwio/parquet/reader/TimestampColumnReader.h"
 #include "bolt/dwio/parquet/reader/VariantColumnReader.h"
 #include "bolt/dwio/parquet/thrift/codegen/parquet_types.h"
+#include "bolt/type/DecimalUtil.h"
 namespace bytedance::bolt::parquet {
+
+namespace {
+
+template <typename T>
+void applyDecimalScaleMultiplier(const VectorPtr& result, T multiplier) {
+  auto* flatResult = result->asUnchecked<FlatVector<T>>();
+  auto* rawValues = flatResult->mutableRawValues();
+  const auto* rawNulls = flatResult->rawNulls();
+  const auto size = flatResult->size();
+  if (!rawNulls) {
+    for (vector_size_t row = 0; row < size; ++row) {
+      rawValues[row] *= multiplier;
+    }
+  } else {
+    for (vector_size_t row = 0; row < size; ++row) {
+      if (!bits::isBitNull(rawNulls, row)) {
+        rawValues[row] *= multiplier;
+      }
+    }
+  }
+}
+
+void rescaleDecimalValues(
+    const TypePtr& fileType,
+    const TypePtr& requestedType,
+    const VectorPtr& result) {
+  BOLT_CHECK(fileType->isDecimal());
+  BOLT_CHECK(requestedType->isDecimal());
+
+  const auto fileScale = getDecimalPrecisionScale(*fileType).second;
+  const auto requestedScale = getDecimalPrecisionScale(*requestedType).second;
+  const auto scaleAdjust = requestedScale - fileScale;
+  BOLT_USER_CHECK_GE(
+      scaleAdjust,
+      0,
+      "Parquet does not support decimal scale narrowing: {}",
+      scaleAdjust);
+  BOLT_USER_CHECK_LE(
+      scaleAdjust,
+      LongDecimalType::kMaxPrecision,
+      "Decimal scale adjustment exceeds max precision: {}",
+      scaleAdjust);
+
+  // Schema validation guarantees enough target precision for scale widening.
+  if (scaleAdjust > 0) {
+    if (requestedType->isShortDecimal()) {
+      applyDecimalScaleMultiplier<int64_t>(
+          result, static_cast<int64_t>(DecimalUtil::kPowersOfTen[scaleAdjust]));
+    } else {
+      applyDecimalScaleMultiplier<int128_t>(
+          result, DecimalUtil::kPowersOfTen[scaleAdjust]);
+    }
+  }
+}
+
+} // namespace
 
 void IntegerColumnReader::getValues(const RowSet& rows, VectorPtr* result) {
   bool needConvertion = (castExprSet_ && castExprSet_->size() != 0);
@@ -74,6 +131,10 @@ void IntegerColumnReader::getValues(const RowSet& rows, VectorPtr* result) {
     getUnsignedIntValues(rows, requestedType, result);
   } else {
     getIntValues(rows, requestedType, result);
+  }
+
+  if (fileType.type()->isDecimal() && requestedType->isDecimal() && !allNull_) {
+    rescaleDecimalValues(fileType.type(), requestedType, *result);
   }
 
   if (needConvertion) {
--- a/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
+++ b/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
@@ -73,35 +73,15 @@ void applyDecimalScaleMultiplier(const VectorPtr& result, T multiplier) {
 }
 
 void rescaleDecimalValues(
-    const TypePtr& fileType,
     const TypePtr& requestedType,
-    const VectorPtr& result) {
-  BOLT_CHECK(fileType->isDecimal());
-  BOLT_CHECK(requestedType->isDecimal());
-
-  const auto fileScale = getDecimalPrecisionScale(*fileType).second;
-  const auto requestedScale = getDecimalPrecisionScale(*requestedType).second;
-  const auto scaleAdjust = requestedScale - fileScale;
-  BOLT_USER_CHECK_GE(
-      scaleAdjust,
-      0,
-      "Parquet does not support decimal scale narrowing: {}",
-      scaleAdjust);
-  BOLT_USER_CHECK_LE(
-      scaleAdjust,
-      LongDecimalType::kMaxPrecision,
-      "Decimal scale adjustment exceeds max precision: {}",
-      scaleAdjust);
-
-  // Schema validation guarantees enough target precision for scale widening.
-  if (scaleAdjust > 0) {
-    if (requestedType->isShortDecimal()) {
-      applyDecimalScaleMultiplier<int64_t>(
-          result, static_cast<int64_t>(DecimalUtil::kPowersOfTen[scaleAdjust]));
-    } else {
-      applyDecimalScaleMultiplier<int128_t>(
-          result, DecimalUtil::kPowersOfTen[scaleAdjust]);
-    }
+    const VectorPtr& result,
+    int32_t scaleAdjust) {
+  if (requestedType->isShortDecimal()) {
+    applyDecimalScaleMultiplier<int64_t>(
+        result, static_cast<int64_t>(DecimalUtil::kPowersOfTen[scaleAdjust]));
+  } else {
+    applyDecimalScaleMultiplier<int128_t>(
+        result, DecimalUtil::kPowersOfTen[scaleAdjust]);
   }
 }
 
@@ -133,8 +113,15 @@ void IntegerColumnReader::getValues(const RowSet& rows, VectorPtr* result) {
     getIntValues(rows, requestedType, result);
   }
 
-  if (fileType.type()->isDecimal() && requestedType->isDecimal() && !allNull_) {
-    rescaleDecimalValues(fileType.type(), requestedType, *result);
+  if (!allNull_ && fileType.type()->isDecimal() && requestedType->isDecimal()) {
+    const auto fileScale = getDecimalPrecisionScale(*fileType.type()).second;
+    const auto requestedScale = getDecimalPrecisionScale(*requestedType).second;
+
+    // Schema validation rejects scale narrowing and guarantees enough target
+    // precision for scale widening.
+    if (fileScale < requestedScale) {
+      rescaleDecimalValues(requestedType, *result, requestedScale - fileScale);
+    }
   }
 
   if (needConvertion) {
--- a/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
+++ b/bolt/dwio/parquet/reader/ParquetColumnReader.cpp
@@ -48,44 +48,7 @@
 #include "bolt/dwio/parquet/reader/TimestampColumnReader.h"
 #include "bolt/dwio/parquet/reader/VariantColumnReader.h"
 #include "bolt/dwio/parquet/thrift/codegen/parquet_types.h"
-#include "bolt/type/DecimalUtil.h"
 namespace bytedance::bolt::parquet {
-
-namespace {
-
-template <typename T>
-void applyDecimalScaleMultiplier(const VectorPtr& result, T multiplier) {
-  auto* flatResult = result->asUnchecked<FlatVector<T>>();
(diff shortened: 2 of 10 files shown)


bolt/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -76,6 +133,10 @@ void IntegerColumnReader::getValues(const RowSet& rows, VectorPtr* result) {
     getIntValues(rows, requestedType, result);
   }
 
+  if (fileType.type()->isDecimal() && requestedType->isDecimal() && !allNull_) {
```
wangxinshuo-bolt: Function entry should be narrower

bolt/dwio/parquet/reader/ParquetReaderCast.h
```diff
@@ -30,7 +30,9 @@ inline bool isReaderCastFilterMismatch(
       ((fileKind == TypeKind::TINYINT || fileKind == TypeKind::SMALLINT ||
         fileKind == TypeKind::INTEGER) &&
        requestedType->kind() == TypeKind::DOUBLE) ||
-      (fileType->isDate() && requestedType->isVarchar());
+      (fileType->isDate() && requestedType->isVarchar()) ||
```
wangxinshuo-bolt: 1. This function can be refactored
2. What problems would the filter have without this constraint?

bolt/dwio/common/SelectiveColumnReader.cpp
```diff
@@ -253,7 +253,19 @@ void SelectiveColumnReader::getIntValues(
       }
```
wangxinshuo-bolt, line 253: Need to use your company email to register a new account

wangxinshuo-bolt (changes requested):

bolt/dwio/common/SelectiveColumnReader.cpp
```diff
@@ -253,7 +253,19 @@ void SelectiveColumnReader::getIntValues(
...
       break;
```
wangxinshuo-bolt, line 254: Review the community's pull requests to see what other scenarios can be supported together

bolt/dwio/parquet/reader/ParquetReaderCast.h
mikasa1110: Refactored. These casts change the type or representation of values. Applying a filter built for the requested type directly to file-type statistics may produce **invalid comparisons and incorrectly skip row groups.** Therefore, type-dependent metadata filters are disabled, while value-independent filters such as IS NULL remain safe.

bolt/dwio/parquet/reader/ParquetColumnReader.cpp
mikasa1110: done

bolt/dwio/common/SelectiveColumnReader.cpp
mikasa1110: - INT32 → DOUBLE：Bolt implemented
- INT32 → BOOLEAN：Bolt implemented more.
- Decimal narrowing：supported, throw mismatch. 
So no other scenarios now.

bolt/dwio/common/SelectiveColumnReader.cpp
mikasa1110: done @mikasa1110-bolt

bolt/dwio/parquet/reader/ParquetReaderCast.h
```diff
@@ -23,14 +23,32 @@ namespace bytedance::bolt::parquet {
...
+    case TypeKind::INTEGER:
+      // Reader casts change the interpretation of DATE-to-VARCHAR and
+      // integer-to-DOUBLE filters.
+      return requestedKind == TypeKind::DOUBLE ||
+          (fileType->isDate() && requestedType->isVarchar());
+    case TypeKind::BIGINT:
+    case TypeKind::HUGEINT:
+      // Decimal type changes can alter the scale or native storage width.
+      return fileType->isDecimal() && requestedType->isDecimal() &&
+          !fileType->equivalent(*requestedType);
```
liuneng1994: [P2] equivalent() includes decimal precision, so this also classifies DECIMAL(10,2) to DECIMAL(15,2) as an unsafe filter mismatch even though the scale and native BIGINT representation are unchanged. I reproduced this on cf455f9 with values 1.00 and 2.00 and a pushed equality predicate for DECIMAL(15,2) value 1.00: the query should return one row, but it throws Cannot apply BIGINT filter to physical BIGINT Parquet column c0. Please treat only a scale change or a BIGINT/HUGEINT representation change as a decimal filter mismatch, and add coverage for same-scale precision widening.

bolt/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -76,6 +113,17 @@ void IntegerColumnReader::getValues(const RowSet& rows, VectorPtr* result) {
...
   }
 
+  if (!allNull_ && fileType.type()->isDecimal() && requestedType->isDecimal()) {
+    const auto fileScale = getDecimalPrecisionScale(*fileType.type()).second;
+    const auto requestedScale = getDecimalPrecisionScale(*requestedType).second;
+
+    // Schema validation rejects scale narrowing and guarantees enough target
+    // precision for scale widening.
+    if (fileScale < requestedScale) {
+      rescaleDecimalValues(requestedType, *result, requestedScale - fileScale);
```
liuneng1994: [P1] The scale conversion here happens after the pushed value filter has already run in readCommon(), so entry points that do not reject this filter can silently drop matching rows. I reproduced this on cf455f9 with a direct Parquet RowReader, dictionary encoding disabled, and type-dependent metadata pruning disabled: a DECIMAL(10,2) raw value 100 read as DECIMAL(15,4) with an equality filter on raw 10000 should return one row containing 10000, but the result contains zero rows. This isolates the per-value filter path: it compares the requested-scale filter value 10000 with the file-scale value 100 before this code can rescale it. Please translate the filter to file scale or prevent this pushdown on every reader entry point and evaluate it after materialization.

mikasa1110: Revised. For widened Decimal columns, value-dependent filters are now rejected at both RowReader creation and resetFilterCaches(). They are evaluated only after materialization and rescaling; added UT coverage for the reset path.

bolt/dwio/parquet/reader/ParquetReaderCast.h
mikasa1110: Fixed. Decimal filters are now considered mismatched only when the scale changes or the physical representation changes between BIGINT and HUGEINT. Same-scale precision widening remains eligible for pushdown. Added UT coverage for this case.

bolt/dwio/parquet/reader/ParquetColumnReader.cpp
mikasa1110: @liuneng1994  plz help to take a look when u r free. Thanks!

bolt/dwio/parquet/reader/ParquetReaderCast.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+} // namespace
+
+void validateReaderCastFilter(
+    const TypePtr& fileType,
+    const TypePtr& requestedType,
+    const common::ScanSpec& scanSpec,
+    const std::string& path) {
+  auto* filter = scanSpec.filter();
```
liuneng1994, line 67: [P1] Validate metadata filters before reader-cast pruning

`MetadataFilter` leaves are stored separately from `scanSpec.filter()`, but this validator only checks the latter. `ParquetData::filterRowGroups()` still evaluates every `scanSpec.metadataFilterAt(i)` against file statistics or dictionaries.

I reproduced a wrong-result case on this head: a `DECIMAL(10,2)` file containing 1,000 rows of `1.00`, requested as `DECIMAL(15,4)`, with metadata predicate `c0 IN (1.0000, 3.0000)`. Reader construction succeeds, but dictionary pruning returns 0 rows instead of 1,000 because requested-scale constants are compared with file-scale values. Please validate the metadata-filter leaves as well, or otherwise disable type-dependent metadata filtering for mismatched reader casts.

bolt/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1607,6 +1608,8 @@ class ParquetRowReader::Impl {
       requestedType_ = readerBase_->schema();
     }
 
+    validateReaderCastFilter();
```
liuneng1994: [P2] Run reader-cast filter validation for zero-row-group files

The `rowGroups_.empty()` return above occurs before `requestedType_` is initialized and before this validation call. As a result, the same unsafe value-dependent filter has different construction behavior depending on whether the file contains a row group.

I reproduced this with a zero-row-group `DECIMAL(10,2)` file requested as `DECIMAL(15,4)` and an equality filter in the requested scale: `createRowReader()` succeeds, while the identical schema and filter on a non-empty file throws as intended. Please initialize the requested type and validate the filter before the empty-file return, while retaining the early return before column-reader construction.

bolt/dwio/parquet/reader/ParquetReaderCast.cpp
mikasa1110-bolt: Addressed by narrowing the supported scope. This PR now supports same-scale Decimal precision widening only; scale-changing casts such as DECIMAL(10,2) → DECIMAL(15,4) fail during reader construction before metadata pruning. Rescaling and its metadata-filter semantics are left for separate follow-up work.

bolt/dwio/parquet/reader/ParquetReader.cpp
mikasa1110-bolt: Fixed. requestedType_ initialization and reader-cast validation now run before the zero-row-group early return. Scale-changing Decimal casts are consistently rejected for both empty and non-empty files, with a regression test covering the empty-file case.

bolt/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1174,17 +1175,18 @@ TypePtr ReaderBase::convertType(
...
         const auto filePrecision = schemaElement.precision;
         const auto fileScale = schemaElement.scale;
         checkRequested([&](const TypePtr& t) {
           if (!t->isDecimal()) {
             return false;
           }
           auto [precision, scale] = getDecimalPrecisionScale(*t);
-          return scale >= fileScale &&
-              (precision - filePrecision) >= (scale - fileScale);
+          return scale == fileScale && precision >= filePrecision;
```
mikasa1110-bolt, line 1189: @zhangxffff  plz help to review the changes here. I checked [PR576](https://github.com/bytedance/bolt/pull/576). Currently bolt does not perform rescaling value for decimal.Maybe we should reject scale mismatches entirely until rescaling is implemented?

wangxinshuo-bolt (approved):
LGTM

bolt/dwio/parquet/reader/ParquetReader.cpp
zhangxffff: > @zhangxffff plz help to review the changes here. I checked [PR576](https://github.com/bytedance/bolt/pull/576). Currently bolt does not perform rescaling value for decimal.Maybe we should reject scale mismatches entirely until rescaling is implemented?

Agreed. Better to reject implicit scale changes.

liuneng1994 (approved):

PR #948, opened 2026-09-01
fix: align year formatting for large years with Spark

### LINKED ISSUE #947: [Bug] Year formatting for years beyond 9999 diverges from Spark
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

Differential testing of `DateTimeFormatter::format` (the `from_unixtime` / `date_format` path) against Java ground truth — 325,730 cases covering boundary years, era/sign patterns, four timezones and both time policies, with `java.time.DateTimeFormatter` (what Spark's corrected policy uses, including Spark's `y`→`u` pattern rewrite) and `java.text.SimpleDateFormat` (Spark's legacy policy) as oracles — uncovered four related defects in year formatting, all introduced or exposed by the timestamp range extension (12703b65, "Extend timestamp range with CivilDateTime"):

1. **Leading `+` printed under LEGACY policy.** The spark-compatibility rule "year should contain sign if > 9999" matches Java's `SignStyle.EXCEEDS_PAD` (corrected policy only). Spark's legacy path (`SimpleDateFormat`) never prints `+`, but Bolt printed it for every policy.
2. **Wrong sign condition vs `SignStyle.EXCEEDS_PAD`.** Java prints `+` only when the year has more digits than the pattern width. Bolt hardcoded `year > 9999`, so `yyyyy` formatting year 57887 (5 digits = width) printed `+57887` while Spark prints `57887`.
3. **Silently corrupt output / spurious errors for years with 6+ digits.** `maxResultSize()` still sized YEAR / YEAR_OF_ERA fields for the pre-extension ±32767 range (6 chars). After the range extension to ±292278994, years with 6+ digits were silently truncated into corrupt output (e.g. `yyyyyy` on year −2 → `-00000` instead of `-000002`; `-100001-12-31 15:54:16` lost its trailing digit), and `yyyy G` with large BC years threw `BoltRuntimeError` because the `+` overflowed the undersized buffer. In the differential run this accounted for ~93k wrong results and ~13k spurious errors.
4. **Proleptic year 0 era off-by-one.** With an era in the pattern (or under LEGACY), year 0 formatted as `0000 BC`; Java/Spark print `0001 BC` (proleptic year 0 is 1 BC).

Additionally, the existing `FromUnixtimeLargeValues*SparkParity` tests recorded Spark's **default**-policy output (corrected formatting, with `+`) but ran under Bolt's `QueryConfig` default `timeParserPolicy = "legacy"` — they only passed because of defect 1 masking the policy mismatch.

### Reproduction Steps

With `SPARK_COMPATIBLE` build, `timeParserPolicy = "legacy"`:

```sql
SELECT from_unixtime(1764593923251, 'yyyy-MM-dd HH:mm:ss');
-- Bolt:  +57887-10-03 18:40:51
-- Spark (spark.sql.legacy.timeParserPolicy=LEGACY): 57887-10-03 18:40:51
```

With `timeParserPolicy = "corrected"`:

```sql
SELECT from_unixtime(1764570096000, 'yyyyy');
-- Bolt:  +57887        Spark: 57887   (5 digits do not exceed width 5)

SELECT from_unixtime(3093527980800, 'yyyy-MM-dd HH:mm:ss');
-- Bolt:  corrupt/truncated output     Spark: +100000-01-01 00:00:00

SELECT from_unixtime(-3217862448344, 'yyyy G');
-- Bolt:  BoltRuntimeError             Spark: +100002 BC

SELECT from_unixtime(-62167219200, 'yyyy G');
-- Bolt:  0000 BC                      Spark: 0001 BC
```

### Bolt Version / Commit ID

main branch @ f2fd4eaf

### System Configuration

- **OS**: Debian (Linux 6.12)
- **Compiler**: GCC (Release)
- **Build Type**: Release
- **CPU Arch**: x86_64 AVX2
- **Framework**: Spark (SPARK_COMPATIBLE build); ground truth from OpenJDK 21 `java.time` / `SimpleDateFormat`

### Logs / Stack Trace

```text
yyyy G, corrected, year -100001 (era +100002 BC):
  ERROR: BoltRuntimeError (result buffer overflow from undersized maxResultSize)
yyyyyy, corrected, year -2:
  expected -000002, got -00000 (truncated)
```

### Expected Behavior

Year formatting matches Spark for both policies: legacy (`SimpleDateFormat`) never prints `+`; corrected follows `SignStyle.EXCEEDS_PAD` (sign only when digits exceed pattern width); years up to ±292278994 format completely without truncation or errors; year 0 formats as era year 1 BC.

### Additional context

After the fix, the 325,730-case differential suite matches Java on all ~150k legacy cases. The remaining corrected-side diffs are pre-existing timezone-rule issues unrelated to year formatting and worth separate follow-up: (a) from 2038 onward (end of the 32-bit tzfile transition table) DST zones like `America/Los_Angeles` format summer instants with standard-time offset, 1 hour off from Spark/Java which extrapolate the DST rule indefinitely; (b) pre-1883 instants use the IANA LMT offset while `SimpleDateFormat` uses the whole-hour raw offset.

zhangxffff:

Year formatting in `DateTimeFormatter::format` (the `from_unixtime` / `date_format` path) diverges from Spark for years beyond 9999, in four ways uncovered by a 325,730-case differential test against Java ground truth (`java.time.DateTimeFormatter` with Spark's `y`→`u` rewrite for the corrected policy, `java.text.SimpleDateFormat` for legacy). See the linked issue for full details.

Issue Number: close #947

### Type of Change

- [ ] 📝 Documentation only

### Description

Four fixes in `DateTimeFormatter.cpp`:

1. **Gate the leading `+` on non-LEGACY policies.** Spark's legacy path (`SimpleDateFormat`) never prints `+` for years beyond 9999; only the corrected path (java.time `SignStyle.EXCEEDS_PAD`) does. The sign was previously printed unconditionally.
2. **Match `SignStyle.EXCEEDS_PAD` exactly.** Java prints the sign only when the year has more digits than the pattern width, not for any year > 9999. Replaced the hardcoded `year > 9999` with `countDigits(year) > minRepresentDigits` — e.g. `yyyyy` on year 57887 (5 digits = width 5) must print `57887`, not `+57887`.
3. **Fix `maxResultSize()` for the extended year range.** It still sized YEAR / YEAR_OF_ERA fields for the pre-CivilDateTime ±32767 range (6 chars). Since the range extension to ±292278994, years with 6+ digits were silently truncated into corrupt output, and `yyyy G` with large BC years threw `BoltRuntimeError` (the `+` overflowed the buffer). Now sized for 9 digits plus a sign; CENTURY_OF_ERA for 7 digits.
4. **Year-0 era off-by-one.** Proleptic year 0 is 1 BC: the era adjustment used `year < 0` and formatted year 0 as `0000 BC` instead of `0001 BC` (Java/Spark behavior). Changed to `year <= 0` (and the same for the UTC-year computation in the `+`-suppression check).

Test changes:

- `FromUnixtimeLargeValuesSparkParity` / `FromUnixtimeLargeValuesSparkTimezone` recorded Spark's **default**-policy output (corrected formatting, with `+`) but ran under Bolt's `QueryConfig` default `timeParserPolicy = "legacy"` — they only passed because bug 1 masked the policy mismatch. The full case table now runs under both policies: `corrected` expecting the `+` prefix and `legacy` expecting none. (Production via Gluten is unaffected: the native side defaults to `corrected` when the session conf is not set.)
- New `FromUnixtimeYearSignEdgeCases` pins Java-derived expectations for: a 6-digit year with sign (previously truncated), the EXCEEDS_PAD width rule, negative-year zero padding, year 0 as `0001 BC`, a large BC era year (previously threw), and legacy without `+`.
- `JodaDateTimeFormatterTest.formatResultSize` updated to the new (correct) size bounds.

**Verification:** after the fix the differential suite matches Java on all ~150k legacy cases and all corrected cases except two known pre-existing classes unrelated to year formatting (post-2037 DST extrapolation and pre-1883 LMT timezone rules, plus the deliberate UTC-year `+`-suppression from 9f5eefa54) — overall diffs dropped from 110,223 to 6,782, none of them year-digit related.

### Performance Impact

`maxResultSize()` now reserves a few more bytes per year field for string results; the formatting hot loop is unchanged apart from a `countDigits` call on the already-computed year.

### Release Note

```text
Release Note:
- Fixed year formatting for years beyond 9999 to match Spark: no leading '+' under legacy time parser policy, '+' only when the year exceeds the pattern width (Java SignStyle.EXCEEDS_PAD), no truncated/corrupt output or errors for years with 6+ digits, and proleptic year 0 now formats as era year 0001 BC.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/functions/lib/DateTimeFormatter.cpp
+++ b/bolt/functions/lib/DateTimeFormatter.cpp
@@ -1151,8 +1151,9 @@ uint32_t DateTimeFormatter::maxResultSize(const tz::TimeZone* timezone) const {
         size += 2;
         break;
       case DateTimeFormatSpecifier::YEAR_OF_ERA:
-        // Timestamp is in [-32767-01-01, 32767-12-31] range.
-        size += std::max((int)token.pattern.minRepresentDigits, 6);
+        // Timestamp covers years up to +/-292278994 (9 digits), plus one
+        // char for a leading sign.
+        size += std::max((int)token.pattern.minRepresentDigits, 9) + 1;
         break;
       case DateTimeFormatSpecifier::DAY_OF_WEEK_0_BASED:
       case DateTimeFormatSpecifier::DAY_OF_WEEK_1_BASED:
@@ -1166,12 +1167,16 @@ uint32_t DateTimeFormatter::maxResultSize(const tz::TimeZone* timezone) const {
       case DateTimeFormatSpecifier::WEEK_YEAR:
         [[fallthrough]];
       case DateTimeFormatSpecifier::YEAR:
-        // Timestamp is in [-32767-01-01, 32767-12-31] range.
+        // Timestamp covers years up to +/-292278994 (9 digits), plus one
+        // char for a leading sign.
         size += token.pattern.minRepresentDigits == 2
             ? 2
-            : std::max((int)token.pattern.minRepresentDigits, 6);
+            : std::max((int)token.pattern.minRepresentDigits, 9) + 1;
         break;
       case DateTimeFormatSpecifier::CENTURY_OF_ERA:
+        // Century of +/-292278994 has up to 7 digits.
+        size += std::max((int)token.pattern.minRepresentDigits, 7);
+        break;
       case DateTimeFormatSpecifier::DAY_OF_YEAR:
         size += std::max((int)token.pattern.minRepresentDigits, 3);
         break;
@@ -1293,13 +1298,18 @@ int32_t DateTimeFormatter::format(
                 padContent(std::abs(year) % 100, '0', 2, maxResultEnd, result);
           } else {
             year = year <= 0 ? std::abs(year - 1) : year;
-            // spark compatibility: year should contain sign if > 9999.
+            // spark compatibility: Java DateTimeFormatter uses
+            // SignStyle.EXCEEDS_PAD for 4+ pattern letters, printing '+'
+            // when the year has more digits than the pattern width; legacy
+            // SimpleDateFormat never prints '+'.
             // However, if a timezone is applied and the same instant in UTC
             // has year-of-era <= 9999 (e.g., Asia/Shanghai at
             // 10000-01-01 07:59:59 equals 9999-12-31 23:59:59 UTC), suppress
             // the leading '+'.
-            bool addPlus = ::bytedance::bolt::kSparkCompatible && year > 9999 &&
-                token.pattern.minRepresentDigits >= 4;
+            bool addPlus = ::bytedance::bolt::kSparkCompatible &&
+                timePolicy != TimePolicy::LEGACY &&
+                token.pattern.minRepresentDigits >= 4 && year > 0 &&
+                countDigits(year) > token.pattern.minRepresentDigits;
             if (addPlus && timezone != nullptr) {
               const auto civilDateTimeUtc =
                   util::toCivilDateTime(timestamp, allowOverflow, isPrecision);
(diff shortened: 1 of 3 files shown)


wangxinshuo-bolt (approved):
LGTM

PR #963, opened 2026-09-04
fix: decide year plus sign by local year only, matching Spark

### LINKED ISSUE #426: [Bug] from_unixtime returns unnecessary '+' given a large value in Spark
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

`from_unixtime(253402300799, 'yyyy-MM-dd HH:mm:ss')` returns '+10000-01-01 07:59:59' whereas Spark returns '10000-01-01 07:59:59'.

### Reproduction Steps

Run the following SQL in Spark on Bolt
```
set spark.sql.session.timeZone=Asia/Shanghai;
SELECT from_unixtime(253402300799, 'yyyy-MM-dd HH:mm:ss');
```

### Bolt Version / Commit ID

0c01dc4019433e363b6d4ee63624f08a843ef042

### System Configuration

```markdown
```

### Logs / Stack Trace

```shell

```

### Expected Behavior

_No response_

### Additional context

_No response_

### LINKED ISSUE #962: [Bug] Year '+' sign decided by UTC year instead of local year with a timezone
### Component Selection

- [ ] Build
- [ ] Other

### Describe the Bug

When a timezone is applied, `date_format` / `from_unixtime` decide the year's leading `+` from the **UTC** year instead of the local (timezone-adjusted) year, which diverges from Spark.

`DateTimeFormatter::format` prints the year sign following Java's `SignStyle.EXCEEDS_PAD`: a `+` when the year has more digits than the pattern width. But for the timezone case, the sign is then overridden by comparing the **UTC** year-of-era against a fixed `9999` threshold (introduced in `9f5eefa54`, PR #427). This assumes Spark decides the sign from UTC time. It does not: Spark formats via `java.time` (`formatter.withZone(zoneId).format(instant)`), so the year handed to the formatter — and thus the sign — is the **local** year alone. UTC never participates.

Consequences:
1. **Sign dropped when it should stay.** `set spark.sql.session.timeZone=Asia/Shanghai; from_unixtime(253402300799, 'yyyy-MM-dd HH:mm:ss')` — local year is 10000, so Spark (corrected) prints `+10000-01-01 07:59:59`, but Bolt suppresses the `+` because the UTC year is 9999.
2. **Wider patterns break at their own boundary.** Even after aligning the threshold, a fixed `9999` is wrong for widths above 4: `yyyyy` should print `+` only when the year exceeds 5 digits.

Root cause note: the UTC-based rule was added in PR #427 to fix issue #426, which reported `from_unixtime(253402300799, ...)` at `Asia/Shanghai` returning a `+` that Spark did not. That observation was actually **legacy**-policy behavior (SimpleDateFormat never prints `+`); the reporter attributed it to a UTC rule. The correct fix for #426 is the legacy gate (already merged in #948), not a UTC comparison.

### Reproduction Steps

Verified against Apache Spark 3.2.4 and 3.5.9:

```sql
set spark.sql.session.timeZone=Asia/Shanghai;

-- corrected (default policy)
select from_unixtime(253402300799, 'yyyy-MM-dd HH:mm:ss');
-- Spark: +10000-01-01 07:59:59      Bolt: 10000-01-01 07:59:59  (sign wrongly suppressed)

-- legacy policy
set spark.sql.legacy.timeParserPolicy=LEGACY;
select from_unixtime(253402300799, 'yyyy-MM-dd HH:mm:ss');
-- Spark: 10000-01-01 07:59:59       Bolt: 10000-01-01 07:59:59  (agree)

-- wider pattern, corrected: UTC 99999 vs local 100000
set spark.sql.session.timeZone=Asia/Shanghai;
set spark.sql.legacy.timeParserPolicy=CORRECTED;
select from_unixtime(3093527952000, 'yyyyy');
-- Spark: +100000                    Bolt (fixed-9999 UTC rule): 100000
```

### Bolt Version / Commit ID

main branch @ 625ee2ee (contains #948 `7707ea35`)

### System Configuration

- **OS**: Debian (Linux 6.12)
- **Compiler**: GCC (Release)
- **Build Type**: Release
- **CPU Arch**: x86_64 AVX2
- **Framework**: Spark (SPARK_COMPATIBLE); ground truth from Spark 3.2.4 (JDK 11) and 3.5.9

### Expected Behavior

The `+` follows the local (timezone-adjusted) year only, per Java `SignStyle.EXCEEDS_PAD`, with no UTC comparison. `date_format`/`from_unixtime` should match Spark for both corrected and legacy policies across timezones.

### Additional context

Source evidence — Apache Spark tag `v3.2.4`, pinned to commit [`0ae10ac1`](https://github.com/apache/spark/tree/0ae10ac18298d1792828f1d59b652ef17462d76e) (lightweight tag → this commit), permalinks below:

1. corrected — zone used once, only to project the instant into local time; the formatter then sees the local year ([`TimestampFormatter.scala#L129-L139`](https://github.com/apache/spark/blob/0ae10ac18298d1792828f1d59b652ef17462d76e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/TimestampFormatter.scala#L129-L139)):

```scala
override def format(instant: Instant): String = {
  try {
    formatter.withZone(zoneId).format(instant)
  } catch checkFormattedDiff(...)
}
override def format(us: Long): String = {
  val instant = DateTimeUtils.microsToInstant(us)
  format(instant)
}
```

2. pattern — `y` → `u` when no era designator, so 4+ letters map to `SignStyle.EXCEEDS_PAD` on the local year ([`DateTimeFormatterHelper.scala#L334-L339`](https://github.com/apache/spark/blob/0ae10ac18298d1792828f1d59b652ef17462d76e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeFormatterHelper.scala#L334-L339)):

```scala
if (!eraDesignatorContained) {
  patternPart.replace("y", "u")
} else {
  patternPart
}
```

3. legacy — `LegacyFastTimestampFormatter` uses `FastDateFormat` (SimpleDateFormat semantics), which never prints `+` ([`TimestampFormatter.scala#L250-L277`](https://github.com/apache/spark/blob/0ae10ac18298d1792828f1d59b652ef17462d76e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/TimestampFormatter.scala#L250-L277)).

Differential testing of 325,730 cases against Java ground truth (itself verified against Spark 3.2.4 on 162,750 cases) shows zero sign-related diffs once the UTC suppression is removed.

zhangxffff:

With a timezone applied, `date_format` / `from_unixtime` decided the year's leading `+` from the **UTC** year against a fixed `9999` threshold, instead of from the local (timezone-adjusted) year. Spark decides the sign from the local year only (Java `SignStyle.EXCEEDS_PAD`), so Bolt dropped the `+` where Spark keeps it (e.g. `Asia/Shanghai` local year 10000, UTC year 9999).

Issue Number: close #962

### Type of Change

- [ ] 📝 Documentation only

### Description

`9f5eefa54` (PR #427) added a UTC-based suppression: when a timezone is set, the `+` for years beyond 9999 was suppressed if the same instant's **UTC** year did not exceed 9999. It was introduced to fix #426, where `from_unixtime(253402300799, ...)` at `Asia/Shanghai` printed a `+` that Spark did not — but that observation was **legacy**-policy behavior (SimpleDateFormat never prints `+`), misattributed to a UTC rule. Under corrected policy Spark actually **does** print `+10000-...` there.

This PR removes the UTC suppression in both the `YEAR_OF_ERA` and `YEAR` branches. The `+` now follows the local year alone, exactly like Java's `SignStyle.EXCEEDS_PAD` (`countDigits(localYear) > patternWidth`), which is what Spark's `Iso8601TimestampFormatter` does — `formatter.withZone(zoneId).format(instant)` uses the zone only to project the instant into local time, then formats the local year. The legacy gate (no `+` under legacy policy), merged in #948, remains and is the correct fix for #426.

**Source evidence** — Apache Spark tag `v3.2.4`, pinned to commit [`0ae10ac1`](https://github.com/apache/spark/tree/0ae10ac18298d1792828f1d59b652ef17462d76e) (a lightweight tag pointing straight at this commit), so the permalinks below are stable:

1. corrected — zone is used once, only to project the instant into local time ([`TimestampFormatter.scala#L129-L139`](https://github.com/apache/spark/blob/0ae10ac18298d1792828f1d59b652ef17462d76e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/TimestampFormatter.scala#L129-L139)):

```scala
override def format(instant: Instant): String = {
  try {
    formatter.withZone(zoneId).format(instant)   // zoneId used once, for local projection
  } catch checkFormattedDiff(toJavaTimestamp(instantToMicros(instant)),
    (t: Timestamp) => format(t))
}

override def format(us: Long): String = {
  val instant = DateTimeUtils.microsToInstant(us)   // absolute instant, no zone
  format(instant)
}
```

2. pattern — `y`→`u` when no era, so 4+ letters map to `SignStyle.EXCEEDS_PAD` on the local year ([`DateTimeFormatterHelper.scala#L334-L339`](https://github.com/apache/spark/blob/0ae10ac18298d1792828f1d59b652ef17462d76e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeFormatterHelper.scala#L334-L339)):

```scala
// In DateTimeFormatter, 'u' supports negative years. We substitute 'y' to 'u'
// here ... We only do this substitution when there is no era designator found.
if (!eraDesignatorContained) {
  patternPart.replace("y", "u")
} else {
  patternPart
}
```

3. legacy — `FastDateFormat` (SimpleDateFormat semantics), never prints `+` ([`TimestampFormatter.scala#L250-L277`](https://github.com/apache/spark/blob/0ae10ac18298d1792828f1d59b652ef17462d76e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/TimestampFormatter.scala#L250-L277)):

```scala
class LegacyFastTimestampFormatter(pattern: String, zoneId: ZoneId, locale: Locale)
    extends TimestampFormatter {
  @transient private lazy val fastDateFormat =
    FastDateFormat.getInstance(pattern, TimeZone.getTimeZone(zoneId), locale)
  // ...
  override def format(timestamp: Long): String = {
    val julianMicros = rebaseGregorianToJulianMicros(timestamp)
    cal.setTimeInMillis(...)
    fastDateFormat.format(cal)
  }
```

Behavior changes (all verified against Spark 3.2.4/3.5.9):
- `Asia/Shanghai`, corrected: `from_unixtime(253402300799, 'yyyy-MM-dd HH:mm:ss')` → `+10000-01-01 07:59:59` (was `10000-...`).
- Wider patterns at their own boundary: `yyyyy` prints `+` only when the year exceeds 5 digits — a fixed 9999 threshold was also wrong here.

Tests now exercise **both** corrected and legacy policies:
- `FromUnixtimeLargeValuesSparkParity` already ran the full case table under both policies.
- `FromUnixtimeLargeValuesSparkTimezone` is reworked to run every timezone case under both policies (corrected with `+`, legacy without), with expectations taken from real Spark.
- `FromUnixtimeYearSignEdgeCases` adds legacy assertions mirroring its corrected sign/width cases, plus the `yyyyy`/`yyyy` timezone-boundary cases.

**Verification:** after removing the suppression, the 325,730-case differential against the Spark-verified Java ground truth has zero sign-related diffs; the only remaining diffs are unrelated timezone-rule issues (pre-1883 LMT, post-2037 DST extrapolation).

### Performance Impact

Removes a `toCivilDateTime` call from the timezone formatting path (the suppression no longer recomputes the UTC civil date).

### Release Note

```text
Release Note:
- Fixed date_format/from_unixtime to decide the year's leading '+' from the local (timezone-adjusted) year, matching Spark, instead of the UTC year. With a session timezone, years beyond the pattern width now print '+' consistently (e.g. Asia/Shanghai '+10000-01-01 07:59:59' under corrected policy).
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/functions/lib/DateTimeFormatter.cpp
+++ b/bolt/functions/lib/DateTimeFormatter.cpp
@@ -1301,23 +1301,12 @@ int32_t DateTimeFormatter::format(
             // spark compatibility: Java DateTimeFormatter uses
             // SignStyle.EXCEEDS_PAD for 4+ pattern letters, printing '+'
             // when the year has more digits than the pattern width; legacy
-            // SimpleDateFormat never prints '+'.
-            // However, if a timezone is applied and the same instant in UTC
-            // has year-of-era <= 9999 (e.g., Asia/Shanghai at
-            // 10000-01-01 07:59:59 equals 9999-12-31 23:59:59 UTC), suppress
-            // the leading '+'.
+            // SimpleDateFormat never prints '+'. Like Java, the decision is
+            // based on the local (timezone-adjusted) year only.
             bool addPlus = ::bytedance::bolt::kSparkCompatible &&
                 timePolicy != TimePolicy::LEGACY &&
                 token.pattern.minRepresentDigits >= 4 && year > 0 &&
                 countDigits(year) > token.pattern.minRepresentDigits;
-            if (addPlus && timezone != nullptr) {
-              const auto civilDateTimeUtc =
-                  util::toCivilDateTime(timestamp, allowOverflow, isPrecision);
-              const auto utcYearEra = civilDateTimeUtc.date.year <= 0
-                  ? std::abs(civilDateTimeUtc.date.year - 1)
-                  : civilDateTimeUtc.date.year;
-              addPlus = utcYearEra > 9999;
-            }
             if (addPlus) {
               *result++ = '+';
             }
@@ -1397,23 +1386,12 @@ int32_t DateTimeFormatter::format(
             // spark compatibility: Java DateTimeFormatter uses
             // SignStyle.EXCEEDS_PAD for 4+ pattern letters, printing '+'
             // when the year has more digits than the pattern width; legacy
-            // SimpleDateFormat never prints '+'.
-            // If a timezone is applied and the same instant in UTC has
-            // (adjusted) year <= 9999, suppress the leading '+'.
+            // SimpleDateFormat never prints '+'. Like Java, the decision is
+            // based on the local (timezone-adjusted) year only.
             bool addPlus = ::bytedance::bolt::kSparkCompatible &&
                 timePolicy != TimePolicy::LEGACY &&
                 token.pattern.minRepresentDigits >= 4 && year > 0 &&
                 countDigits(year) > token.pattern.minRepresentDigits;
-            if (addPlus && timezone != nullptr) {
-              const auto civilDateTimeUtc =
-                  util::toCivilDateTime(timestamp, allowOverflow, isPrecision);
-              auto utcYear = civilDateTimeUtc.date.year;
-              if (utcYear <= 0 &&
-                  (hasEra_ || timePolicy == TimePolicy::LEGACY)) {
-                utcYear = std::abs(utcYear) + 1;
-              }
-              addPlus = utcYear > 9999;
-            }
             if (addPlus) {
               *result++ = '+';
             }
--- a/bolt/functions/sparksql/tests/SparkSqlDateTimeFunctionsTest.cpp
+++ b/bolt/functions/sparksql/tests/SparkSqlDateTimeFunctionsTest.cpp
@@ -2282,19 +2282,33 @@ TEST_F(SparkSqlDateTimeFunctionsTest, FromUnixtimeYearSignEdgeCases) {
   // because the '+' overflowed the undersized result buffer.
   EXPECT_EQ(fromUnixTime(-3217862448344, "yyyy G"), "+100002 BC");
 
+  // With a timezone, the sign still follows the local year only (Java
+  // semantics): UTC 99999-12-31 16:00:00 is Shanghai 100000-01-01
+  // 00:00:00, whose 6 digits exceed both widths below.
+  auto fromUnixTimeTz = [&](int64_t unixTime, const std::string& format) {
+    return evaluateOnce<std::string>(
+               fmt::format("from_unixtime(c0, '{}', 'Asia/Shanghai')", format),
+               std::optional<int64_t>{unixTime})
+        .value();
+  };
+  EXPECT_EQ(fromUnixTimeTz(3093527952000, "yyyyy"), "+100000");
+  EXPECT_EQ(fromUnixTimeTz(3093527952000, "yyyy"), "+100000");
+
+  // Legacy policy (SimpleDateFormat) never prints a leading '+', for both
+  // the same instants and the same widths as above.
   setTimeParserPolicy("legacy");
-  // SimpleDateFormat never prints a leading '+'.
   EXPECT_EQ(
       fromUnixTime(3093527980800, "yyyy-MM-dd HH:mm:ss"),
       "100000-01-01 00:00:00");
+  // Width equal to the digit count still has no sign, matching corrected.
+  EXPECT_EQ(fromUnixTime(1764570096000, "yyyyy"), "57887");
+  EXPECT_EQ(fromUnixTimeTz(3093527952000, "yyyyy"), "100000");
+  EXPECT_EQ(fromUnixTimeTz(3093527952000, "yyyy"), "100000");
   resetQueryConfig();
 }
 
 TEST_F(SparkSqlDateTimeFunctionsTest, FromUnixtimeLargeValuesSparkTimezone) {
-  // Expected values follow Spark's default (corrected) formatting; Bolt's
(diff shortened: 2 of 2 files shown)


wangxinshuo-bolt (approved):
LGTM

PR #969, opened 2026-09-04
feat(arrow): export constant arrays as dictionaries

### LINKED ISSUE #968: [Feature] Add an opt-in Arrow Dictionary representation for constant arrays
# [feat] Add an opt-in Arrow Dictionary representation for constant arrays

## Feature Category

Performance Optimization

## Problem / Use Case

Bolt currently exports a `ConstantVector<Array<T>>` as an Arrow run-end encoded
array when `ArrowOptions::flattenConstant` is disabled. This representation is
compact, but some Arrow consumers do not support run-end encoded arrays. Such
consumers must request flattened constants instead.

Flattening a constant array duplicates the same list and all of its elements for
every selected row. For a batch of `N` rows and a constant array with `M`
elements, this turns one logical list into `N` physical copies. The cost becomes
especially large for long arrays and variable-width elements.

Issue #148 and PR #142 added support for materializing dictionary/constant
encoded complex vectors when flattening is requested. This proposal addresses a
different case: preserving a compact representation for consumers that support
Arrow Dictionary encoding but not run-end encoding.

## Proposed Solution

Add an opt-in `ArrowOptions` flag that exports supported non-null
`ConstantVector<Array<T>>` values as:

```text
Dictionary<int32, List<T>>
  indices:    [0, 0, ..., 0]
  dictionary: [the single referenced array value]
```

The default remains disabled, so existing callers continue to receive the
current run-end encoded representation. `flattenConstant=true` keeps precedence
and continues to request fully materialized output.

The initial supported element types would be Boolean, signed integer types,
Float, Double, VARCHAR, and VARBINARY. Date, Timestamp, Decimal, nested Array,
Map, Row, null constants, and arbitrary DictionaryVector inputs would keep the
existing behavior.

If the backing `ArrayVector` has dictionary-encoded elements, normalize only the
elements before export. This keeps the outer constant compact while avoiding a
nested `Dictionary<List<Dictionary<T>>>` representation.

Arrow schema and array exporters must use the same eligibility predicate so the
index schema and buffers remain symmetric.

## References / Prior Art

- Arrow Columnar Format: Dictionary-encoded layout
- Arrow Columnar Format: Run-end encoded layout
- #148 / #142 for flattened export of encoded complex vectors

## Importance

Low

## Willingness to Contribute

Yes, I can submit a PR

mikasa1110-bolt:

Issue Number: close #968

Bolt exports `ConstantVector<Array<T>>` as run-end encoded Arrow data when
constant flattening is disabled. Some Arrow consumers support Dictionary arrays
but not run-end encoded arrays. Their current fallback is to flatten the
constant, which duplicates the same list and its elements for every row.

This PR adds a compact, opt-in Dictionary representation for that compatibility
case while preserving all existing defaults.

### Type of Change

- [ ] 📝 Documentation only

### Description

- add `ArrowOptions::arrayConstantAsDictionary`, disabled by default
- export eligible non-null `ConstantVector<Array<T>>` values as
  `Dictionary<int32, List<T>>`, with one dictionary value and all-zero indices
- use one eligibility predicate for ArrowSchema and ArrowArray export
- export only the referenced backing value, including constants with a non-zero
  backing index
- normalize dictionary-encoded array elements while preserving the outer
  constant encoding, avoiding nested `Dictionary<List<Dictionary<T>>>` output
- initially allow Boolean, signed integers, Float, Double, VARCHAR, and VARBINARY
- retain the existing REE behavior for unsupported types, null constants, empty
  vectors, and when the new option is disabled
- keep `flattenConstant=true` as the highest-precedence request
- include `arrayConstantAsDictionary` in the reusable Arrow schema cache key
- normalize array elements before both reusable ArrowArray export and schema
  cache lookup, keeping cached schemas and reusable arrays structurally aligned

### Performance Impact

- [x] **No Impact**: The new option is disabled by default, so existing callers
  retain the current export path.

When enabled for a constant array, the exported representation changes from
`N` copies of the list payload to `N` int32 indices plus one list value. This
removes payload duplication from the Arrow boundary.

### Release Note

```text
Release Note:
- Add an opt-in Arrow Dictionary representation for supported constant arrays.
```

### Checklist (For Author)

- [x] I have verified the code with a local Release build.

Validation:

```text
cmake --build _build/Release --target bolt_arrow_bridge_test -j 8
./_build/Release/bolt/vector/arrow/tests/bolt_arrow_bridge_test \
  --gtest_filter='ArrowBridgeArrayExportTest.*'

60 tests passed.
The 10 focused constant-array Dictionary and fallback tests also passed.
Two reusable-pool regression tests cover option toggling and a first batch with
dictionary-encoded array elements.
clang-format 14.0.6 and git diff --check passed.
```

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/vector/arrow/Bridge.cpp
+++ b/bolt/vector/arrow/Bridge.cpp
@@ -60,6 +60,58 @@ static constexpr size_t kMaxReusableScratchBufferBytes{32UL << 10};
 constexpr const char kArrowExtensionNameKey[] = "ARROW:extension:name";
 constexpr const char kSparkVariantExtensionName[] = "spark.variant";
 
+bool isSupportedArrayConstantElementType(const Type& type) {
+  if (type.isDate() || type.isDecimal()) {
+    return false;
+  }
+
+  switch (type.kind()) {
+    case TypeKind::BOOLEAN:
+    case TypeKind::TINYINT:
+    case TypeKind::SMALLINT:
+    case TypeKind::INTEGER:
+    case TypeKind::BIGINT:
+    case TypeKind::REAL:
+    case TypeKind::DOUBLE:
+    case TypeKind::VARCHAR:
+    case TypeKind::VARBINARY:
+      return true;
+    default:
+      return false;
+  }
+}
+} // namespace
+
+bool shouldExportArrayConstantAsDictionary(
+    const BaseVector& vector,
+    const ArrowOptions& options) {
+  if (vector.encoding() != VectorEncoding::Simple::CONSTANT ||
+      options.flattenConstant || !options.arrayConstantAsDictionary ||
+      vector.typeKind() != TypeKind::ARRAY || vector.size() == 0 ||
+      (vector.mayHaveNulls() && vector.isNullAt(0))) {
+    return false;
+  }
+  return isSupportedArrayConstantElementType(*vector.type()->childAt(0));
+}
+
+namespace {
+
+void normalizeArrayConstantForDictionaryExport(
+    const VectorPtr& vector,
+    const ArrowOptions& options) {
+  if (!shouldExportArrayConstantAsDictionary(*vector, options)) {
+    return;
+  }
+
+  // Preserve the outer Constant<Array<T>>, but flatten encoded elements such
+  // as those produced by ARRAY_DISTINCT. Consumers of this path support only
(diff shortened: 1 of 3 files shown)


wangxinshuo-bolt (approved):
LGTM

PR #978, opened 2026-09-07
feat(paimon): add generic Bolt filesystem support

### LINKED ISSUE #1009: [Feature] Enable Paimon to use registered Bolt filesystems with consistent option handling
### Feature Category

Connector / Storage (Hive, Iceberg, etc.)

### Problem / Use Case

The Paimon datasource currently uses Paimon’s local filesystem and a separate Bolt-backed HDFS adapter. This prevents it from consistently reusing Bolt’s registered storage backends and their configuration.
Configuration handling also has several gaps:
- Connector session properties do not override connector defaults.
- Filesystem credentials, endpoints, and file-open options are not consistently forwarded.
- Table-level reader options can be overwritten by connector defaults.
- Parsing a malformed fallback value can fail even when a valid query override exists.
A generic adapter also needs consistent file metadata and must preserve each backend’s rename and deletion semantics.

### Proposed Solution

Use a shared Bolt filesystem adapter
Replace the HDFS-specific adapter with PaimonBoltFileSystem, registered under the bolt identifier. Resolve each path through Bolt’s filesystem registry and pass the resolved configuration to the selected backend.
Backend availability remains dependent on build options and filesystem registration.
Define configuration precedence
- Merge connector session properties over connector defaults.
- Allow table properties to override filesystem defaults.
- For natural read size, read coalescing, and timestamp precision, apply this precedence:
  explicit query configuration > table properties > connector session properties > connector defaults
- Parse fallback reader options only when no higher-priority value exists.
- Forward configured file-open options to both read and write operations.
Provide consistent metadata and filesystem behavior
- Add a generic FileSystem::fileInfo() API exposing directory status, file size, and modification time.
- Implement metadata retrieval for local, HDFS, S3, GCS, and ABFS backends, including supported object-store directory representations.
- Handle local file: paths consistently for directory creation and removal.
- Reject renames across different filesystem backends or HDFS authorities.
- Use the backend’s non-recursive removal operation for non-recursive deletion.
- Return an explicit unsupported status for GCS directory deletion, while retaining ordinary file deletion.
- Include the affected path and underlying error in filesystem failures.
Add regression coverage
Cover option precedence, overridden malformed defaults, configuration forwarding, local filesystem operations, metadata retrieval, rename restrictions, and deletion behavior. Verify that rejected GCS directory deletion preserves directory markers, children, and sibling objects.

### References / Prior Art

Bolt filesystem abstraction: bolt/common/file/FileSystems.h
Paimon datasource integration: bolt/connectors/paimon/PaimonDataSource.cpp

### Importance

High (Needed for production)

### Willingness to Contribute

Yes, I can submit a PR

Weixin-Xu:
The Paimon connector currently uses Paimon's local filesystem by default and provides a Bolt-backed adapter only for HDFS. As a result, Paimon cannot consistently reuse Bolt's registered filesystem implementations or propagate connector, session, and table-level filesystem options.

This PR introduces a generic Bolt-backed Paimon filesystem adapter so Paimon can access local, HDFS, and other registered storage backends through the same filesystem registry.

Issue Number: close #1009 

### Type of Change

- [ ] 📝 Documentation only

### Description

- Generalize `PaimonBoltHdfsFileSystem` into `PaimonBoltFileSystem` and register it using the `bolt` identifier.
- Resolve Paimon paths through Bolt's filesystem registry instead of relying on filesystem-specific Paimon implementations.
- Support open, create, mkdir, rename, delete, status, listing, and existence operations through the generic adapter.
- Add `FileSystem::fileInfo()` and implement it for local and HDFS filesystems to provide file type, size, and modification time.
- Reject rename operations across different Bolt filesystem implementations or different HDFS authorities.
- Propagate filesystem configuration using the following precedence:
  - session properties override connector defaults;
  - table properties override connector and session defaults;
  - query-level reader options override connector-level reader defaults.
- Always enforce `file-system=bolt` so table properties cannot bypass the Bolt filesystem adapter.
- Remove the Paimon connector's direct dependency on Paimon's local filesystem and Bolt's HDFS implementation.
- Add tests for local filesystem operations, generic file metadata, configuration propagation, rename validation, error handling, and existing HDFS behavior.

### Performance Impact

    <details>

    ```text
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
- add generic filesystem datasource support
- honor Paimon options and filesystem semantics
```

### Checklist (For Author)

### Breaking Changes

- [ ] No
- [x] Yes (Description: ...)
    <details>

    ```text
    Breaking Changes:
    - Downstream binaries using FileSystem must be rebuilt.
    - Filesystem implementations used by Paimon should override fileInfo().
    - Existing source implementations remain compatible because the default method is non-pure virtual.
    ```
    </details>

### DIFF
--- a/bolt/connectors/paimon/PaimonBoltFileSystem.cpp
+++ b/bolt/connectors/paimon/PaimonBoltFileSystem.cpp
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#include "bolt/connectors/paimon/PaimonBoltHdfsFileSystem.h"
+#include "bolt/connectors/paimon/PaimonBoltFileSystem.h"
 
 #include <algorithm>
 #include <atomic>
@@ -28,8 +28,6 @@
 #include "bolt/common/config/Config.h"
 #include "bolt/common/file/File.h"
 #include "bolt/common/file/FileSystems.h"
-#include "bolt/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.h"
-#include "bolt/connectors/hive/storage_adapters/hdfs/RegisterHdfsFileSystem.h"
 
 #include "paimon/factories/factory_creator.h"
 #include "paimon/result.h"
@@ -40,7 +38,7 @@ namespace bytedance::bolt::connector::paimon {
 namespace {
 
 constexpr std::string_view kHdfsScheme{"hdfs://"};
-constexpr std::string_view kIdentifier{"bolt_hdfs"};
+constexpr std::string_view kIdentifier{"bolt"};
 
 std::string uriAuthorityPrefix(const std::string& uri) {
   if (uri.rfind(std::string(kHdfsScheme), 0) != 0) {
@@ -68,9 +66,9 @@ std::string joinDirAndBasename(
   return dir + "/" + basename;
 }
 
-class PaimonBoltHdfsInputStream final : public ::paimon::InputStream {
+class PaimonBoltInputStream final : public ::paimon::InputStream {
  public:
-  PaimonBoltHdfsInputStream(
+  PaimonBoltInputStream(
       std::shared_ptr<bytedance::bolt::ReadFile> file,
       std::string uri)
       : file_(std::move(file)), uri_(std::move(uri)) {}
@@ -156,9 +154,9 @@ class PaimonBoltHdfsInputStream final : public ::paimon::InputStream {
   std::atomic<int64_t> pos_{0};
 };
 
-class PaimonBoltHdfsOutputStream final : public ::paimon::OutputStream {
+class PaimonBoltOutputStream final : public ::paimon::OutputStream {
  public:
-  PaimonBoltHdfsOutputStream(
+  PaimonBoltOutputStream(
       std::shared_ptr<bytedance::bolt::WriteFile> file,
       std::string uri)
       : file_(std::move(file)), uri_(std::move(uri)) {}
@@ -211,9 +209,9 @@ class PaimonBoltHdfsOutputStream final : public ::paimon::OutputStream {
   int64_t pos_{0};
 };
 
-class PaimonBoltHdfsBasicFileStatus final : public ::paimon::BasicFileStatus {
+class PaimonBoltBasicFileStatus final : public ::paimon::BasicFileStatus {
  public:
-  PaimonBoltHdfsBasicFileStatus(std::string path, bool isDir)
+  PaimonBoltBasicFileStatus(std::string path, bool isDir)
       : path_(std::move(path)), isDir_(isDir) {}
 
   bool IsDir() const override {
@@ -229,9 +227,9 @@ class PaimonBoltHdfsBasicFileStatus final : public ::paimon::BasicFileStatus {
   bool isDir_{false};
 };
 
-class PaimonBoltHdfsFileStatus final : public ::paimon::FileStatus {
+class PaimonBoltFileStatus final : public ::paimon::FileStatus {
  public:
-  PaimonBoltHdfsFileStatus(
+  PaimonBoltFileStatus(
       std::string path,
       bool isDir,
       uint64_t len,
@@ -277,8 +275,8 @@ std::unordered_map<std::string, std::string> toUnordered(
 void ensurePaimonFactoryRegistered() {
   static std::once_flag flag;
   std::call_once(flag, []() {
-    auto* factory = new bytedance::bolt::connector::paimon::
-        PaimonBoltHdfsFileSystemFactory();
+    auto* factory =
+        new bytedance::bolt::connector::paimon::PaimonBoltFileSystemFactory();
     ::paimon::FactoryCreator::GetInstance()->Register(
         factory->Identifier(), factory);
   });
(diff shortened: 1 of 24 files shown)


bolt/connectors/paimon/PaimonDataSource.cpp
```diff
@@ -168,6 +180,66 @@ std::pair<std::shared_ptr<::paimon::Predicate>, core::TypedExprPtr> planFilter(
...
+  // Paimon's filesystem must always delegate to Bolt. Reader options are
+  // resolved last so query-level values take precedence over connector/table
+  // defaults.
+  options.insert_or_assign(::paimon::Options::FILE_SYSTEM, "bolt");
+  options.insert_or_assign(
+      ::paimon::Options::READ_BATCH_SIZE,
+      std::to_string(paimonConfig.readBatchSize()));
+  const auto readOptions =
+      resolvePaimonDataSourceReadOptions(queryConfig, paimonConfig);
+  options.insert_or_assign(
```
liuneng1994, line 234: This overwrites table-level values for `naturalReadSize`, `coalesceReads`, and `readTimestampUnit` whenever the query config does not specify them. For example, with `1024` in the connector config and `2048` in the table properties, the resolved value is still `1024`. Could we resolve these options with the intended precedence: query > table > session/connector?

bolt/connectors/paimon/PaimonBoltFileSystem.cpp
```diff
@@ -327,196 +325,179 @@ PaimonBoltHdfsFileSystem::Create(const std::string& path, bool overwrite)
...
     }
-
-    auto* hdfsFs =
-        dynamic_cast<bytedance::bolt::filesystems::HdfsFileSystem*>(fs.get());
-    if (!hdfsFs) {
-      return ::paimon::Status::NotImplemented(
-          "hdfs:// path did not resolve to HdfsFileSystem");
-    }
-    if (!hdfsFs->stat(directory).isDir) {
+    if (!fs->fileInfo(directory).isDirectory) {
```
liuneng1994, line 433: This makes `fileInfo()` mandatory for every filesystem used by Paimon, but this PR only implements it for local and HDFS. S3, GCS, ABFS, and existing external implementations still inherit the default `fileInfo not implemented`, so Paimon directory discovery fails even when the filesystem is registered. We should either implement the metadata API for every supported backend or limit the adapter to backends that provide it.

bolt/connectors/paimon/PaimonBoltFileSystem.cpp
```diff
@@ -277,47 +275,47 @@ std::unordered_map<std::string, std::string> toUnordered(
...
 
 ::paimon::Result<std::unique_ptr<::paimon::InputStream>>
-PaimonBoltHdfsFileSystem::Open(const std::string& path) const {
-  auto fs =
-      bytedance::bolt::filesystems::getFileSystem(path, connectorProperties_);
+PaimonBoltFileSystem::Open(const std::string& path) const {
   try {
+    auto fs =
+        bytedance::bolt::filesystems::getFileSystem(path, connectorProperties_);
     auto file = fs->openFileForRead(path, {});
```
liuneng1994: The filesystem receives the merged config, but the per-open `FileOptions` are empty here, and `Create()` only sets the creation flags. This drops options such as `bolt.io.file.buffer.size`, `bolt.dfs.replication`, and `bolt.dfs.blocksize`, which HDFS reads from `FileOptions`. Could we populate the options with `copyOpenFileOptionsFromConfig()` before both read and write opens?

bolt/connectors/paimon/PaimonBoltFileSystem.cpp
```diff
@@ -327,196 +325,179 @@ PaimonBoltHdfsFileSystem::Create(const std::string& path, bool overwrite)
...
-    if (!hdfsFs) {
-      return ::paimon::Status::NotImplemented(
-          "hdfs:// path did not resolve to HdfsFileSystem");
-    }
-
-    if (hdfsFs->stat(path).isDir) {
+    auto fs =
+        bytedance::bolt::filesystems::getFileSystem(path, connectorProperties_);
+    if (fs->fileInfo(path).isDirectory) {
       if (!recursive) {
```
liuneng1994: Non-recursive deletion should still succeed for an empty directory. The current check rejects all directories when `recursive` is false. A local filesystem test confirms that file deletion works and non-empty directories are rejected correctly, but empty-directory deletion now fails. Could we reject only non-empty directories here?

bolt/connectors/paimon/PaimonBoltFileSystem.cpp
```diff
@@ -277,246 +275,240 @@ std::unordered_map<std::string, std::string> toUnordered(
...
+::paimon::Status PaimonBoltFileSystem::ListDir(
     const std::string& directory,
     std::vector<std::unique_ptr<::paimon::BasicFileStatus>>* file_status_list)
     const {
-  auto fs = bytedance::bolt::filesystems::getFileSystem(
-      directory, connectorProperties_);
   try {
+    auto fs = bytedance::bolt::filesystems::getFileSystem(
+        directory, connectorProperties_);
     if (!fs->exists(directory)) {
```
liuneng1994: The adapter still assumes Bolt `exists()` / `list()` match the Paimon filesystem directory contract, but object stores do not behave that way. S3 prefix-only directories can be hidden by the exact `HeadObject` check, and GCS `list(path)` returns bucket-wide keys, which can be rejoined under the wrong directory here. I think this path needs object-store-aware listing: no exact-exists short-circuit for directories, and direct-child results normalized to the requested path.

bolt/connectors/paimon/PaimonBoltFileSystem.cpp
```diff
@@ -277,246 +275,240 @@ std::unordered_map<std::string, std::string> toUnordered(
...
 ::paimon::Result<std::unique_ptr<::paimon::OutputStream>>
-PaimonBoltHdfsFileSystem::Create(const std::string& path, bool overwrite)
-    const {
-  auto fs =
-      bytedance::bolt::filesystems::getFileSystem(path, connectorProperties_);
+PaimonBoltFileSystem::Create(const std::string& path, bool overwrite) const {
   try {
+    auto fs =
+        bytedance::bolt::filesystems::getFileSystem(path, connectorProperties_);
     if (fs->exists(path)) {
```
liuneng1994, line 374: This still breaks GCS creates. `GcsFileSystem::exists()` only checks that the bucket exists, so any new object under an existing bucket is treated as already present and `Create(..., false)` returns `Invalid` before reaching the GCS writer. Can we either make GCS `exists()` object-aware, or skip this pre-check and let the backend enforce overwrite?

liuneng1994 (approved):

PR #979, opened 2026-09-07
fix(parquet): report truncated definition levels

### LINKED ISSUE #980: [Bug] Truncated Parquet definition levels produce a generic decoder failure
### Component Selection

### Describe the Bug

When a malformed Parquet data page declares more values than its bounded definition-level payload encodes, the RLE/bit-pack decoder eventually reaches the end of the buffer and reports only a generic varint-header failure. The error does not identify that the definition-level count is short or provide enough page context to locate the corrupt input.

Definition levels carry nullness and value alignment, so missing levels cannot be reconstructed or padded safely.

### Reproduction Steps

1. Construct a bounded RLE/bit-packed definition-level stream containing two runs that encode 8 values in total.
2. Initialize the decoder with an expected page value count of 10.
3. Request all 10 definition levels.
4. Observe a generic truncated-varint/header error after the 8 encoded values are consumed.

### Bolt Version / Commit ID

`main` at `7e7e6727bd77b14c92701733fe7856145e316d4f`

### System Configuration

- **OS**: Platform-independent unit-test reproduction
- **Compiler**: Any supported compiler
- **Build Type**: Release or Debug
- **CPU Arch**: Any supported architecture
- **Framework**: Native Parquet reader

### Logs / Stack Trace

```shell
Parquet RLE/bit-pack decoder header failure: buffer_remaining=0 requested_values=2
```

### Expected Behavior

The reader should fail deterministically with a dedicated invalid-level-count diagnostic that reports the encoded, expected, and missing value counts together with the level role and page location. It must not pad the missing definition levels or continue decoding misaligned values.

### Additional context

The same decoder is also used for dictionary IDs. Generic truncated-header failures should retain role and bounded-buffer context, while the dedicated count mismatch should apply when an expected definition-level count is available.

fzhedu:

When a malformed Parquet page declares more values than its bounded definition-level payload encodes, the RLE/bit-pack decoder reaches the end of the buffer and reports only a generic varint-header failure. Missing definition levels carry nullness and value alignment, so they cannot be reconstructed or padded safely.

Issue Number: close #980

### Type of Change

- [ ] 📝 Documentation only

### Description

- Track the number of values consumed by `RleBpDecoder`.
- Pass the declared definition-level count and page location into bounded definition-level decoders.
- Fail with `PARQUET_INVALID_LEVEL_COUNT` when the page data ends before the declared value count.
- Report the level role, row group, column, page, encoded, expected, and missing counts, buffer offset, requested count, and bit width.
- Preserve contextual diagnostics for other malformed RLE/bit-pack headers, including dictionary-ID streams.
- Add focused tests for empty headers, truncated dictionary-ID headers, and definition-level count mismatches.

The reader deliberately fails instead of padding missing levels or silently accepting corrupt input.

### Performance Impact

- [x] **No Impact**: The change adds decoder counters and error context; the new diagnostic branch is reached only for malformed input.

### Validation

- Three focused `RleBpDecoder` regression tests passed in the original change environment.
- Eight self-contained Parquet reader regression tests passed in the original change environment.
- A malformed-input replay consistently failed fast with the expected encoded, expected, and missing counts.
- `pre-commit run --files bolt/dwio/parquet/reader/PageReader.cpp bolt/dwio/parquet/reader/RleBpDataDecoder.h bolt/dwio/parquet/reader/RleBpDecoder.cpp bolt/dwio/parquet/reader/RleBpDecoder.h bolt/dwio/parquet/tests/reader/RleBpDecoderTest.cpp` passed locally.
- `git diff --check upstream/main...HEAD` passed locally.
- A fresh local focused-test build was not completed because CMake generation in the existing checkout fails on an unrelated missing `duckdb::duckdb` target. GitHub build-and-test jobs provide the clean-environment validation.

### Release Note

```text
Release Note:
- Report a dedicated error with page context when a Parquet definition-level stream contains fewer values than the page declares.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

### DIFF
--- a/bolt/dwio/parquet/reader/PageReader.cpp
+++ b/bolt/dwio/parquet/reader/PageReader.cpp
@@ -181,6 +181,7 @@ void copyRawDataPageV1RepDefs(
 
 void makeDataPageV1RepDefDecoders(
     DataPageV1RepDefs repDefs,
+    int32_t numValues,
     int32_t maxRepeat,
     int32_t maxDefine,
     int32_t rowGroupOrdinal,
@@ -205,7 +206,8 @@ void makeDataPageV1RepDefDecoders(
           "definition-level",
           rowGroupOrdinal,
           columnOrdinal,
-          pageOrdinal);
+          pageOrdinal,
+          numValues);
     }
     wideDefineDecoder = std::make_unique<::arrow::util::RleDecoder>(
         reinterpret_cast<const uint8_t*>(repDefs.defineData),
@@ -543,6 +545,7 @@ void PageReader::prepareDataPageV1(
   } else {
     makeDataPageV1RepDefDecoders(
         repDefs,
+        numRepDefsInPage_,
         maxRepeat_,
         maxDefine_,
         cryptoCtx_.rowGroupOrdinal,
@@ -626,6 +629,7 @@ bool PageReader::tryPrepareDataPageV1RepDefOnly(
   } else {
     makeDataPageV1RepDefDecoders(
         repDefs,
+        numRepDefsInPage_,
         maxRepeat_,
         maxDefine_,
         cryptoCtx_.rowGroupOrdinal,
@@ -719,7 +723,8 @@ void PageReader::prepareDataPageV2(
         "definition-level",
         cryptoCtx_.rowGroupOrdinal,
         type_ ? static_cast<int32_t>(type_->column()) : -1,
-        pageOrdinal_);
+        pageOrdinal_,
+        numRepDefsInPage_);
     wideDefineDecoder_ = std::make_unique<::arrow::util::RleDecoder>(
         reinterpret_cast<const uint8_t*>(pageData_ + repeatLength),
         defineLength,
--- a/bolt/dwio/parquet/reader/PageReader.cpp
+++ b/bolt/dwio/parquet/reader/PageReader.cpp
@@ -183,6 +183,9 @@ void makeDataPageV1RepDefDecoders(
     DataPageV1RepDefs repDefs,
     int32_t maxRepeat,
     int32_t maxDefine,
+    int32_t rowGroupOrdinal,
+    int32_t columnOrdinal,
+    int32_t pageOrdinal,
     std::unique_ptr<::arrow::util::RleDecoder>& repeatDecoder,
     std::unique_ptr<RleBpDecoder>& defineDecoder,
     std::unique_ptr<::arrow::util::RleDecoder>& wideDefineDecoder) {
@@ -198,7 +201,11 @@ void makeDataPageV1RepDefDecoders(
       defineDecoder = std::make_unique<RleBpDecoder>(
           repDefs.defineData,
           repDefs.defineData + repDefs.defineSize,
-          ::arrow::bit_util::NumRequiredBits(maxDefine));
+          ::arrow::bit_util::NumRequiredBits(maxDefine),
+          "definition-level",
+          rowGroupOrdinal,
+          columnOrdinal,
+          pageOrdinal);
     }
     wideDefineDecoder = std::make_unique<::arrow::util::RleDecoder>(
         reinterpret_cast<const uint8_t*>(repDefs.defineData),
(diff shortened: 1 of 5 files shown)


bolt/dwio/parquet/reader/RleBpDecoder.cpp
```diff
@@ -92,18 +96,55 @@ void RleBpDecoder::readBits(
...
   folly::ByteRange headerRange(
       reinterpret_cast<const unsigned char*>(bufferStart_),
       reinterpret_cast<const unsigned char*>(bufferStart_ + maxVarIntLen));
   // decodeVarint() would advance headerRange's begin pointer.
-  auto indicator = folly::decodeVarint(headerRange);
+  uint64_t indicator;
+  try {
+    indicator = folly::decodeVarint(headerRange);
+  } catch (const std::exception& error) {
+    if (expectedValues_ >= 0 && bufferStart_ == bufferEnd_ &&
```
liuneng1994, line 115: The count mismatch is only detected when parsing the next run header fails. It does not cover a truncated payload after a valid header. For example, `{0x02}` declares one RLE value but has no value byte, and `{0x03}` declares one bit-packed group but has no packed payload. I reproduced both cases on this commit: `readBits()` returns without throwing and reads beyond the bounded range. Please validate the current run's payload size before loading, copying, or skipping it, and add coverage for both RLE and bit-packed payload truncation.

bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -201,7 +205,12 @@ void makeDataPageV1RepDefDecoders(
       defineDecoder = std::make_unique<RleBpDecoder>(
           repDefs.defineData,
           repDefs.defineData + repDefs.defineSize,
-          ::arrow::bit_util::NumRequiredBits(maxDefine));
+          ::arrow::bit_util::NumRequiredBits(maxDefine),
+          "definition-level",
+          rowGroupOrdinal,
+          columnOrdinal,
+          pageOrdinal,
+          numValues);
```
liuneng1994, line 213: This only applies `expectedValues` when `maxDefine == 1`. Nested columns still use Arrow's `RleDecoder`, including the default streaming path. I reproduced a V2 page that declares 10 definition levels but encodes only 8: streaming reports the generic `Invalid definition RLE run`, while legacy mode accepts the page because `readPageDefLevels()` ignores the short `GetBatch()` result. Please enforce `num_values` across the wide, streaming, and preloaded definition-level paths as well.

bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -941,7 +958,12 @@ void PageReader::prepareDataPageV2(
     defineDecoder_ = std::make_unique<RleBpDecoder>(
         pageData_ + repeatLength,
         pageData_ + repeatLength + defineLength,
```
liuneng1994, line 960: `repeatLength` and `defineLength` come from the page header, but this path does not verify that their sum fits the compressed and uncompressed page sizes before using them for pointer arithmetic, copying, and size subtraction. A one-byte page body with `definition_levels_byte_length = 100` caused a SIGSEGV in a focused test. The streaming V2 path already performs these checks; please apply the same validation here before constructing the decoders or copying raw levels.

bolt/dwio/parquet/reader/PageReader.cpp
```diff
@@ -941,7 +958,12 @@ void PageReader::prepareDataPageV2(
...
-        ::arrow::bit_util::NumRequiredBits(maxDefine_));
+        ::arrow::bit_util::NumRequiredBits(maxDefine_),
+        "definition-level",
+        cryptoCtx_.rowGroupOrdinal,
```
liuneng1994, line 963: The reported location is inaccurate for normal unencrypted files. `rowGroupOrdinal` is assigned only inside the encrypted-column branch, so the common path reports `row_group=-1`. Also, `pageOrdinal_` is incremented before decoder construction, which makes the first data page report `page=1`, while the existing AAD path treats it as ordinal 0. My PageReader reproduction produced `row_group=-1 column=0 page=1`. Please initialize the row-group ordinal independently of encryption and use a consistent page-numbering convention.

liuneng1994 (changes requested):

liuneng1994:
This feels more complex than necessary: it threads Parquet-specific context and count state through the generic RLE decoder and dictionary path, yet still misses truncated run payloads and the nested/streaming/preloaded level paths; could we keep payload-bound checks inside `RleBpDecoder`, validate decoded counts and V2 level lengths centrally in `PageReader`, and cover the result with focused flat and nested integration tests?

PR #783, opened 2026-07-24
 refactor: measure splitTime in doSplit and shuffleWriteTime at operator level

HaoChen-ch:
### Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🚀 Performance improvement (optimization)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🔨 Refactoring (no logic changes)
- [ ] 🔧 Build/CI or Infrastructure changes
- [ ] 📝 Documentation only

### Description

 1. Measure splitTime directly. Replace the subtraction with a dedicated NanosecondTimer splitTimer(&splitTime_) scoped to the real split work inside doSplit() (buildPartition2Row + preAlloc + splitRowVector). finalizeMetrics() now simply assigns metrics_.splitTime= splitTime_.

  2. Measure shuffleWriteTime at the operator level. Move it into SparkShuffleWriter (ShuffleWriterNode), wrapping addInput (split), reclaim (spill) and noMoreInput→stop. It now reflects the full wall time of the shuffle-write operator (init + addInput + reclaim + stop) instead of just split + stop.

  3. Narrow the flattenTime scope. Tighten the NanosecondTimer(&flattenTime_) blocks so they cover only the flatten calls (ensureFlatten / ensurePartialFlatten / ensureVectorLoaded), excluding the subsequent split/evict logic.
  4. Remove the now-unused totalSplitTime_ and stopTime_ members and their timers in split() / stop() across all three writers.

### Performance Impact

- [ ] **No Impact**: This change does not affect the critical path (e.g., build system, doc, error handling).
- [ ] **Positive Impact**: I have run benchmarks.
    <details>
    <summary>Click to view Benchmark Results</summary>

    ```text
    Paste your google-benchmark or TPC-H results here.
    Before: 10.5s
    After:   8.2s  (+20%)
    ```
    </details>
- [ ] **Negative Impact**: Explained below (e.g., trade-off for correctness).

### Release Note

Please describe the changes in this PR

Release Note:

```text
Release Note:
- Fixed a crash in `substr` when input is null.
- optimized `group by` performance by 20%.
```

### Checklist (For Author)

- [ ] I have added/updated unit tests (ctest).
- [ ] I have verified the code with local build (Release/Debug).
- [ ] I have run clang-format / linters.
[...]

PR #987, opened 2026-09-10
fix(shuffle): account external reclaims separately from shuffle write time

zhangxffff:

Issue Number: N/A — found while investigating a production metrics anomaly.

The reported shuffle write time could exceed the wall-clock duration of the task
that produced it. In one observed case a single task reported `shuffle total
write time` of 45.7 minutes while the task itself ran for 35.994 minutes (+27%),
and stage-wide shuffle write time (1,382.54 task-hours) exceeded the total
executor runtime of all attempts (1,325.67 task-hours).

Root cause: #783 (`refactor: measure splitTime in doSplit and shuffleWriteTime at
operator level`, commit 5cb127fe19) moved the `shuffleWriteTime` measurement to
the operator and added a third `NanosecondTimer` in
`SparkShuffleWriter::reclaim()` on the same counter that `addInput()` and
`noMoreInput()` (around `stop()`) already time. A reclaim caused by this
operator's own memory pressure runs *nested* inside those sections — the driver
thread is suspended inside `split()` while the arbitration and the eviction run
— so its wall time was accumulated twice. `NanosecondTimer` only performs
`counter += elapsed`, nothing de-duplicates overlapping intervals. Before #783
the split/stop time was measured at writer level and a nested reclaim was
covered exactly once by the enclosing split interval.

### Type of Change
- [ ] 📝 Documentation only

### Description

`SparkShuffleWriter::reclaim()` no longer adds its wall time to
`shuffleWriteTime`:

* A reclaim that is nested in a timed `addInput()`/`stop()` section is already
  part of that section's wall time, so `shuffleWriteTime` now accounts for it
  exactly once. As a result `shuffleWriteTime` can no longer exceed the duration
  of the task that reported it.
* Reclaims that are *not* covered by a timed section — those are triggered by
  memory pressure from outside this operator, e.g. another operator allocating
  while the writer is idle, or a shrink requested from another task — are timed
  into the new `externalReclaimTime` metric, so that cost stays visible without
  being double counted.

Telling the two cases apart requires knowing whether a timed section is active,
so `addInput()` and the `stop()` section scope a small `ShuffleWriteSection`
helper that times the section and keeps the `inShuffleSection_` flag set for
exactly that window. The flag is atomic because the reclaim callback can run on
another thread while the driver thread is suspended inside the timed section.

`ShuffleWriterMetrics` gains `externalReclaimTime` (additive field, existing
consumers keep working) and a `toString()` that formats every metric in a
human-readable way. `SparkShuffleWriter::noMoreInput()` logs these metrics at
`VLOG(1)` when the writer stops (per-partition arrays on a separate `VLOG(1)`
line), which makes this class of anomaly visible without a debugger or a custom
build.

The shuffle writer and reader operators also log their options when they are
constructed (`SparkShuffleWriter` / `SparkShuffleReader`), so the shuffle
configuration a task actually ran with is in the executor log (`VLOG(1)`)
next to the metrics. `ShuffleWriterOptions`, `ShuffleReaderOptions` and
`PartitionWriterOptions` get a `toString()` for that (formatters for the
partitioning, row format and codec names included); the implementations live in
`Options.cpp`.

Validation:

* `clang-format` 14.0.6 (the CI version) is clean on all touched C++ files.
* In the project devcontainer (Linux aarch64, gcc 12.5.0) the changed
  translation units `bolt/shuffle/sparksql/ShuffleWriterNode.cpp` and
  `bolt/shuffle/sparksql/Options.cpp` compile with the real Release flags taken
  from the build's `compile_commands.json`, and
  `bolt/shuffle/sparksql/tests/ShuffleMemoryTest.cpp` passes a syntax check.
* Not done locally: a full `make release_with_test` and running the gtest. The
  local conan cache has no `llvm-core` package for the current recipe revision,
  so conan would rebuild LLVM from source first. CI covers it: `lint`,
  `build-test (release)`, `build-test (release_spark)` (both run
  `make ctest_release`) and `build-only (debug_spark_with_test)` all pass on
  this head.

### Performance Impact
- [x] **No Impact**: This is instrumentation only. The change adds two atomic
  stores per `addInput()`/`stop()` call and one clock read per reclaim that is
  not covered by a timed section; the split/spill hot path is untouched. The
  added log line is emitted once per shuffle writer (at `stop()`).

### Release Note

```text
Release Note:
- Fixed shuffle write time being double counted when a memory reclaim ran inside
  a timed shuffle write call, which could report more time than the task itself took.
- Added an external reclaim time metric to the shuffle writer metrics and log all
  shuffle writer metrics when the writer stops.
- Log the shuffle writer and reader options when the operators are constructed.
```

### Checklist (For Author)

      Partial: the changed TU and the test TU compile in the devcontainer, see
      Validation above; a full build needs the missing conan LLVM package.

### Breaking Changes

- [x] No

### DIFF
--- a/bolt/shuffle/sparksql/CMakeLists.txt
+++ b/bolt/shuffle/sparksql/CMakeLists.txt
@@ -54,6 +54,7 @@ bolt_add_library(
   partitioner/Partitioning.cpp
   partitioner/RoundRobinPartitioner.cpp
   partitioner/SinglePartitioner.cpp
+  Options.cpp
   Payload.cpp
   ShuffleColumnarToRowConverter.cpp
   ShuffleMemoryPool.cpp
--- a/bolt/shuffle/sparksql/Options.cpp
+++ b/bolt/shuffle/sparksql/Options.cpp
@@ -1,0 +1,190 @@
+/*
+ * Copyright (c) ByteDance Ltd. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "bolt/shuffle/sparksql/Options.h"
+
+#include <fmt/format.h>
+#include <folly/String.h>
+
+#include "bolt/common/base/SuccinctPrinter.h"
+
+namespace bytedance::bolt::shuffle::sparksql {
+
+namespace {
+
+// Partitioning and row format names, matching the names used by the
+// configuration (see toPartitioning() and row::RowFormat).
+const char* partitioningName(Partitioning partitioning) {
+  switch (partitioning) {
+    case Partitioning::kSingle:
+      return "single";
+    case Partitioning::kRoundRobin:
+      return "rr";
+    case Partitioning::kHash:
+      return "hash";
+    case Partitioning::kRange:
+      return "range";
+  }
+  return "unknown";
+}
+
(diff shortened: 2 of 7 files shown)


HaoChen-ch:
LGTM

HaoChen-ch (approved):

Issue #991, opened 2026-09-10
[Build] setup-dev-env.sh generates an unusable Conan profile on Apple Clang 18+


linhongyu510:
### Summary of the Build Issue

`scripts/setup-dev-env.sh` runs `conan profile detect`, which writes the host compiler's major version into the default profile. On a machine with Apple Clang 18 or newer, that version is not in Conan 2.22.2's list of valid `apple-clang` versions, so every subsequent Conan invocation aborts during profile parsing — before dependency resolution, and long before any compilation.

The failure is in the generated profile, not in the toolchain: Apple Clang 21 itself compiles the project's C++17 fine. A fresh macOS setup following CONTRIBUTING.md therefore cannot get past step one without hand-editing `~/.conan2/profiles/default`.

### Build Command

```shell
# Following CONTRIBUTING.md "Option 1: Automatic Setup"
scripts/setup-dev-env.sh

# Or, minimally, just the profile detection that script performs:
conan profile detect --force
make release_with_test
```

### Build Environment (Crucial)

- **OS**: macOS 26.5.1 (25F80), arm64 (Apple Silicon)
- **Compiler**: Apple clang version 21.0.0 (clang-2100.1.1.101)
- **Conan Version**: 2.22.2 (the version pinned in `requirements.txt`)
- **CMake Version**: 4.4.3
- **Ninja Version**: 1.13.2
- **Python Version**: 3.14.7

### Conan Profile / Settings (Optional but Recommended)

```toml
# Exactly what `conan profile detect --force` produced on this machine:
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=21
os=Macos
```

### Error Log / Output

`conan profile detect` reports the compiler correctly:

```text
detect_api: Found apple-clang 21.0
detect_api: apple-clang>=13, using the major as version
...
Saving detected profile to /Users/<user>/.conan2/profiles/default
```

Any command that then reads that profile fails:

<details>
<summary>Click to expand</summary>

```text
$ conan profile show -pr:h ~/.conan2/profiles/default
ERROR: Invalid setting '21' is not a valid 'settings.compiler.version' value.
[...]

PR #993, opened 2026-09-10
feat(sparksql): support Spark elt

### LINKED ISSUE #923: [Feature] Support Spark elt
### Feature Category

SQL Function / Operator

### Problem / Use Case

Apache Spark provides `elt`, and Gluten already maps it (`Sig[Elt](ELT)` in
`ExpressionMappings.scala`), but Bolt registers no `elt`, so the expression
falls back.

### Proposed Solution

Implement Spark-compatible `elt`:

```text
elt(n: INT, input1, input2, ...) -> same type as inputs
```

Returns the `n`-th input (1-based). Spark returns NULL when `n` is out of
range, and the function is variadic over STRING (and BINARY) arguments.
Spark's `Elt` also carries a `failOnError` flag (visible in the plan above as
the trailing `false`), which follows `spark.sql.ansi.enabled`.

An alternative that needs no Bolt change is a Gluten-side rewrite to
`element_at(array(input1, input2, ...), n)` — both `array` and `element_at` are
already registered in Bolt's sparksql registry. Out-of-range behaviour would
need checking, since Spark's `elt` returns NULL while `element_at` on arrays
can throw depending on configuration.

### Reproduction Steps

```bash
```

Then run:

```sql
select elt(2, cast(id as string), 'b') as r from range(10);
```

```text
*(1) Project [elt(2, cast(id#746L as string), b, false) AS r#740]
+- *(1) ColumnarToRow
```

```text
```

regression, not a failure.

### References / Prior Art

- Spark SQL reference: https://spark.apache.org/docs/latest/api/sql/string-functions/#elt
- Apache Spark `Elt`
- MySQL `ELT()`, which Spark's semantics follow

### Acceptance Criteria

- Match Spark for in-range, out-of-range (returns NULL), and NULL index.
- Cover both STRING and BINARY inputs.
- Cover ANSI mode on/off if `failOnError` is implemented.

### Importance

Medium (Nice to have)

### Willingness to Contribute

Yes, but I need guidance

linhongyu510:

Issue Number: close #923

Bolt has no `elt` function, so Spark queries using `elt(n, input1, input2, ...)` fall back rather than running natively. This adds it.

### Type of Change

- [ ] 📝 Documentation only

### Description

`elt(n, input1, input2, ...)` returns the n-th input using a 1-based index, and returns NULL when the index is NULL or outside `[1, numInputs]`.

**Why a native function rather than a Gluten-side rewrite.** The issue suggests rewriting to `element_at(array(...), n)` so no Bolt change is needed. That rewrite does not preserve `elt` semantics. Comparing against the behavior asserted today in `ElementAtTest.cpp`, for an array of `{10, 11, 12}`:

| index | Spark `elt` | Bolt `element_at` (as asserted today) |
|---|---|---|
| 1..3 | 1st..3rd input | 10 / 11 / 12 — matches |
| 4 (out of upper bound) | NULL | NULL — matches |
| 0 | NULL | **throws** `"SQL array indices start at 1"` |
| -1 | NULL | **12** (last element) |

The upper-bound case agrees, but index `0` and negative indices do not: the rewrite would turn NULL-returning queries into failures, and would silently return the wrong element for negative indices. This is not a config matter — `element_at` is already registered as the non-ANSI variant in `RegisterMap.cpp`:

```cpp
// This is the semantics of spark.sql.ansi.enabled = false.
registerElementAtFunction(prefix + "element_at", true);
```

so the throw on `0` is the non-ANSI behavior, not something ANSI mode relaxes. Guarding each call site with `if(n >= 1 and n <= k, ...)` would build an intermediate array per row and push correctness onto every rewrite site.

**How it works.** `args[0]` is the index and `args[1..]` are the candidate inputs.

The result begins as all-NULL over the selected rows, so rows with a NULL or out-of-range index simply stay NULL and need no special casing. Rows are then **grouped by index value**, and each candidate input is copied at most once per batch with `BaseVector::copy`. The number of copy calls is therefore bounded by the number of inputs rather than the number of rows.

Using `copy` rather than manual per-row extraction means null propagation, string buffer sharing, and constant- / dictionary-encoded inputs are all handled by existing vector code rather than reimplemented here.

`isDefaultNullBehavior()` returns `false`, since a NULL index must produce NULL rather than propagate input nullability.

Signatures cover `varchar` and `varbinary` with `.variableArity()`, matching Spark's STRING/BINARY variadic inputs. Following `LeastGreatest.cpp`, which requires all value arguments to share one type, mixed input types are rejected in the factory; the comparison is against the first input rather than the index argument.

**Scope.** Spark's `Elt` carries a `failOnError` flag tied to `spark.sql.ansi.enabled`. This PR implements the non-ANSI behavior only, which is the same scope as the current `element_at` registration. ANSI-mode failure can follow separately if you'd like it.

### Performance Impact

New function; no existing code path is modified. The only changes to existing files are one source entry, one test entry, and the registration call.

### Release Note

```text
Release Note:
- Added the Spark `elt` function, which returns the n-th input using a 1-based index and returns NULL for a NULL or out-of-range index.
```

### Checklist (For Author)

- [x] I have added/updated unit tests (ctest). `EltTest.cpp`, 14 cases.
- [x] I have verified the code with local build (Release/Debug). Release build on macOS arm64; all 14 pass.
- [x] I have run clang-format / linters. clang-format 14.0.6, clean on all changed files.

<details>
<summary>Local test run</summary>

```text
Note: Google Test filter = EltTest.*
[==========] Running 14 tests from 1 test suite.
[----------] 14 tests from EltTest
[       OK ] EltTest.inRange (0 ms)
[       OK ] EltTest.singleInput (0 ms)
[       OK ] EltTest.indexOutOfRange (0 ms)
[       OK ] EltTest.nullIndex (0 ms)
[       OK ] EltTest.nullInput (0 ms)
[       OK ] EltTest.emptyString (0 ms)
[       OK ] EltTest.longStrings (0 ms)
[       OK ] EltTest.unicode (0 ms)
[       OK ] EltTest.vectorizedFlatIndex (0 ms)
[       OK ] EltTest.vectorizedWithNullInputs (0 ms)
[       OK ] EltTest.constantIndex (0 ms)
[       OK ] EltTest.constantInput (0 ms)
[       OK ] EltTest.dictionaryInput (0 ms)
[       OK ] EltTest.varbinary (0 ms)
[----------] 14 tests from EltTest (0 ms total)
[  PASSED  ] 14 tests.
```
</details>

Coverage: in-range selection and the single-input case; out-of-range indices (above the input count, `0`, negatives, `INT32_MIN`, `INT32_MAX`); NULL index; NULL in the selected input and NULL in a non-selected input (must not leak); empty, long (non-inlined) and unicode strings; multi-row dispatch with a different index per row; constant index; constant input; dictionary-encoded input; VARBINARY.

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)

---

Two things I want to flag rather than leave implicit:

1. **The Gluten fallback criterion in the issue is not verified.** Confirming the plan no longer falls back needs a Gluten + Bolt environment I don't have running, so I can't claim it. Everything else in the issue's acceptance criteria is covered by the tests above. Happy to follow up if someone can point me at a working setup.

2. **Two open questions from the issue thread**, repeated here since they may affect review: whether ANSI / `failOnError` should land in this PR or separately, and whether rejecting mixed input types is right versus applying implicit coercion. I picked the conservative option for both and can adjust.

I also hit two unrelated macOS build blockers on the way (#951 / #953 for `glog with_unwind`, and #991 for the Conan `apple-clang` version). Neither is part of this PR — flagging only in case CI on macOS is affected.

### DIFF
--- a/bolt/functions/sparksql/Elt.cpp
+++ b/bolt/functions/sparksql/Elt.cpp
@@ -1,0 +1,144 @@
+/*
+ * Copyright (c) ByteDance Ltd. and/or its affiliates.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "bolt/functions/sparksql/Elt.h"
+
+#include "bolt/expression/EvalCtx.h"
+#include "bolt/type/Type.h"
+namespace bytedance::bolt::functions::sparksql {
+namespace {
+
+/// Implements Spark's elt(n, input1, input2, ...).
+///
+/// The index argument is args[0]; the candidate inputs are args[1..n-1].
+/// Rows are grouped by the index value so that each candidate input is copied
+/// at most once, which keeps the number of copies bounded by the number of
+/// inputs rather than by the number of rows.
+class EltFunction final : public exec::VectorFunction {
+ public:
+  // elt() must return NULL, rather than propagating a NULL input, when the
+  // index is NULL or out of range. It therefore cannot use the default null
+  // behavior.
+  bool isDefaultNullBehavior() const override {
+    return false;
+  }
+
+  void apply(
+      const SelectivityVector& rows,
+      std::vector<VectorPtr>& args,
+      const TypePtr& outputType,
+      exec::EvalCtx& context,
+      VectorPtr& result) const override {
+    // args[0] is the index; the remaining arguments are the candidate inputs.
+    const auto numInputs = static_cast<int32_t>(args.size()) - 1;
+
+    context.ensureWritable(rows, outputType, result);
+
+    // Rows whose index is NULL or out of range stay NULL, so start from all
+    // NULL and only overwrite the rows that select a valid input.
+    rows.applyToSelected(
+        [&](vector_size_t row) { result->setNull(row, true); });
+
+    exec::LocalDecodedVector indexHolder(context, *args[0], rows);
+    const auto* decodedIndex = indexHolder.get();
+
+    // Group the rows by the input they select, so that each input vector is
+    // copied at most once for the whole batch.
+    exec::LocalSelectivityVector inputRowsHolder(context, rows.end());
+    auto* inputRows = inputRowsHolder.get();
+
+    for (int32_t input = 1; input <= numInputs; ++input) {
+      inputRows->clearAll();
+      bool hasRows = false;
+
+      rows.applyToSelected([&](vector_size_t row) {
+        if (decodedIndex->isNullAt(row)) {
+          return;
+        }
+        if (decodedIndex->valueAt<int32_t>(row) == input) {
+          inputRows->setValid(row, true);
+          hasRows = true;
+        }
+      });
+
+      if (!hasRows) {
+        continue;
+      }
+      inputRows->updateBounds();
+
+      // copy() handles nulls in the source, as well as dictionary- and
+      // constant-encoded inputs, so the selected input does not need to be
+      // decoded here.
+      result->copy(args[input].get(), *inputRows, nullptr, false);
+    }
+  }
+};
+
+} // namespace
+
+std::shared_ptr<exec::VectorFunction> makeElt(
+    const std::string& name,
+    const std::vector<exec::VectorFunctionArg>& inputArgs,
+    const core::QueryConfig& /*config*/) {
+  // One index argument plus at least one candidate input.
+  BOLT_USER_CHECK_GE(
+      inputArgs.size(),
+      2,
+      "{} requires an index argument and at least one input.",
+      name);
+
+  BOLT_USER_CHECK_EQ(
+      inputArgs[0].type->kind(),
+      TypeKind::INTEGER,
+      "The first argument of {} must be an integer.",
+      name);
+
+  // Spark requires all candidate inputs to share the same type.
+  for (size_t i = 2; i < inputArgs.size(); ++i) {
+    BOLT_USER_CHECK(
+        *inputArgs[i].type == *inputArgs[1].type,
+        "All inputs of {} must have the same type. Got {} and {}.",
+        name,
+        inputArgs[1].type->toString(),
+        inputArgs[i].type->toString());
+  }
+
+  const auto kind = inputArgs[1].type->kind();
+  BOLT_USER_CHECK(
+      kind == TypeKind::VARCHAR || kind == TypeKind::VARBINARY,
+      "{} does not support inputs of type {}.",
+      name,
+      inputArgs[1].type->toString());
+
+  return std::make_shared<EltFunction>();
+}
+
+std::vector<std::shared_ptr<exec::FunctionSignature>> eltSignatures() {
+  std::vector<std::shared_ptr<exec::FunctionSignature>> signatures;
+  for (const auto& type : {"varchar", "varbinary"}) {
+    // integer, T... -> T
+    signatures.emplace_back(exec::FunctionSignatureBuilder()
+                                .returnType(type)
+                                .argumentType("integer")
+                                .argumentType(type)
+                                .variableArity()
+                                .build());
+  }
+  return signatures;
+}
+
+} // namespace bytedance::bolt::functions::sparksql
--- a/bolt/functions/sparksql/CMakeLists.txt
+++ b/bolt/functions/sparksql/CMakeLists.txt
@@ -36,6 +36,7 @@ bolt_add_library(
   DecimalArithmetic.cpp
   DecimalCompare.cpp
   DecimalVectorFunctions.cpp
+  Elt.cpp
   FormatNumber.cpp
   Hash.cpp
   ICURegexFunctions.cpp
--- a/bolt/functions/sparksql/Elt.h
+++ b/bolt/functions/sparksql/Elt.h
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) ByteDance Ltd. and/or its affiliates.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "bolt/expression/VectorFunction.h"
+namespace bytedance::bolt::functions::sparksql {
+
+/// elt(n, input1, input2, ...) -> input_n
+///
+/// Returns the n-th input, where n is 1-based. Supported types for the
+/// variadic inputs are VARCHAR and VARBINARY.
+///
+/// Null behavior (matches Spark with spark.sql.ansi.enabled = false):
+///   - Returns NULL when n is NULL.
+///   - Returns NULL when n is out of the range [1, number of inputs], which
+///     includes n <= 0. Note this differs from element_at, which throws on
+///     index 0 and treats negative indices as offsets from the end.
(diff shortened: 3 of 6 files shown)


linhongyu510:
Design questions from #923 are now resolved by the maintainer, and both answers match this PR as written — no code changes needed:

- **Non-ANSI semantics first** — this PR has no throwing path; out-of-range and NULL indices return NULL. ANSI/`failOnError` can follow as a separate change.
- **Homogeneous candidate types required** — enforced in `Elt.cpp` via `BOLT_USER_CHECK(*inputArgs[i].type == *inputArgs[1].type, ...)`; no implicit conversion is added, matching Spark's analyzer-side coercion.

**Gluten offloading** — I was asked to verify this. I could not run a full Gluten + Spark stack in this environment, so instead of leaving the checklist item ambiguous I traced the name binding in `apache/incubator-gluten@main`:

| Step | Evidence |
|---|---|
| Spark `Elt` → name constant | `ExpressionMappings.scala:69` — `Sig[Elt](ELT)` |
| constant → wire name | `shims/common/.../ExpressionNames.scala:103` — `final val ELT = "elt"` |
| exposure to backends | `expressionsMap` (line 373) = all signatures minus the user-configurable `expressionBlacklist`; no `elt`-specific gate |
| Bolt side (this PR) | `RegisterString.cpp:116` — `prefix + "elt"` |

Gluten emits `elt`, Bolt now registers `elt`, and nothing else gates it — the previous behaviour was a plain missing-function fallback.

**Still unverified:** the runtime Substrait round-trip and the offloaded result of `select elt(2, cast(id as string), 'b') as r from range(10)`. The static chain above is necessary but not sufficient, so the corresponding checklist box stays unticked.

bolt/functions/sparksql/Elt.cpp
```diff
@@ -0,0 +1,144 @@
...
+///
+/// The index argument is args[0]; the candidate inputs are args[1..n-1].
+/// Rows are grouped by the index value so that each candidate input is copied
+/// at most once, which keeps the number of copies bounded by the number of
+/// inputs rather than by the number of rows.
+class EltFunction final : public exec::VectorFunction {
+ public:
+  // elt() must return NULL, rather than propagating a NULL input, when the
+  // index is NULL or out of range. It therefore cannot use the default null
+  // behavior.
```
zhangxffff, line 34: This comment is misleading about reason why cannot use the default null behavior.

With default null behavior, the engine sets a row to NULL when any argument is NULL, and apply() never sees that row. The real reason why we cannot  is that a NULL in an input that isn't selected must not make the result NULL. For example, elt(1, 'a', NULL) should return 'a', but the default null behavior would return NULL.

bolt/functions/sparksql/Elt.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    exec::LocalDecodedVector indexHolder(context, *args[0], rows);
+    const auto* decodedIndex = indexHolder.get();
+
+    // Group the rows by the input they select, so that each input vector is
+    // copied at most once for the whole batch.
+    exec::LocalSelectivityVector inputRowsHolder(context, rows.end());
+    auto* inputRows = inputRowsHolder.get();
+
+    for (int32_t input = 1; input <= numInputs; ++input) {
```
zhangxffff, line 63: Looping over the inputs isn't efficient here. For each input, the loop scans every selected row again, so the cost grows with rows × inputs.

It would be simpler and faster to loop over the rows once. For each row, read the index and read the value from `args[index]` using `DecodeVector`, then copy value straight into result.

linhongyu510:
The license gate failure is fixed in `5ba21d2`.

`Elt.h`, `Elt.cpp` and `EltTest.cpp` carried an extra line in their headers:

```
 * SPDX-License-Identifier: Apache-2.0
```

`.licenserc.yaml` accepts that line only in the derived-file variant, which also requires the Facebook copyright and the "This file has been modified by ByteDance Ltd." notice above it. These three files are new rather than derived, so the plain ByteDance-only header is the correct one, and it does not include an SPDX line. Removing it is the whole change — three deletions, no other edits.

Verified with the repository's own checker rather than by eye:

```
$ go run -C .github/license_check . bolt/functions/sparksql/Elt.h \
    bolt/functions/sparksql/Elt.cpp bolt/functions/sparksql/tests/EltTest.cpp
[INFO] License check passed successfully
$ echo $?
0
```

And confirmed the check is actually load-bearing here by putting the SPDX line back into `Elt.h` alone, which reproduces the original failure:

```
[ERROR] the following files don't have a valid license header:
$ echo $?
1
```

The other red checks on this PR (`build-only`, `build-test`, `merge-queue-gate`) ran in the same workflow as `lint`; I could not build the C++ targets locally to confirm they were downstream of the license gate, so I'd rather let the fresh run say than claim it. Could you re-run CI when convenient?

bolt/functions/sparksql/Elt.cpp
linhongyu510: You are right, and the comment was worse than imprecise — it described a mechanism that does not happen. With the default null behavior the engine sets the row to NULL when any argument is NULL and `apply()` is never called for it, so "propagating a NULL input" was not the issue. The actual reason is the one you gave: a NULL in an input this row does not select must not make the result NULL, i.e. `elt(1, 'a', NULL)` is `'a'`.

Rewritten in `ab2076e` to say exactly that, with your example inline.

bolt/functions/sparksql/Elt.cpp
linhongyu510: Done in `ab2076e`, implemented the way you described.

Each input is now decoded once up front, then the rows are walked once: each row reads its index and reads only the input that index selects, through `DecodedVector`. The per-input rescan is gone, so the cost is O(rows) instead of O(rows × inputs).

Two details worth flagging in the new version:

**Values are referenced, not copied.** The result uses `setNoCopy` plus `acquireSharedStringBuffers`, matching the idiom already in `SplitFunctions.cpp` and `StringSplitSQL.cpp`. That avoids duplicating the bytes as well as the scan — the previous `copy()` per input moved data even for rows that were already correct.

**Dictionary and constant inputs need no special handling here.** `acquireSharedStringBuffers` calls `wrappedVector()` internally and takes the buffers from the base, so the borrow is still correct for the encodings `dictionaryInput` and `constantInput` cover. Reading through `DecodedVector` handles the index mapping.

Added `longStringsOutliveInputs`: the input vectors are dropped before the result is asserted, so non-inlined strings only read correctly if the result really retained the buffers. That invariant is free in a copying implementation and load-bearing in this one, so it deserved its own test.

One thing I should be straight about: **I could not build or run the suite locally** — this checkout has no Conan/vendored dependencies, so `clang++ -fsyntax-only` stops at `fmt/format.h`. Rather than claim a green run I do not have, I verified statically against the headers in-tree: `FlatVector<StringView>::setNoCopy` and `acquireSharedStringBuffers` are the real template specializations in `FlatVector.cpp:180` / `:197`, `LocalDecodedVector` has a move constructor so the `std::vector` of holders is valid, and `DecodedVector::valueAt` / `isNullAt` are `const`. CI on this PR is the real check; if it turns red I will fix it rather than leave it.

bolt/functions/sparksql/Elt.cpp
```diff
@@ -0,0 +1,148 @@
...
+      inputHolders.emplace_back(context, *args[input], rows);
+      decodedInputs.push_back(inputHolders.back().get());
+    }
+
+    // The result borrows the inputs' string buffers instead of copying their
+    // bytes, so a selected value is referenced rather than duplicated.
+    for (int32_t input = 1; input <= numInputs; ++input) {
+      flatResult->acquireSharedStringBuffers(args[input].get());
+    }
```
zhangxffff, line 71: Maybe we could use a std::vector<bool> to track which input columns are referenced and only acquire shared buffers for those columns. This could help reduce memory usage.

zhangxffff: @linhongyu510 Thanks for your contribution, only a small suggestion here. btw, you should also solve the clang-format problem to esure check pass.

linhongyu510:
The remaining `lint` failure is the clang-format hook, now fixed in `3fa3504`.

Only `EltTest.cpp` was affected — one line in `longStringsOutliveInputs` exceeded the column limit:

```cpp
result = evaluate("elt(c0, c1, c2)", makeRowVector({index, input0, input1}));
```

Ran the repo's pinned `clang-format` (mirrors-clang-format `v14.0.6`, per `.pre-commit-config.yaml`) with `-i`; it wraps that call across two lines. Formatting only, no behavior change. Re-checked all four changed C++ files (`Elt.h`, `Elt.cpp`, `RegisterString.cpp`, `EltTest.cpp`) with `--dry-run --Werror` — all clean now.

The workflow run on `3fa3504` shows `action_required` (the external-fork approval gate), so the checks won't re-run until a maintainer approves them. Could you approve the run? Should be green after that.

bolt/functions/sparksql/Elt.cpp
linhongyu510: Done in `4cf88cb`. I scan the index column once first to mark which inputs are actually referenced by some selected row (a `std::vector<bool>`), then acquire shared string buffers only for those — an input no row indexes into no longer pins its buffers into the result. The extra pass is O(rows) and runs before the write loop, so the row walk itself is unchanged.

Also fixed the clang-format issue (the earlier `EltTest.cpp` line plus this new block), verified with the repo's pinned clang-format v14.0.6. Thanks for the review!

PR #1003, opened 2026-09-14
fix(flinksql): align round semantics and return_type

tianjq16:

Issue Number: N/A

### Type of Change

- [ ] 📝 Documentation only

### Description

For non-DECIMAL numeric types, the semantics are the same as Spark SQL, so the existing implementation can be reused directly.

For DECIMAL types, in Flink, the return type of ROUND(DECIMAL(p, s), r) follows the rules below:

| Condition | Return Type |
| --- | --- |
| r >= s | DECIMAL(p, s) |
| r < 0 | DECIMAL(min(38, 1 + p - s), 0) |
| 0 <= r < s  | DECIMAL(1 + p - s + r, r)

Flink's return-type inference for the ROUND function differs from that of Presto and Spark. Therefore, we need to implement a new Flink-compatible decimal_round function.

In addition to return-type inference, the new Flink-compatible decimal_round function also differs in its handling of overflow and NULL values.

### Performance Impact

    <details>

    ```text
    N/A
    ```
    </details>

### Release Note

Release Note:

```text
Release Note:
- Add Flink-compatible round function.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)
    <details>

    ```text
    N/A
    ```
    </details>


bolt/functions/flinksql/specialforms/DecimalRound.cpp
```diff
@@ -0,0 +1,257 @@
...
+                 inputPrecision_,
+                 inputScale_,
+                 resultPrecision_,
+                 resultScale_,
+                 result)
+          .ok();
+    }
+
+    TResult rounded;
+    DecimalUtil::divideWithRoundUp<TResult, TInput, int128_t>(
```
yangzhg, line 102: 这里TResult 如果是shortdecimal 会不会 丢失精度

luozenglin, line 102: `divideWithRoundUp<TResult, TInput, int128_t>` stores the dividend in its result type (`R unsignedDividendRescaled(a)`). For a long-decimal input and a short-decimal result, this narrows the unscaled input to `int64_t` before division.

For example, `ROUND(CAST('9999999999.9999999999' AS DECIMAL(20,10)), -1)` should return `10000000000`, but a standalone reproduction of this arithmetic path returns `776627960` in both unoptimized and optimized builds. The final precision check cannot detect this because the corrupted result still fits.

Please retain `int128_t` intermediates through division, rounding, and multiplication, then narrow after validating the final precision. Add negative-scale long-to-short tests whose unscaled inputs exceed `int64_t`, including negative values.

bolt/functions/flinksql/specialforms/DecimalRound.cpp
```diff
@@ -0,0 +1,257 @@
...
+        rounded, input, divideFactor_.value(), false, 0, 0);
+    int128_t product;
```
luozenglin, line 104: The called helper tests `remainder * 2 >= unsignedDivisor`. When the divisor is `10^38`, a valid decimal remainder can overflow signed `int128_t` during this multiplication.

For the maximum `DECIMAL(38,0)` value rounded at scale `-38`, Flink returns NULL because the rounded value is `10^38`. A standalone reproduction of this path instead returns `0`; UBSan reports signed integer overflow in the helper. The multiplication check below runs after the rounding decision has already been corrupted.

Please use an overflow-safe comparison. For these positive powers of ten, comparing the nonnegative remainder against `divisor / 2` avoids multiplication. Add coverage where `inputScale - roundScale == 38`, including both signs.

bolt/functions/flinksql/specialforms/DecimalRound.cpp
```diff
@@ -0,0 +1,257 @@
...
+                 inputScale_,
+                 resultPrecision_,
+                 resultScale_,
+                 result)
+          .ok();
+    }
+
+    TResult rounded;
+    DecimalUtil::divideWithRoundUp<TResult, TInput, int128_t>(
+        rounded, input, divideFactor_.value(), false, 0, 0);
```
luozenglin, line 103: A zero input with a supported negative scale, such as `ROUND(CAST('0.00' AS DECIMAL(5,2)), -1)`, reaches this call. Inside `divideWithRoundUp`, the rescaling check evaluates `kPowersOfTen[aRescale] * unsignedDividendRescaled / unsignedDividendRescaled`, which becomes `0 / 0`.

A standalone UBSan reproduction reports division by zero. Ordinary builds returned the expected zero in this experiment, so the existing zero-valued assertions do not establish that this path is free of undefined behavior.

Please return zero before this division, or make the helper explicitly handle zero and the no-rescaling case. Validate both short and long decimals under UBSan.

tianjq16:
@yangzhg @luozenglin Thanks for your comment.

Initially, for simplicity, I directly reused the implementation from `sparksql::DecimalRoundFunction` without checking whether overflow could lead to incorrect results. The implementation of `sparksql::DecimalRoundFunction` appears to have the same issue.

I chose not to modify the logic in `DecimalUtil::divideWithRoundUp` because I think it is unnecessarily complex and relatively inefficient for implementing the `ROUND` function. Instead, I implemented simpler logic specifically for the `scale < 0` case, which avoids the complexity introduced by potential overflow.

Finally, I think the following items should be investigated, but they are outside the scope of this PR and can be tracked in separate issues:

- Confirm whether Spark SQL and Presto SQL have the same issue.
- Investigate the potential overflow issue in divideWithRoundUp.

yangzhg (approved):
LGTM

PR #1016, opened 2026-09-16
fix: make FloatingDecimal power-of-five cache thread-safe

### LINKED ISSUE #1015: [Bug] Data race in FloatingDecimal::big5pow power-of-five cache
### Component Selection

- Core Engine (Expression eval, Memory, Vector)

### Describe the Bug

`FloatingDecimal::big5pow(int p)` computes `5^p` for decimal-to-floating-point conversion. It caches results in a **static, mutable** member `std::vector<boost::multiprecision::cpp_int> FloatingDecimal::b5p;`, which it lazily grows and populates on demand:

```cpp
std::vector<boost::multiprecision::cpp_int> FloatingDecimal::b5p;

boost::multiprecision::cpp_int FloatingDecimal::big5pow(int p) {
  BOLT_USER_CHECK(p >= 0);

  bool include = b5p.size() > p;
  if (!include) {
    b5p.resize(p + 1);      // <-- unsynchronized mutation of shared state
  }
  if (b5p[p] != 0) {
    return b5p[p];
  }
  ...
  return (b5p[p] = ...);    // <-- unsynchronized write of shared state
}
```

Because `b5p` is a process-wide shared object accessed with no synchronization, concurrent calls to `big5pow` from multiple threads race on `b5p.resize(...)` and on element reads/writes. `std::vector::resize` may reallocate and move the backing storage while another thread is reading it, leading to:

- data races (flagged by TSAN),
- corrupted / partially-written `cpp_int` values → **wrong conversion results**,
- use-after-free / heap corruption when a reader dereferences storage freed by a concurrent `resize` → **crashes**.

Decimal→double/float conversion is on a hot, highly concurrent path (many worker threads convert decimals in parallel), so this is readily triggerable under load.

### Reproduction Steps

Run `big5pow` from many threads concurrently for exponents beyond the small precomputed range (i.e. exponents that force the cache to grow). A minimal reproducer (also added as a unit test in the fix):

```cpp
constexpr int32_t kNumThreads = 32;
std::vector<std::thread> threads;
for (int32_t i = 0; i < kNumThreads; ++i) {
  threads.emplace_back([&, i]() {
    for (int32_t it = 0; it < 100; ++it) {
      const int32_t exponent = 64 + i * 9 + it % 3;
      EXPECT_EQ(
          FloatingDecimal::big5pow(exponent),
          boost::multiprecision::pow(
              boost::multiprecision::cpp_int{5}, exponent));
    }
  });
}
for (auto& t : threads) t.join();
```

Under TSAN this reports data races on `b5p`; without a sanitizer it can produce mismatched values or crash intermittently.

### Bolt Version / Commit ID

main branch

### System Configuration

- **OS**: Linux
- **Compiler**: Clang
- **Build Type**: Debug / Release (ASAN/TSAN)
- **CPU Arch**: x86_64
- **Framework**: Spark (Gluten Bolt backend)

### Expected Behavior

`FloatingDecimal::big5pow` should be safe to call concurrently from multiple threads and always return the correct value of `5^p`, without data races, corrupted results, or crashes.

### Additional context

The set of powers of five needed is bounded and cheap to precompute once. The fix replaces the lazily grown shared vector with an immutable lookup table built a single time via a function-local `static` (thread-safe initialization guaranteed by the C++ standard), and falls back to `boost::multiprecision::pow` for exponents beyond the cached range. The now-unused `long5pow`, `maxLongFive`, and `b5p` members are removed.

wangxinshuo-bolt:

Issue Number: close #1015

`FloatingDecimal::big5pow(int p)` cached powers of five in a static, mutable
`std::vector<boost::multiprecision::cpp_int> b5p` that it lazily `resize`d and
populated on demand. This shared state was accessed with no synchronization, so
concurrent calls from multiple threads raced on `b5p.resize(...)` and on element
reads/writes. Since `resize` can reallocate the backing storage while another
thread is reading it, this caused data races (flagged by TSAN), corrupted
conversion results, and intermittent use-after-free crashes on the hot,
highly-concurrent decimal→floating-point conversion path.

### Type of Change

- [ ] 📝 Documentation only

### Description

- Replace the lazily grown, shared `b5p` vector with an immutable lookup table
  of `5^p` for `p` in `[0, 340]`, built exactly once via a function-local
  `static` (thread-safe initialization is guaranteed by the C++ standard).
- For exponents beyond the cached range, compute the value directly with
  `boost::multiprecision::pow(5, p)` instead of mutating shared state.
- Remove the now-unused `long5pow`, `maxLongFive`, and `b5p` members from
  `FloatingDecimal`.

Why this approach: the set of powers of five that can be requested is bounded
and cheap to precompute a single time, so an immutable table eliminates the
mutation-under-concurrency entirely rather than guarding it with a lock. This
keeps the read path allocation- and lock-free while being correct under
concurrency. `340` covers the maximum exponent reachable via decimal
conversion.

### Performance Impact

- [x] **No Impact**: The cache is built once at first use; subsequent lookups are
  a plain indexed read into a fixed array (no locking, no allocation), matching
  or improving on the previous cached-read behavior.

### Release Note

Release Note:

```text
Release Note:
- Fixed a data race / potential crash in decimal-to-floating-point conversion caused by an unsynchronized power-of-five cache in FloatingDecimal::big5pow.
```

### Checklist (For Author)

### Breaking Changes

- [x] No
- [ ] Yes (Description: ...)


fzhedu (approved):
LGTM

